PR #60, opened 2018-06-28
Fill op

kbalka:


src/ngraph_builder.cc
```diff
@@ -857,6 +857,51 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+      TF_RETURN_IF_ERROR(op->input_node(0, &tf_dims));
+      TF_RETURN_IF_ERROR(op->input_node(1, &tf_value));
+
+      try {
+        ng_op_map.at(tf_dims->name()); 
+      }
+      catch (const std::out_of_range&) {
+        return tf::errors::InvalidArgument("Missing input: " + tf_dims->name());
+      }
+      auto ng_dims = ng_op_map.at(tf_dims->name());
```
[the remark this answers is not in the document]
kbalka: Yes, that was my idea. I think its impossible to declare `auto` variable without initialization. I've changed type to `shared_ptr<ng::Node> ` and declared variable before `try` block.

src/ngraph_builder.cc
```diff
@@ -857,6 +857,49 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+      shared_ptr<ng::Node> ng_value;
+      try {
+        ng_value = ng_op_map.at(tf_value->name()); 
+      } catch (const std::out_of_range&) {
+        return tf::errors::InvalidArgument("Missing input: " + tf_value->name());
+      }
+      ng_value = ng_op_map.at(tf_value->name());
+     
+      auto ng_dims_const =
+          std::dynamic_pointer_cast<ng::op::Constant>(ng_dims);
```
aprocter: Instead of trying to force the argument to `ng::op::Constant` here, we should use the value in the `_ngraph_fill_static_dims` field. (Under some circumstances, the constant node can end up being placed in a separate graph.)

test/python/test_fill.py
```diff
@@ -0,0 +1,75 @@
...
+import pytest
+
+from common import NgraphTest
+
+
+class TestFillOperations(NgraphTest):
+  @pytest.mark.parametrize(
+      ("shape", "value"),
+      ((
+       ([1, 2, 3, 1], 10),
```
aprocter: Might also try shapes of `[]` and `[0]`, if those happen to give reasonable results in the non-nGraph impelmentation (I don't know if they actually do).

PR #64, opened 2018-07-02
[WIP] osx wheel

megawidget:
It "works" but hardly elegant.

Another issue with the current wheel is that all the libraries are copied instead of symlinked.  No idea why.  CMake?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ endif()
 ExternalProject_Add(
     ext_ngraph
     GIT_REPOSITORY https://github.com/NervanaSystems/ngraph
-    GIT_TAG 894aed9a61d33fa32eb5d2f226fa26733bb416b1
+    GIT_TAG e4db82ecfff6941700915d617c423a23da880847
     CMAKE_ARGS -DNGRAPH_INSTALL_PREFIX=${CMAKE_CURRENT_BINARY_DIR}/ngraph/ngraph_dist -DNGRAPH_USE_PREBUILT_LLVM=${NGRAPH_USE_PREBUILT_LLVM}
     TMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/ngraph/tmp"
     STAMP_DIR "${CMAKE_CURRENT_BINARY_DIR}/ngraph/stamp"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,7 @@ set_target_properties(
 # 1. $ORIGIN/.
 # 2. $ORIGIN/../tensorflow/
 # 
+set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_INSTALL_RPATH "\$ORIGIN:\$ORIGIN/../tensorflow")
 
 # Add the directories to be built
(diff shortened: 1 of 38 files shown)


avijit-nervana:
Ok - this got me past the build issue - thanks for the fix. I also see that the wheel has proper name which is great. But I am seeing the runtime error - which is related to the rpath. Following is the error:
```
>>> import ngraph
TensorFlow version:  v1.8.0-0-g93bc2e2072 1.8.0
Version needed:  v1.8.0-0-g93bc2e2072
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/avijitch/Projects/MacOS/osx/lib/python2.7/site-packages/ngraph/__init__.py", line 43, in <module>
    lib = ctypes.cdll.LoadLibrary(os.path.join(libpath,'libngraph_device.'+ext))
  File "/usr/local/Cellar/python/2.7.13_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ctypes/__init__.py", line 440, in LoadLibrary
    return self._dlltype(name)
  File "/usr/local/Cellar/python/2.7.13_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ctypes/__init__.py", line 362, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/Users/avijitch/Projects/MacOS/osx/lib/python2.7/site-packages/ngraph/libngraph_device.dylib, 6): Library not loaded: libngraph.0.4.dylib
  Referenced from: /Users/avijitch/Projects/MacOS/osx/lib/python2.7/site-packages/ngraph/libngraph_device.dylib
  Reason: image not found
>>> 
```
If I run the otool command here's what I see:
```
(osx) ➜  ngraph otool -L libngraph_device.dylib
libngraph_device.dylib:
	@rpath/libngraph_device.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libtensorflow_framework.so (compatibility version 0.0.0, current version 0.0.0)
	libngraph.0.4.dylib (compatibility version 0.4.0, current version 0.4.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.5.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

Obviously it's working on your system. So - can you update the cmake so that if the `install_name_tool` command fails - the cmake aborts or send out a message or something? 

Then we will know why it didn't work on my system.

avijit-nervana:
I will debug it further. Looks like we need to use `@loader_path` not `@rpath`. With the later it works when I am inside the `<venv>/lib/python<x>/site-packages/ngraph` but not from anywhere else.

megawidget:
```
(.ngraph-tf) ikaploun@ikaploun-mac02 ~/s/n/b/p/dist ~ $ python
Python 3.6.5 (default, Mar 29 2018, 15:37:32) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import ngraph
TensorFlow version:  b'v1.8.0-8-g23c218785e' 1.8.0
Version needed:  b'v1.8.0-8-g23c218785e'
>>> 
```
confirmed it works on my Mac with Python 3

PR #65, opened 2018-07-02
Implement keep_dims for mean and sum

aprocter:


avijit-nervana:
Please resolve the conflict with mater and push an update.

aprocter:
Since #68 is in I can remove the DT_STRING stuff... will do that now.

src/ngraph_cluster.cc
```diff
@@ -167,13 +166,15 @@ tf::Status NGraphClusterPass::IdentifyClusters(tf::Graph* graph) {
     }
 
     if (!gc.InsertEdge(cluster_map[src]->index, cluster_map[dst]->index)) {
-      return tf::errors::InvalidArgument(
+      NGRAPH_VLOG(5) << "Failing due to cycle";
+      return tf::errors::Unimplemented(
```
aprocter: Kinda snuck this one in there... one of the scripts we were testing with catches `InvalidArgument` and silently ignores it, which is an odd choice but there you have it. `Unimplemented` is probably a better choice anyway.

PR #68, opened 2018-07-03
 Move type checks for stubbed ops into confirmation pass

aprocter:


src/ngraph_utils.cc
```diff
@@ -140,4 +140,24 @@ tf::Status TFTensorShapeToNGraphShape(const tf::TensorShape& tf_shape,
   return tf::Status::OK();
 }
 
+const std::set<tf::DataType>* NGraphDTypes() {
```
avijit-nervana: Since this is a super set of Numeric and Index datatypes - is there a better way to represent this as opposed to duplicating the code?

src/ngraph_utils.h
```diff
@@ -156,6 +156,18 @@ tf::Status TFDataTypeToNGraphElementType(tf::DataType tf_dt,
 tf::Status TFTensorShapeToNGraphShape(const tf::TensorShape& tf_shape,
                                       ngraph::Shape* ng_shape);
 
+// Returns a pointer to a set containing all TensorFlow dtypes supported by the
+// nGraph bridge. This set is a singleton object.
+const std::set<tf::DataType>* NGraphDTypes();
```
avijit-nervana: Why not a `const &` as opposed to `*`?

src/ngraph_utils.h
```diff
@@ -156,6 +156,18 @@ tf::Status TFDataTypeToNGraphElementType(tf::DataType tf_dt,
...
+
+// Returns a pointer to a set containing all *numeric* TensorFlow dtypes
+// supported by the nGraph bridge. This set is a singleton object.
+const std::set<tf::DataType>* NGraphNumericDTypes();
```
avijit-nervana: Same comment as above

src/ngraph_op_kernels.cc
```diff
@@ -114,18 +114,22 @@ class NGraphIdentityOp : public OpKernel {
...
-  REGISTER_KERNEL_BUILDER(Name("Identity")                          \
-                              .Device(ngraph_bridge::DEVICE_NGRAPH) \
-                              .TypeConstraint<T>("T"),              \
-                          NGraphIdentityOp);
-
-REGISTER_IDENTITY(float);
-REGISTER_IDENTITY(int32);
-REGISTER_IDENTITY(int64);
+// A bit of unfortuante silliness here: TypeConstraint takes a gtl::ArraySlice
+// for the allowed types, but NGraphDTypes is an std::set and there is
```
avijit-nervana: We can use gtl::ArraySlice as the storage for NGraphDataTypes - no?

src/ngraph_confirm_pass.cc
```diff
@@ -60,7 +60,12 @@ extern const char* const DEVICE_NGRAPH;
 // kernels will not be allowed for nodes we did not mark during this pass, and
 // placement will fall back on CPU.
 //
-// The per-op checks are implemented by callbacks of the type:
+// There are two pieces here. The first is a type constraint checker, which
+// supplants the type checking machinery usually used with
+// REGISTER_KERNEL_BUILDER. The second is a set of finer-grained per-op checks
+// called "confirmation functions".
```
avijit-nervana: May be illustrate what is `confirmation pass` and have an example?

src/ngraph_utils.h
aprocter: I had a reason, but I'm realizing now that it was over-paranoid. Will change this.

src/ngraph_utils.cc
aprocter: I kinda like it better this way. If we do anything fancy with the initialization we'll need a mutex/is_initialized dance, which seems a bit hairy.

src/ngraph_op_kernels.cc
aprocter: I reckon we can, actually. Will do

PR #70, opened 2018-07-06
Conv2DBackpropInput

megawidget:
clearly, I'm still confused; currently the test crashes

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -695,13 +695,16 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
       ng_dilations[1] = tf_dilations[1];
 
       if (is_nhwc) {
-        // FIXME: need to reshape "out_backprop" here.
         ng_image_shape[0] = tf_input_sizes[1];
         ng_image_shape[1] = tf_input_sizes[2];
         ng_batch_shape = { static_cast<unsigned long>(tf_input_sizes[0]),
                            static_cast<unsigned long>(tf_input_sizes[3]),
                            static_cast<unsigned long>(tf_input_sizes[1]),
                            static_cast<unsigned long>(tf_input_sizes[2]) };
+        auto& s = ng_out_backprop->get_shape();
+        ng::Shape reshaped{s[0], s[3], s[1], s[2]};
+        ng_out_backprop = make_shared<ng::op::Reshape>(
+            ng_out_backprop, ng::AxisVector{0, 3, 1, 2}, reshaped);
       } else {
         ng_image_shape[0] = tf_input_sizes[2];
         ng_image_shape[1] = tf_input_sizes[3];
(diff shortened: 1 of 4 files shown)


src/ngraph_builder.cc
```diff
@@ -633,8 +633,152 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+    else if (op->type_string() == "Conv2DBackpropInput") {
+      if (op->num_inputs() != 3) {
+        return tf::errors::InvalidArgument(
+            "Number of inputs is not 3 for Conv2DBackpropInput");
+      }
+
+      tf::Node *tf_filter, *tf_out_backprop;
+      TF_RETURN_IF_ERROR(op->input_node(1, &tf_filter));
+      TF_RETURN_IF_ERROR(op->input_node(2, &tf_out_backprop));
+      auto ng_filter = ng_op_map.at(tf_filter->name());
```
avijit-nervana: Need to catch the `std::out_of_range` exception if the lookup fails. See this: https://en.cppreference.com/w/cpp/container/map/at

src/ngraph_builder.cc
```diff
@@ -633,8 +633,152 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+      auto ng_out_backprop = ng_op_map.at(tf_out_backprop->name());
```
avijit-nervana: Same comment as above.
Retur tf::errors::Internal with the message detailing the cause. See other places of this file.

src/ngraph_builder.cc
```diff
@@ -633,8 +633,152 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+      TF_RETURN_IF_ERROR(
+          tf::GetNodeAttr(op->attrs(), "dilations", &tf_dilations));
+      TF_RETURN_IF_ERROR(
+          tf::GetNodeAttr(op->attrs(), "padding", &tf_padding_type));
+      TF_RETURN_IF_ERROR(
+          tf::GetNodeAttr(op->attrs(), "data_format", &tf_data_format));
+
+      if (tf_data_format != "NHWC" && tf_data_format != "NCHW") {
+        return tf::errors::InvalidArgument(
+            "Conv2DBackpropInput data format is neither NHWC nor NCHW");
```
avijit-nervana: Also append what specific data_format passed in i.e., the value of `tf_data_format`.

src/ngraph_builder.cc
```diff
@@ -633,8 +633,152 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+                           static_cast<unsigned long>(tf_input_sizes[1]),
+                           static_cast<unsigned long>(tf_input_sizes[2]),
+                           static_cast<unsigned long>(tf_input_sizes[3]) };
+      }
+
+      NGRAPH_VLOG(3) << "ng_strides: " << ng::join(ng_strides);
+      NGRAPH_VLOG(3) << "ng_dilations: " << ng::join(ng_dilations);
+      NGRAPH_VLOG(3) << "ng_image_shape: " << ng::join(ng_image_shape);
+
+      {
```
avijit-nervana: Any specific reason for this new scope?

test/python/test_conv2dbackpropinput.py
```diff
@@ -0,0 +1,111 @@
...
+        t3 = constant_op.constant(x1, shape=out_backprop_in_sizes)
+        inp = nn_ops.conv2d_backprop_input(
+            t1, t2, t3,
+            strides=[1, 1, 2, 2], padding=padding, data_format='NCHW')
+        value = sess.run(inp)
+
+    # To validate on the CPU side we will need to run in NHWC, because the CPU
+    # implementation of conv/conv backprop does not support NCHW. We will
+    # transpose on the way in and on the way out.
+    with self.session as sess:
```
avijit-nervana: It's better to run this explicitly on CPU i.e., with device('/device:CPU'). This is from readability point of view (functionally it's just fine as is).

src/ngraph_builder.cc
megawidget: You'd think I would know about this, having done it before :|

src/ngraph_builder.cc
megawidget: it's copypasta from elsewhere, but I believe it's to not leak the `auto &s` outside, given that it's a bad but convenient variable name.

avijit-nervana: I agree that there won't be any leak - but for reusing a variable name let's not introduce a new scope. This kind of scopes are best used for temporary locking/unlocking.

aprocter: The copypasta is from code I wrote. I can't remember why I did the new scope, to be honest.

megawidget: It's because you're reusing variable names, which makes sense to me in context.  Can we keep it?

aprocter: Oh! You're right. :) (Didn't spot/had forgotten the other use of `s` further up in the file.) I do agree that the nested scope makes sense here, but don't feel strongly about it.

avijit-nervana: Yes - I would recommend removing the scope and renaming the variable :)

megawidget: It would then deviate from the rest of the code.  Can I fix this in refactoring instead?

avijit-nervana: No - let's not combine the two.

PR #72, opened 2018-07-09
more on slice and strided slice

wenzhe-nrv:
current tested cases
test cases:
```
a[:]
a[...]
a[:, :]
a[:, ...]
a[1:, :-2]
a[::2, :-2]
a[1, :]
a[0]
a[0][1]
a[-1]
```
add more debug log, todo clean some before merge


src/ngraph_builder.cc
```diff
@@ -1716,20 +1727,38 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
       }
       auto lower_vec = ng_begin_const->get_vector<int>();
 
-      auto ng_size_const =
-          std::dynamic_pointer_cast<ng::op::Constant>(ng_size->second);
-      if (ng_size_const == nullptr) {
+      auto ng_end_const =
+          std::dynamic_pointer_cast<ng::op::Constant>(ng_end->second);
```
aprocter: We need to be storing this information as an attribute during the confirmation pass, rather than casting nodes to `Constant`.

src/ngraph_builder.cc
```diff
@@ -1716,20 +1727,38 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
-      NGRAPH_VLOG(3) << "Size input for StridedSlice: " << ng::join(size_vec);
-      if (std::any_of(size_vec.begin(), size_vec.end(), [](int i){ return i <= 0; })) {
-        std::transform(size_vec.begin(), size_vec.end(), input_shape.begin(),
-                       size_vec.begin(), [](int first, int second) {
+      NGRAPH_VLOG(3) << "End input for StridedSlice: " << ng::join(end_vec);
+
+      auto& input_shape = ng_input->second->get_shape();
+      NGRAPH_VLOG(3) << "Input shape for StridedSlice: " << ng::join(input_shape);
+
+      if (tf_shrink_axis_mask != 0) {
```
aprocter: From what I can tell from the docs, `shrink_axis_mask` is a bit vector where a "1" indicates that an axis, having been chopped down to one dimension, should be removed from the output shape.

diyessi: Can we keep the stuff like -1 means from the end kind of stuff out of our ops and let the python bindings deal with all the creeping featurism in the python versions?

aprocter: @diyessi Sorry, missed this. For sure---this is the point where we translate's TensorFlow's fancy ops to nGraph's RISC-y ones.

src/ngraph_builder.cc
aprocter: We'll still need to make this change. Take a look at `Reshape`'s confirmation function/builder case for one example of the pattern we need here.

avijit-nervana:
@wenzhe-nrv Is this done?

wenzhe-nrv:
@avijit-nervana I think this is done. The bug in transformer could be a different ticket and PR. I'm looking into that.

avijit-nervana:
@aprocter Could you please review this again as the changes are in place? Thanks.

PR #74, opened 2018-07-11
[Do NOT merge] add gpu support

fengleitian:
This PR is a temp solution to enable gpu_backend. Not intend to merge to master.
* it will fixed backend to GPU;
* it will pass tf_exec.axpy;
* currently will get segmentation fault during tear down, might be some bugs.


ransford2011:
@avijit-nervana @fengleitian .  I think we can close this now due #134

avijit-nervana:
I agree. Closing this.

PR #79, opened 2018-07-13
sanitize cmake build type

megawidget:
Note that this does *not* work with python tests.
It's as though python was not compiled with `-fsanitize` :trollface:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,11 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 include(ExternalProject)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-comment -Wno-sign-compare")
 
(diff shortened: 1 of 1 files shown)


megawidget:
BTW I tried using LD_PRELOAD to address it not working with python and that completely and utterly failed to work.  If someone knows more about this I'll be happy to fix that.  It's still useful because it works with the C++ tests, and is entirely optional.

CMakeLists.txt
```diff
@@ -22,6 +22,28 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake")
 include(ExternalProject)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-comment -Wno-sign-compare")
 
+SET(CMAKE_CXX_FLAGS_SANITIZE
```
avijit-nervana: Can you please put a comment block indicating what this flag does and how to use it? Thanks.

megawidget: :+1:

PR #81, opened 2018-07-16
Langjian/batch mat mul

jianyinglang:
Implemented BatchMatMul for 2D, 3D, and general multi-dimensions; Add the C++ unit tests.


test/tf_exec.cpp
```diff
@@ -104,4 +103,82 @@ TEST(tf_exec, axpy) {
...
+void AssertTensorEquals(tf::Tensor T1, tf::Tensor T2) {
+  auto T_size = T1.flat<float>().size();
+  for (int k=0; k<T_size; k++) {
+    auto a = T1.flat<float>().data()[k];
+    auto b = T2.flat<float>().data()[k];
+    EXPECT_FLOAT_EQ(a, b);
+  } 
+}
+
+TEST(tf_exec, BatchMatMul) {
```
aprocter: Let's add some more unit tests for some trickier cases, specifically ones where the lengths of the batch axes don't happen to match either of the matrix dimensions. (For example, try `{2,3,4,5}` with `{2,3,5,4}`. Probably will want to generate the coefficients in a loop, of course. :))

aprocter: Also, we should run some weird corner cases like ones where the batch or matrix sizes are zero. For example, both of these should work:

```
>>> from tensorflow.python.ops.gen_math_ops import batch_mat_mul

>>> x = tf.constant([],shape=[2,0,2,2])
>>> y = tf.constant([],shape=[2,0,2,2])
>>> tf.Session().run(batch_mat_mul(x,y))
array([], shape=(2, 0, 2, 2), dtype=float32)

>>> x = tf.constant([],shape=[2,3,2,0])
>>> y = tf.constant([],shape=[2,3,0,2])
>>> tf.Session().run(batch_mat_mul(x,y))
array([[[[0., 0.],
         [0., 0.]],

        [[0., 0.],
         [0., 0.]],

        [[0., 0.],
         [0., 0.]]],


       [[[0., 0.],
         [0., 0.]],

        [[0., 0.],
         [0., 0.]],

        [[0., 0.],
         [0., 0.]]]], dtype=float32)
```

src/ngraph_builder.cc
```diff
@@ -314,6 +314,136 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+ 
+      tf::Node* tf_lhs;
+      tf::Node* tf_rhs;
+      TF_RETURN_IF_ERROR(op->input_node(0, &tf_lhs));
+      TF_RETURN_IF_ERROR(op->input_node(1, &tf_rhs));
+  
+      try {
+        ng_op_map.at(tf_lhs->name());
+      } catch(const std::out_of_range&) {
+          return tf::errors::NotFound(tf_lhs->name(),"is not found in ng_op_map!");
```
aprocter: Please remove the exclamation points from the error messages.

src/ngraph_builder.cc
```diff
@@ -314,6 +314,136 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+          return tf::errors::InvalidArgument(
+              "ng_lhs_shape and ng_rhs_shape must be the same for BatchMatMul for each dimension",
+               i);
+        }
+        out_axes.push_back(i);
+      }
+
+      bool tf_adj_x = false;
+      bool tf_adj_y = false;
+      TF_RETURN_IF_ERROR(tf::GetNodeAttr(op->attrs(), "adj_x", &tf_adj_x));
```
aprocter: Let's add some unit tests for cases where `adj_x` and/or `adj_y` are true.

jianyinglang: Added the unit test for adj_x and adj_y

src/ngraph_builder.cc
```diff
@@ -314,6 +314,136 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+      ng_rhs_shape = ng_rhs->get_shape();
+ 
+      if(ng_lhs_shape[n_dims-1] != ng_rhs_shape[0]) {
+        return tf::errors::InvalidArgument(
+            "The last dimension of ng_lhs and the first dimension of ng_rhs should have the same size!"
+            );
+      }
+      if (n_dims == 2) {
+        ng_op_map[op->name()] = make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs);
+      } else if (n_dims == 3) {
```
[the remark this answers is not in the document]
aprocter: We can still insert the reshapes/slices conditionally. There seems to be a lot of duplicated logic between `n_dims==2` and `n_dims==3`.

test/tf_exec.cpp
aprocter: The new tests look good! However, I think it is probably better to split them up into separate tests for the individual cases.

test/tf_exec.cpp
```diff
@@ -106,6 +105,149 @@ TEST(tf_exec, axpy) {
...
+  }
+  tf::Tensor Y1(tf::DT_FLOAT, tf::TensorShape({2, 3, 4, 5}));
+  auto Y1_flat = Y1.flat<float>();
+  for (int i = 0; i < Y1_flat.size(); i++) {
+    Y1_flat.data()[i] = -0.5f*i;
+  }
+
+  tf::Tensor X2(tf::DT_FLOAT, tf::TensorShape({2, 0, 4, 5}));
+  auto X2_flat = X2.flat<float>();
+  for (int i = 0; i < X2_flat.size(); i++) {
```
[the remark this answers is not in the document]
jianyinglang: Removed that

test/tf_exec.cpp
jianyinglang: Split the test cases to 0D, 2D, 3D and general

src/ngraph_builder.cc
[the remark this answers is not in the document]
jianyinglang: Removed the `n_dims==3` case

test/tf_exec.cpp
```diff
@@ -106,6 +105,180 @@ TEST(tf_exec, axpy) {
   }
 }
 
+void AssertTensorEquals(tf::Tensor T1, tf::Tensor T2) {
```
avijit-nervana: Good custom comparator!

src/ngraph_builder.cc
```diff
@@ -361,6 +361,120 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+            "The last dimension of ng_lhs and the first dimension of ng_rhs should have the same size"
+            );
+      }
+      if (n_dims == 2) {
+        ng_op_map[op->name()] = make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs);
+      } else {
+        auto output_type = ng_lhs->get_element_type(); 
+        auto output_shape = ng_lhs_shape;
+        output_shape[n_dims-1] = ng_rhs_shape[1];
+        auto output_tensor = make_shared<ngraph::op::Parameter>(output_type, output_shape);
```
aprocter: This is unused, so let's delete it.

src/ngraph_builder.cc
```diff
@@ -361,6 +361,120 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+          compound_size *= output_shape[i];
+        }
+        auto dot_axes = out_axes;
+        dot_axes.push_back(n_dims-2);
+        dot_axes.push_back(n_dims-1);
+        for (int i=0; i<out_axes.size(); i++) {
+          dot_axes.push_back(n_dims+i);
+        }
+        ng::Shape dot_shape = {compound_size, ng_lhs_shape[n_dims-2], ng_rhs_shape[1], compound_size};
+        auto dot_reshape = make_shared<ngraph::op::Reshape>(dot_output, dot_axes, dot_shape);
```
aprocter: This reshape should be conditional on `n_dims > 3`.

src/ngraph_builder.cc
```diff
@@ -361,6 +361,120 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
+        vector<shared_ptr<ngraph::Node>> tmp_tensors;
+        for (size_t i = 0; i < dot_shape[0]; i++) { 
+          const std::vector<size_t> lower_bound{i, 0, 0, i};
+          const std::vector<size_t> upper_bound{i+1, dot_shape[1], dot_shape[2], i+1};
+          auto slice_out = make_shared<ngraph::op::Slice>(dot_reshape, lower_bound, upper_bound);
+          auto reshape_out = make_shared<ngraph::op::Reshape>(slice_out, ng::AxisVector{0, 1, 2, 3}, tmp_shape);
+          tmp_tensors.push_back(reshape_out);
+        }
+        auto concat_op = make_shared<ngraph::op::Concat>(tmp_tensors, 0);
+        ng_op_map[op->name()] = make_shared<ngraph::op::Reshape>(concat_op, ng::AxisVector{0, 1, 2 }, output_shape);
```
aprocter: This reshape also should be conditional on `n_dims > 3`.

src/ngraph_builder.cc
jianyinglang: deleted `output_type` and `output_tensor`

src/ngraph_builder.cc
jianyinglang: Added the condition for `dot_reshape` only applies for `n_dims > 3` case

src/ngraph_builder.cc
jianyinglang: Added the condition for reshape `concat_op` only for n_dims > 3 case

aprocter:
Thanks for the changes, think we're ready to roll now!

PR #83, opened 2018-07-17
minimum OP

megawidget:
bonus: cleaned up tests to use `import ngraph` and fixed the mac build :trollface:

### DIFF
--- a/python/CreatePipWhl.cmake
+++ b/python/CreatePipWhl.cmake
@@ -57,6 +57,30 @@ if (PYTHON)
             install_name_tool -change 
             libngraph.0.4.dylib 
             @loader_path/libngraph.0.4.dylib 
+            ${CMAKE_CURRENT_BINARY_DIR}/python/ngraph/libcpu_backend.dylib
+            RESULT_VARIABLE result
+            ERROR_VARIABLE ERR
+            ERROR_STRIP_TRAILING_WHITESPACE
+        )
+        if(${result})
+            message(FATAL_ERROR "Cannot update @loader_path")
+        endif()
+        execute_process(COMMAND 
+            install_name_tool -change 
+            libcodegen.0.4.dylib 
+            @loader_path/libcodegen.0.4.dylib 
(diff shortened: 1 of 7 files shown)


python/CreatePipWhl.cmake
```diff
@@ -53,6 +53,30 @@ if (PYTHON)
         # 1. Get the NGRAPH_VERSION exported from nGraph CMake
         # 2. If #1 isn't possible, then determine this by looking at the 
         # dependency list?
+        execute_process(COMMAND
```
aprocter: Looks like this command is repeated below (just past the greentext).

test/python/conftest.py
```diff
@@ -22,12 +21,7 @@
 
 from tensorflow.python.framework import ops
 
-from common import LIBNGRAPH_DEVICE
-
-
-@pytest.fixture(scope='session', autouse=True)
-def load_ngraph_device():
-  cdll.LoadLibrary(os.path.join('../../src', LIBNGRAPH_DEVICE))
+import ngraph  # noqa
```
avijit-nervana: We cannot do this at the unit test level. The unit tests must not require the Python package (which is built using `make install` that logically comes after running the tests and ensuring they all pass. Please revert his change.

megawidget: Ack.  Okay.

python/CreatePipWhl.cmake
megawidget: Yes, with different arguments. :|

aprocter: Ooh, there it is. `libcpu_backend` vs `libngraph_device`. Okey dokey.

avijit-nervana: While you are at it - would you mind parameterize the version numbers i.e., `.0.4.0` --> NGRAPH_REL_VERSION or something. For now we will hardcode this but in future - we will figure out a way to get this from nGraph programmatically. Thanks :)

avijit-nervana: Actually nevermind. I am doing a PR to update the nGraph version - I will do this as part of that.

PR #84, opened 2018-07-17
Add Square/Rsqrt; fix for dilated conv

aprocter:
Two fixes related to SSD-VGG16. (1) Add Square/Rsqrt ops. (2) Fix computation of the needed padding for dilated conv.


src/ngraph_builder.cc
```diff
@@ -69,11 +68,16 @@ static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map) {
   tf::Node* tf_input;
   TF_RETURN_IF_ERROR(op->input_node(0, &tf_input));
   auto ng_input = ng_op_map.at(tf_input->name());
```
avijit-nervana: Neet to catch the exception when the entry doesn't exist in the map.

src/ngraph_builder.cc
```diff
@@ -59,8 +59,7 @@ static tf::Status MakeConstOp(tf::Node* op, ng::element::Type et,
   return tf::Status::OK();
 }
 
-template <typename T>
-static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map) {
+static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map, std::function<std::shared_ptr<ng::Node>(std::shared_ptr<ng::Node>)> f) {
```
avijit-nervana: `f` is a bad parameter name. Please specify something more descriptive. Also provide an example usage.
Also the line is too long - didn't clang-format fix it?

src/ngraph_builder.cc
```diff
@@ -1730,6 +1740,19 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
     }
+    // -----
+    // Rsqrt
+    // -----
+    else if (op->type_string() == "Rsqrt") {
+      TF_RETURN_IF_ERROR(TranslateUnaryOp(op, ng_op_map,
+         [](std::shared_ptr<ng::Node> n) {
+           auto et = n->get_element_type();
+           auto shape = n->get_shape();
+           std::vector<std::string> constant_values(ng::shape_size(shape),"-0.5");
```
avijit-nervana: Short explanation would be good here (especially the magic number -0.5)

test/tf_exec.cpp
```diff
@@ -131,4 +131,46 @@ TEST(tf_exec, Op_RealDiv) {
 
 }
 
+TEST(tf_exec, Op_Square) {
```
avijit-nervana: C++ Unit test - nice!
(We will have to start reorganizing these tests - but that's a separate PR)

PR #86, opened 2018-07-18
Wenzhe/expand dims

wenzhe-nrv:
SSD inference should run with this PR. 
@sharathns93 could you please verify on your side too?


src/ngraph_builder.cc
```diff
@@ -1172,7 +1172,7 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
       auto& shape = ng_input->second->get_shape();
       auto shape_size = shape.size();
       if (dim_vec[0] < 0) {
-        dim_vec[0] = shape_size + dim_vec[0];
+        dim_vec[0] = shape_size + dim_vec[0] + 1;
```
avijit-nervana: Is it possible to add some comment about the magic number `1` (this is for the benefit of future readers.

PR #87, opened 2018-07-18
Avijit/update ngraph v.5.0

avijit-nervana:


crlishka:
@avijit. Simply uncommenting the resent50 test in CI is not enough to get it running.  We still need to figure out where to put the ImageNet data, as CI precommit checks run in a docker container, and /nfs directories in Intel’s environment do not work in docker.  (See our conversation with Sharath about this.)

avijit-nervana:
TESTNOW

PR #92, opened 2018-07-20
refactored input validation

megawidget:
also, removed some stray-seeming prints

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -44,8 +44,17 @@ const static std::map<tf::DataType, ngraph::element::Type> TF_NGRAPH_TYPE_MAP =
 static tf::Status ValidateInputCount(const tf::Node* op, size_t count) {
   if (op->num_inputs() != count) {
     return tf::errors::InvalidArgument(
-        op->name(), " requires ", count, " input(s), got ", op->num_inputs(),
-        " instead");
+        "\"", op->name(), "\" requires ", count, " input(s), got ",
+        op->num_inputs(), " instead");
+  }
+  return tf::Status::OK();
+}
+
+static tf::Status ValidateInputCountMin(const tf::Node* op, size_t count) {
+  if (op->num_inputs() < count) {
+    return tf::errors::InvalidArgument(
+        "\"", op->name(), "\" requires at least ", count, " input(s), got ",
+        op->num_inputs(), " instead");
   }
   return tf::Status::OK();
 }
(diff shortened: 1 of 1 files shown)


src/ngraph_builder.cc
```diff
@@ -41,6 +41,15 @@ const static std::map<tf::DataType, ngraph::element::Type> TF_NGRAPH_TYPE_MAP =
      {tf::DataType::DT_UINT16, ng::element::u16},
      {tf::DataType::DT_BOOL, ng::element::boolean}};
 
+static tf::Status ValidateInputCount(const tf::Node* op, size_t count) {
+  if (op->num_inputs() != count) {
+    return tf::errors::InvalidArgument(
+        op->name(), " requires ", count, " input(s), got ", op->num_inputs(),
```
avijit-nervana: This is very good error message - descriptive - so that users know what to do. Only suggestion would be to provide \" around the op name.

src/ngraph_builder.cc
```diff
@@ -572,7 +562,8 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
     else if (op->type_string() == "ConcatV2") {
       if (op->num_inputs() < 2) {
         return tf::errors::InvalidArgument(
-            "Number of inputs is not at least 2 for ConcatV2");
+            op->name(), " requires at least 2 inputs, got ", op->num_inputs(),
```
avijit-nervana: Please consider adding another ValidateInputCount(...) which takes a min and max and returns error if the input count is not within the range. Then we can use that here and other ops where there are at least N inputs needed or less than M inputs. Can you think of a lambda as a predicate? Or just use simple overloads.

src/ngraph_builder.cc
```diff
@@ -1612,11 +1575,10 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
     // Pack
     // -------
     else if (op->type_string() == "Pack") {
-
-      std::cout << "Number of inputs : " << op->num_inputs() << std::endl;
       if (op->num_inputs() < 1) {
         return tf::errors::InvalidArgument(
-            "Number of inputs should be at least 1 for Pack");
+            op->name(), " requires at least 1 input, got ", op->num_inputs(),
```
avijit-nervana: The same new suggested validation function in line 565 may be used here as well.

src/ngraph_builder.cc
megawidget: I'll go with simple overloads, because the error message is different.

avijit-nervana:
Let's wait for the CI to finish and I will merge this.

crlishka:
TESTNOW

PR #93, opened 2018-07-20
Implement Op SquaredDifference

shresthamalik:
Implemented Op SquaredDifference.


src/ngraph_builder.cc
```diff
@@ -119,7 +119,73 @@ static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map) {
...
 }
 
-// Helper for Builder::TranslateGraph (elementwise binops)
+// Helper function to translate a binary op
+// Parameters:
+//
+//    tf::Node* op               - TF op being translated. Must have only two inputs.
+//    Builder::OpMap& ng_op_map  - The TF-to-nGraph op map.
+//
+//    std::function<std::shared_ptr<ng::Node>(std::shared_ptr<ng::Node>>
```
aprocter: The signature here is different from `Unary` (two args)

src/ngraph_builder.cc
```diff
@@ -119,7 +119,71 @@ static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map) {
...
+//         [](std::shared_ptr<ng::Node> ng_input1, std::shared_ptr<ng::Node> ng_input2) {
+//           auto ng_diff = std::make_shared<ng::op::Subtract>(input1, input2);
+//           return std::make_shared<ng::op::Multiply>(ng_diff,ng_diff);
+//         }));
+//    }
+//
+
+static tf::Status TranslateBinaryOp(
+    tf::Node* op, Builder::OpMap& ng_op_map,
+    std::function<std::shared_ptr<ng::Node>(std::shared_ptr<ng::Node>, std::shared_ptr<ng::Node>)>
```
avijit-nervana: It's a a good use of `lambda` (and you followed the `TranslateUnaryOp(...)` example which is good). Just the name of the variable is a bit confusing. Instead of calling it `build_graph` consider some other name such as `create_binary_op`. 

Also change the name of the variable that holds the lambda in the `TranslateUnaryOp(...)` as well to make it consistent.

src/ngraph_builder.cc
```diff
@@ -119,7 +119,71 @@ static tf::Status TranslateUnaryOp(tf::Node* op, Builder::OpMap& ng_op_map) {
   return TranslateUnaryOp(op, ng_op_map, [](std::shared_ptr<ng::Node> n) { return make_shared<T>(n); });
 }
 
-// Helper for Builder::TranslateGraph (elementwise binops)
+// Helper function to translate a binary op
```
avijit-nervana: expand this sentence to something like: "that requires combination of multiple nGraph ops"

test/tf_exec.cpp
```diff
@@ -325,6 +325,30 @@ TEST(tf_exec, Op_Square) {
...
+
+  TF_CHECK_OK(session.Run({{A, {{3.f, 5.f}, {2.f, 0.f}}}, {B, {{1.f, 2.f}, {-1.f, 1.f}}}}, {r}, &outputs)); 
+  ASSERT_EQ(outputs[0].shape(), tf::TensorShape({2,2}));
+
+  auto mat = outputs[0].matrix<float>();
+  EXPECT_FLOAT_EQ(4.0, mat(0, 0));
+  EXPECT_FLOAT_EQ(9.0, mat(0, 1));
+  EXPECT_FLOAT_EQ(9.0, mat(1, 0));
+  EXPECT_FLOAT_EQ(1.0, mat(1, 1));
+
```
avijit-nervana: Are there any corner cases we need to test here? Probably not but just checking.

src/ngraph_builder.cc
shresthamalik: Agreed. Have changed as suggested.

avijit-nervana: Looks like you haven't pushed the changes?

src/ngraph_builder.cc
shresthamalik: This function can be used for any binary op. I made changes in the other TranslateBinaryOp() function, to remove the redundant code

test/tf_exec.cpp
shresthamalik: I have passed other numbers and done some sanity checks for this Op. Nothing that needs to be here.

PR #94, opened 2018-07-20
Igor/reshape refactor

megawidget:


src/ngraph_builder.cc
```diff
@@ -41,6 +41,66 @@ const static std::map<tf::DataType, ngraph::element::Type> TF_NGRAPH_TYPE_MAP =
      {tf::DataType::DT_UINT16, ng::element::u16},
      {tf::DataType::DT_BOOL, ng::element::boolean}};
 
+template<size_t a, size_t b, size_t c, size_t d>
+static void Reshape(std::shared_ptr<ng::Node>& ng_node) {
```
avijit-nervana: This is very cool!

Add a unit test.

src/ngraph_builder.cc
```diff
@@ -41,6 +41,66 @@ const static std::map<tf::DataType, ngraph::element::Type> TF_NGRAPH_TYPE_MAP =
...
+template <typename... Arguments, typename T>
+static void NhwcToNgraph(const std::vector<T>& source,
+                         std::vector<size_t>& dest,
+                         Arguments&&... remaining) {
+  dest[0] = source[1];
+  dest[1] = source[2];
+  NhwcToNgraph(remaining...);
+}
+template <typename... Arguments>
+static void NhwcToNgraph(std::shared_ptr<ng::Node>& ng_node,
```
avijit-nervana: While Variadic templates are cool, not sure why we need it here. The main advantage of a Variadic template is to be able to cascade computations - but in this case the last stage is a no-op (i.e., function in line 57). Please remove the variadic part. 

Then add a few C++ unit tests with various corner cases.

megawidget: Maybe I'm missing something but that's how variadic functions work -- you need a terminal condition.

I'll add unit tests though.

avijit-nervana: Yes - I am not confused about how the variadic templates work - my point is that I don't see the need for it in this use case.

megawidget: What would it look like otherwise, then.

src/ngraph_conversions.h
```diff
@@ -0,0 +1,80 @@
...
+template <typename... Arguments, typename T>
+void NchwToNgraph(const std::vector<T>& source, std::vector<size_t>& dest,
+                  Arguments&&... remaining) {
+  dest[0] = source[2];
+  dest[1] = source[3];
+  NchwToNgraph(remaining...);
+}
+
+template <typename... Arguments>
+void TensorflowToNgraph(bool is_nhwc, std::shared_ptr<ng::Node>& ng_input,
```
aprocter: So I guess this function is doing two things:

(1) reshaping the node, which may be in NHWC, to NCHW
(2) extracting the "2D" nGraph-flavored parameters used by batched ops from the "4D" TensorFlow-flavored parameters (where "parameters" means strides, dilation, pooling window shape, ...).

The variadic template stuff is cool but I think ultimately this will be easier to read if we do something like this:

```
ng_input = BatchToNGraph(is_nhwc, ng_input);
ng_strides = BatchedOpParamToNGraph(is_nhwc, tf_strides);
ng_dilations = BatchedOpParamToNGraph(is_nhwc, tf_dilations);
...
```

src/ngraph_conversions.h
```diff
@@ -0,0 +1,80 @@
...
+
+#pragma once
+
+#include "ngraph_builder.h"
+#include "ngraph_log.h"
+
+namespace ngraph_bridge {
+
+template<size_t a, size_t b, size_t c, size_t d>
+void Reshape(std::shared_ptr<ng::Node>& ng_node) {
```
aprocter: I think it's better to return a `shared_ptr` here than it is to mutate through the reference. There may be situations where we want to hang on to the original node as well.

I feel like this needs a more descriptive name. Maybe something like `NGraphDimShuffle`.

Would be cool if we could get this for dimension counts ranging from 0 (yup, really) up to, I dunno, about 6 or 7, along with a dynamic version (one that just takes a vector for the permutation and does the checks at runtime). No need to do it in this PR, though.

megawidget: Hmmm, so far every use case in the code just mutates the existing node.

src/ngraph_conversions.h
megawidget: Yup.

It's quite a bit more typing though, no?  I'm not sure how it's easier to read either.

src/ngraph_conversions.h
megawidget: I'm ok with renaming it -- @avijit-nervana any thoughts? -- except i dislike `Shuffle` in the name as it kind of implies randomness to me.

src/ngraph_conversions.h
avijit-nervana: Simple statements are easy to understand and maintain. Please use the suggested functions by Adam.

megawidget: @aprocter is there a consensus on this? I updated it to use `std::pair` for clarity.  I'm willing to change to avijit's suggestion but ideally this would be the last rewrite.

src/ngraph_builder.cc
```diff
@@ -1055,49 +964,22 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
-
-        ng_dilations[0] = tf_dilations[0];
-        ng_dilations[1] = tf_dilations[1];
-
-        ng_image_shape[0] = s[2];
-        ng_image_shape[1] = s[3];
-      }
+      BatchToNGraph(is_nhwc, ng_input);
+      BatchedOpParamToNGraph(is_nhwc, ng_input->get_shape(), ng_image_shape);
+      // TODO is the below wrong?
```
megawidget: Can someone comment on this?

aprocter: It doesn't look right to me. I believe the pattern should be the same as for the shape.

PR #96, opened 2018-07-23
padding refactor

megawidget:


src/ngraph_builder.cc
```diff
@@ -59,6 +59,38 @@ static tf::Status ValidateInputCountMin(const tf::Node* op, size_t count) {
   return tf::Status::OK();
 }
 
+template <typename T>
+static void MakePadding(const std::string& tf_padding_type,
```
avijit-nervana: Make this a static member of the NGraphBuilder class so that you can add unit test to specifically test this function. 

Then add a C++ unit test to ensure the correctness of this function with various corner cases. The layout of the test would be something like this:
1. Call this  function with various combination of inputs and validate the output
2. Call this function with different supported data types.

megawidget: 👍

src/ngraph_builder.cc
```diff
@@ -970,31 +926,8 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
...
-          size_t padding_lhs = padding_needed / 2;
-          size_t padding_rhs = padding_needed - padding_lhs;
-          ng_padding_below[i] = padding_lhs;
-          ng_padding_above[i] = padding_rhs;
-        }
-      }
-
-      NGRAPH_VLOG(3) << "ng_padding_below: " << ng::join(ng_padding_below);
-      NGRAPH_VLOG(3) << "ng_padding_above: " << ng::join(ng_padding_above);
+      MakePadding(tf_padding_type, ng_image_shape, ng_kernel_shape, ng_strides,
```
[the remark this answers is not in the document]
megawidget: Of course :( I'll look into it.

PR #98, opened 2018-07-24
standardized getting nodes from map

megawidget:
Does this need unit tests? It seems fairly trivial.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -30,6 +30,19 @@
 
 namespace ngraph_bridge {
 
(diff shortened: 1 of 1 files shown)


aprocter:
#82 includes a similar solution for this that also handles the issue of multi-output ops, so I think it'd be best to just go with that.

megawidget:
I suppose if you don't want your outputs named.

avijit-nervana:
Looks like #82 renders this one moot - no?

megawidget:
I suppose so. :(

PR #100, opened 2018-07-25
standardized some confirmation functions

megawidget:
also, some stuff got autoformatted by clang-format -- sorry!

### DIFF
--- a/src/ngraph_confirm_pass.cc
+++ b/src/ngraph_confirm_pass.cc
@@ -21,6 +21,20 @@
 
 using namespace std;
 namespace ngraph_bridge {
+namespace detail {
+// yes, these have to be here
+// defined in a function or even in a class the symbols cannot be found for
+// template resolution
+static const char FILL_DIMS[] = "_ngraph_fill_static_dims";
+static const char MEAN_REDUCTION_AXES[] = "_ngraph_mean_static_axes";
+static const char PAD_PADDING_WIDTHS[] = "_ngraph_pad_static_paddings";
+static const char PROD_REDUCTION_AXES[] = "_ngraph_prod_static_axes";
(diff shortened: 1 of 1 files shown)


avijit-nervana:
@megawidget Please fix the compilation errors (see the CI failures) and update the PR.

src/ngraph_confirm_pass.cc
```diff
@@ -21,6 +21,20 @@
 
 using namespace std;
 namespace ngraph_bridge {
+namespace detail {
+// yes, these have to be here
+// defined in a function or even in a class the symbols cannot be found for
+// template resolution
+static constexpr char FILL_DIMS[] = "_ngraph_fill_static_dims";
```
aprocter: While we're here, can we go ahead and tweak `ngraph_builder.cc` to use these constants as well?

megawidget: That's a great idea. Why didn't I think of that!

src/ngraph_builder.cc
```diff
@@ -1736,6 +1743,8 @@ tf::Status Builder::TranslateGraph(const std::vector<tf::TensorShape>& inputs,
       TF_RETURN_IF_ERROR(GetInputNode(ng_op_map, op, 0, &ng_input));
 
       ng::AxisSet ng_axis_set;
+      // TODO: what is happening here? this is not using the const attr
```
aprocter: Strange, and a good catch. This attribute plays the same role as it does in `Sum` and `Mean`. Think we might have just missed this one when we made the ol' upgrade. If you want to fix in this PR that'd be swell, if not please just change this `TODO` to something like "should get reduction axis from op attribute".

FWIW, something like a `TranslateReductionOp`, analogous to `TranslateBinaryOp`, could unify those three ops here. Probably a separate PR.

PR #104, opened 2018-07-26
removed unused return value spam

megawidget:
also why were objects passed by value?

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -73,12 +73,11 @@ static tf::Status ValidateInputCountMin(const tf::Node* op, size_t count) {
 //    shared_ptr<ng::Node> output_node - ng::Node to store
 //
 
-static tf::Status SaveNgOp(Builder::OpMap& ng_op_map, const std::string op_name,
-                           const shared_ptr<ng::Node> output_node) {
+static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
+                     const shared_ptr<ng::Node>& output_node) {
   // no need to try-catch, map[key] will create vector object
   // if not exists
   ng_op_map[op_name].push_back(output_node);
-  return tf::Status::OK();
 }
 
 // Helper for fetching correct input node from ng_op_map.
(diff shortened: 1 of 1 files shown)


src/ngraph_builder.cc
```diff
@@ -73,12 +73,11 @@ static tf::Status ValidateInputCountMin(const tf::Node* op, size_t count) {
...
 //
 
-static tf::Status SaveNgOp(Builder::OpMap& ng_op_map, const std::string op_name,
-                           const shared_ptr<ng::Node> output_node) {
+static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
+                     const shared_ptr<ng::Node>& output_node) {
   // no need to try-catch, map[key] will create vector object
   // if not exists
   ng_op_map[op_name].push_back(output_node);
-  return tf::Status::OK();
```
kbalka: I kept this line as information for someone reading this code, that this function will always succeed, in case someone would think about try catching it etc.. 
Just for curiosity searched tensorflow code for this pattern, and found similar thing
https://github.com/tensorflow/tensorflow/blob/e1e5f305e5359fd50340ea76ea2f737f6e87a0d7/tensorflow/core/distributed_runtime/message_wrappers.cc#L792
But anyway, feel free to make this function returning `void`, for me both options are acceptable

src/ngraph_builder.cc
```diff
@@ -73,12 +73,11 @@ static tf::Status ValidateInputCountMin(const tf::Node* op, size_t count) {
 //    shared_ptr<ng::Node> output_node - ng::Node to store
 //
 
-static tf::Status SaveNgOp(Builder::OpMap& ng_op_map, const std::string op_name,
-                           const shared_ptr<ng::Node> output_node) {
+static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
+                     const shared_ptr<ng::Node>& output_node) {
```
kbalka: Good catch with `op_name`. In terms of `output_node`, it's in fact a pointer so there shouldn't be much difference
(tensorflow code has both passing by ref and by value).  Ok to change it.

PR #105, opened 2018-07-27
using table dispatch for translate graph

megawidget:
Wow the diff is whack.

I've only moved code into separate functions, and made things constant that should've been constant.


megawidget:
OK, it's done. :( . Git really didn't know how to handle the diff so I had to merge every commit in master entirely by hand, hence the number of merge commits.

PR #109, opened 2018-08-02
Langjian/avg pool grad

jianyinglang:
Implemented AvgPoolGrad op


src/ngraph_builder.cc
```diff
@@ -374,6 +374,98 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
...
+  NGRAPH_VLOG(3) << tf_data_format;
+
+  ng::Shape ng_forward_arg_shape;
+  ng::Strides ng_strides(2);
+  ng::Shape ng_image_shape(2);
+  ng::Shape ng_kernel_shape(2);
+
+  if (tf_orig_input_shape_vec.size() != 4) {
+    return tf::errors::InvalidArgument(
+        "Tensorflow requires input tensor shape being 4", tf_orig_input_shape_vec.size());
```
aprocter: Is this actually required by TF or is it just a limitation in our implementation for now?

Looks like you're checking this in the confirmation pass so it should be okay.

src/ngraph_builder.cc
```diff
@@ -374,6 +374,98 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
...
+  }
+  for (int i = 0; i < tf_orig_input_shape_vec.size(); i++) {
+    ng_forward_arg_shape.push_back(tf_orig_input_shape_vec[i]);
+  }
+
+  if (is_nhwc) {
```
aprocter: Can we use `BatchedOpParamToNGraph` for this?

test/python/test_avgpoolbackprop.py
```diff
@@ -0,0 +1,96 @@
...
+    elif padding == "SAME":
+      grad_input = tf.placeholder(tf.float32, shape=(128, 3, 75, 224))
+
+    with self.device:
+      a = avg_pool_grad(self.forward_arg_shape_NCHW, grad_input, self.ksize, self.strides, 
+        padding=padding, data_format="NCHW")
+      with self.session as sess:
+        (result) = sess.run(a, feed_dict={grad_input : np_nchw})
+    # To validate on the CPU side we will need to run in NHWC, because the CPU
+    # implementation of conv/conv backprop does not support NCHW. We will
```
aprocter: Maybe change the word "conv" here.

src/ngraph_builder.cc
jianyinglang: This is required by TF

src/ngraph_builder.cc
```diff
@@ -374,6 +374,98 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
   return tf::Status::OK();
 }
 
+static tf::Status TranslateAvgPoolGradOp(const tf::Node* op, 
+                                     Builder::OpMap& ng_op_map) { 
+  TF_RETURN_IF_ERROR(ValidateInputCount(op, 2));
+      
+  shared_ptr<ng::Node> ng_grad;
+  TF_RETURN_IF_ERROR(GetInputNode(ng_op_map, op, 1, &ng_grad));
```
megawidget: Instead of this + `ValidateInputCount`, if you merge in latest `master` you can just do

`TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, nullptr, &ng_grad));`

src/ngraph_builder.cc
megawidget: :+1: this is also from refactor changes in `master`

src/ngraph_builder.cc
megawidget: There's probably no reason to validate this twice though is there?

src/ngraph_builder.cc
jianyinglang: Added a new helper function in `ngraph_conversions.h`for this case.

src/ngraph_builder.cc
jianyinglang: deleted it

test/python/test_avgpoolbackprop.py
jianyinglang: changed

src/ngraph_builder.cc
jianyinglang: changed

src/ngraph_builder.cc
```diff
@@ -374,6 +374,98 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
...
+    size_t tmp1 = ng_forward_arg_shape[1];
+    ng_forward_arg_shape[1] = ng_forward_arg_shape[3];
+    size_t tmp2 = ng_forward_arg_shape[2];
+    ng_forward_arg_shape[2] = tmp1;
+    ng_forward_arg_shape[3] = tmp2;
+  }
+
+  BatchToNGraph(is_nhwc, ng_grad);
+  BatchedOpParamToNGraph(is_nhwc, tf_strides, ng_strides);
+  BatchedOpParamToNGraph(is_nhwc, ng_forward_arg_shape, ng_image_shape);
```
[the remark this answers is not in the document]
jianyinglang: No,  the was from a different commit

src/ngraph_builder.cc
```diff
@@ -374,6 +374,98 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
...
+  BatchedOpParamToNGraph(is_nhwc, tf_ksize, ng_kernel_shape);
```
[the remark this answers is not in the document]
jianyinglang: It is either `nhwc` or `nchw`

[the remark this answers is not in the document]
aprocter: > tf Kernel shape is [filter_height, filter_width, in_channels, out_channels] irrespective of NCHW or NHWC. Maybe @megawidget @aprocter could confirm.

Yup, that's true.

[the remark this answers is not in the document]
aprocter: Whoops- in the case of the pooling ops I think @jianyinglang is right, actually. o_O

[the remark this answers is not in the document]
aprocter: Maybe to eliminate confusion we should call it "window shape" instead of "kernel shape"

src/ngraph_builder.cc
```diff
@@ -374,6 +374,87 @@ static tf::Status TranslateAvgPoolOp(const tf::Node* op,
...
+  ng::Shape ng_forward_arg_shape(4);
+  ng::Strides ng_strides(2);
+  ng::Shape ng_image_shape(2);
+  ng::Shape ng_kernel_shape(2);
+
+  BatchedOpParamReshape(is_nhwc, ng_orig_input_shape, ng_forward_arg_shape);
+  BatchToNGraph(is_nhwc, ng_grad);
+  BatchedOpParamToNGraph(is_nhwc, tf_strides, ng_strides);
+  BatchedOpParamToNGraph(is_nhwc, ng_forward_arg_shape, ng_image_shape);
+  BatchedOpParamToNGraph(is_nhwc, tf_ksize, ng_kernel_shape);
```
shresthamalik: ng_forward_arg_shape is in NCHW format (line 420). if data format is NHWC, ng_image_shape will not get the right value (line 423). Similarly for line 424, Ksize is FH,FW,IC,OC, irrespective of data format, ng_kernel_shape is not a function of tf_ksize and is_nhwc.

jianyinglang: I am not sure I understand what you mean. But in `AvgPoolGrad` input `      ksize: A list of `ints` that has length `>= 4`.
        The size of the sliding window for each dimension of the input.
      strides: A list of `ints` that has length `>= 4`.
        The stride of the sliding window for each dimension of the input.

`

src/ngraph_builder.cc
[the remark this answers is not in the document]
shresthamalik: oh ok. Is it specified somewhere, I am looking for the specification?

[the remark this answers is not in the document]
shresthamalik: Got this from @jianyinglang . Thanks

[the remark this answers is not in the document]
jianyinglang: changed the name from `ng_kernel_shape` to `ng_window_shape`

PR #110, opened 2018-08-02
Shrestha/implement op conv2 d backprop filter

shresthamalik:
Implemented Op Conv2dBackpropFilter. Corrected Typo in set-up instructions README.md


test/tf_exec.cpp
```diff
@@ -360,15 +379,110 @@ TEST(tf_exec, Tile) {
...
+    stride_NCHW[3] = stride[2];
+
+    auto r_ngraph = tf::ops::Conv2DBackpropFilter(
+        root_ngraph.WithOpName("r_NGRAPH"), input_data_NCHW, filter_sizes,
+        output_delta_NCHW, stride_NCHW, padding_type, op_attr_nchw);
+
+    // CPU supports only NHWC
+    auto r_cpu = tf::ops::Conv2DBackpropFilter(
+        root.WithOpName("r_CPU"), input_data, filter_sizes, output_delta,
+        stride, padding_type, op_attr_nhwc);
```
jianyinglang: Is the output shape for `NHWC` and `NCHW` the same?

test/tf_exec.cpp
```diff
@@ -360,15 +379,110 @@ TEST(tf_exec, Tile) {
...
+    ASSERT_EQ(outputs_ngraph[0].shape(), outputs_cpu[0].shape());
+    AssertTensorEquals(outputs_ngraph[0], outputs_cpu[0]);
+  }
+
+  // TEST NCHW
+  // Dialtion rates > 1 not supported on CPU
+  // Current testing only with dialtion rate 1
+  tf::ops::Conv2DBackpropFilter::Attrs op_attr_nchw;
+  op_attr_nchw = op_attr_nchw.DataFormat("NCHW");
+  op_attr_nchw = op_attr_nchw.Dilations({1, 1, 1, 1});
```
jianyinglang: Why is `Dilations` attrs only set for `NCHW` case not in `NHWC` case?

src/ngraph_builder.cc
```diff
@@ -674,6 +675,136 @@ static tf::Status TranslateConv2DOp(const tf::Node* op,
   return tf::Status::OK();
 }
 
+static tf::Status TranslateConv2DBackpropFilterOp(const tf::Node* op,
+                                                  Builder::OpMap& ng_op_map) {
+  TF_RETURN_IF_ERROR(ValidateInputCount(op, 3));
+
+  shared_ptr<ng::Node> ng_data_batch, ng_output_delta;
+  TF_RETURN_IF_ERROR(GetInputNode(ng_op_map, op, 0, &ng_data_batch));
+  TF_RETURN_IF_ERROR(GetInputNode(ng_op_map, op, 2, &ng_output_delta));
```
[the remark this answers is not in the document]
shresthamalik: I will make this change.

test/tf_exec.cpp
shresthamalik: Yes, I believe so. 
The output shape is same as filter size/shape which remains the same irrespective of NHWC or NCHW data format

https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/conv2-d-backprop-filter
Returns:
Output: 4-D with shape [filter_height, filter_width, in_channels, out_channels]. Gradient w.r.t. the filter input of the convolution.

test/tf_exec.cpp
shresthamalik: the tf::ops::Conv2DBackpropFilter::Attrs() is an optional argument. If not passed, the default attrs sets data_format NHWC and dilations [1,1,1,1]. So the NHWC test actually runs with the dilation[1,1,1,1] but it is not explicitly passed.

jianyinglang:
Also, could you also remove the unnecessary code changes by rebasing your branch to master?

PR #111, opened 2018-08-03
Implement op ReluGrad

shresthamalik:


src/ngraph_builder.cc
```diff
@@ -1619,11 +1630,12 @@ static tf::Status TranslateSquareOp(const tf::Node* op,
 
 static tf::Status TranslateSquaredDifferenceOp(const tf::Node* op,
                                                Builder::OpMap& ng_op_map) {
-  return TranslateBinaryOp(op, ng_op_map, [](std::shared_ptr<ng::Node> input1,
```
jianyinglang: Is this reformatting intentional or caused by some different branch checkout?

test/python/test_relugrad.py
```diff
@@ -0,0 +1,49 @@
...
+import tensorflow as tf
+from tensorflow.python.framework import constant_op
+from tensorflow.python.ops.gen_nn_ops import relu_grad
+
+from common import NgraphTest
+
+
+class TestCastOperations(NgraphTest):
+  def test_cast_2d(self):
+    gradients = constant_op.constant([1.5, 2.5, 3.5, 4.5, 5.5, 6.5], shape = [2,3])
```
jianyinglang: This test is a good one. Just think of some improvement, maybe?  Could we try some random number generation instead of constant input? (this is totally optional)

src/ngraph_builder.cc
shresthamalik: Reformatting due to clang.

test/python/test_relugrad.py
shresthamalik: Yes, working on this one.

PR #113, opened 2018-08-03
Adding floordiv, floormod and negate ops

sayantan-nervana:
Adding ops, along with some clang-format changes


src/ngraph_builder.cc
```diff
@@ -906,11 +907,34 @@ static tf::Status TranslateFillOp(const tf::Node* op,
...
   }
-  SaveNgOp(ng_op_map, op->name(), make_shared<ng::op::Broadcast>(
-                                      ng_value, ng_output_shape, ng_axis_set));
+  SaveNgOp(
+      ng_op_map, op->name(),
+      make_shared<ng::op::Broadcast>(ng_value, ng_output_shape, ng_axis_set));
   return tf::Status::OK();
 }
 
+auto ng_floordiv = [](std::shared_ptr<ng::Node> ng_input1,
```
avijit-nervana: Shouldn't define this lambda outside of the FloorDivOp(...). The reason you defined it here so that you can use them in the two Floor ops below but I would refactor it the following way:

1. Move the lambda `ng_floordiv` inside the TranslateFloorDivOp and use it there. 
2. In  TranslateFloorModOp() build the nGraph op in lines 931 through line 933 instead of calling the `ng_floordiv` lambda. Then return the resulting op.

test/tf_exec.cpp
```diff
@@ -472,13 +483,94 @@ TEST(tf_exec, Op_Rsqrt) {
...
-  EXPECT_FLOAT_EQ(1.f/2.f, mat(1, 0));
-  EXPECT_FLOAT_EQ(1.f/8.f, mat(1, 1));
+  EXPECT_FLOAT_EQ(1.f / 16.f, mat(0, 0));
+  EXPECT_FLOAT_EQ(1.f / 4.f, mat(0, 1));
+  EXPECT_FLOAT_EQ(1.f / 2.f, mat(1, 0));
+  EXPECT_FLOAT_EQ(1.f / 8.f, mat(1, 1));
+}
+
+template <typename T>
+void approx_matrix_compare(typename tf::TTypes<T>::Matrix mat,
```
avijit-nervana: Can you not use AssertTensorEquals(...)? 
Also - we use CamelCase for functions and snake_case for variable names.

test/tf_exec.cpp
```diff
@@ -472,13 +483,94 @@ TEST(tf_exec, Op_Rsqrt) {
...
+  for (int r = 0; r < ref_ans.size(); r++)
+    for (int c = 0; c < ref_ans[0].size(); c++)
+      EXPECT_FLOAT_EQ(ref_ans[r][c], mat(r, c));
+}
+
+TEST(tf_exec, Op_Negate) {
+  tf::Scope root = tf::Scope::NewRootScope();
+  root = root.WithDevice("/device:NGRAPH:0");
+
+  auto A = tf::ops::Const(root, {{-256.f, 16.5f}, {0.f, 64.f}});
```
avijit-nervana: Use a Tensor instead of a matrix.

test/tf_exec.cpp
```diff
@@ -472,13 +483,94 @@ TEST(tf_exec, Op_Rsqrt) {
...
+  vector<vector<float>> golden = {{256.f, -16.5f}, {0.f, -64.f}};
+
+  approx_matrix_compare(outputs[0].matrix<float>(), golden);
+}
+
+TEST(tf_exec, Op_FloorDiv) {
+  tf::Scope root = tf::Scope::NewRootScope();
+  root = root.WithDevice("/device:NGRAPH:0");
+
+  auto A = tf::ops::Const(root, {{5.f, 6.f, 7.5f, -1.f, 2.f, -3.f},
```
avijit-nervana: Use a Tensor

test/tf_exec.cpp
```diff
@@ -472,13 +483,94 @@ TEST(tf_exec, Op_Rsqrt) {
...
+  vector<vector<float>> golden1 = {golden0};
+  approx_matrix_compare(outputs[0].matrix<float>(), golden0);
+  approx_matrix_compare(outputs[1].matrix<float>(), golden1);
+}
+
+TEST(tf_exec, Op_FloorMod) {
+  tf::Scope root = tf::Scope::NewRootScope();
+  root = root.WithDevice("/device:NGRAPH:0");
+
+  auto A = tf::ops::Const(root, {{5.f, 6.f, 7.5f, -1.f, 2.f, -3.f},
```
avijit-nervana: Use a Tensor

test/tf_exec.cpp
sayantan-nervana: did not notice that function(AssertTensorEquals). will use it.

test/tf_exec.cpp
```diff
@@ -472,13 +483,94 @@ TEST(tf_exec, Op_Rsqrt) {
...
+  auto r = tf::ops::Negate(root.WithOpName("r"), A);
+
+  std::vector<tf::Tensor> outputs;
+  tf::ClientSession session(root);
+
+  TF_CHECK_OK(session.Run({r}, &outputs));
+
+  ASSERT_EQ(outputs[0].shape(), tf::TensorShape({2, 2}));
+
+  vector<vector<float>> golden = {{256.f, -16.5f}, {0.f, -64.f}};
```
shresthamalik: Would 'expected' be a better variable name for golden?

src/ngraph_builder.cc
shresthamalik: @avijit-nervana  Agreed. However, our current design does not allow us to reuse an already implemented op-translation. For simpler ops, it is not much of a replication but there could be other ops where this might amount to significant code replication. I had faced a similar situation, while implementing another op, there too it was 1-2 line of replication. We might have to re-design, if we come across an op which uses 2-3 simpler ops.

avijit-nervana:
@shresthamalik Could you please add a JIRA ticket to track this feature enhancement? Non this sprint - but on top of the backlog.

avijit-nervana:
@sayantan-nervana Please merge master onto your branch, resolve conflicts and push an update.

test/tf_exec.cpp
```diff
@@ -110,13 +110,23 @@ TEST(tf_exec, axpy) {
...
-    auto a = T1.flat<float>().data()[k];
-    auto b = T2.flat<float>().data()[k];
+    auto a = T1_data[k];
+    auto b = T2_data[k];
     EXPECT_FLOAT_EQ(a, b);
   }
 }
 
+void AssignInputValues(tf::Tensor& A) {
+  auto A_flat = A.flat<float>();
```
avijit-nervana: Variable names use snake_case. Please review this document about the coding standards: https://docs.google.com/document/d/1Xi0rIkMeAmV9ofuOkkmaRoaWFv0GzYkaJMiNq5LtVHo/edit#heading=h.jvtff2kbxiny
Even though you merely followed style in the existing code (which was wrong to begin with) - please update this function following the coding convention.

PR #122, opened 2018-08-06
Langjian/bias add grad

jianyinglang:


src/ngraph_builder.cc
```diff
@@ -612,6 +612,45 @@ static tf::Status TranslateBiasAddOp(const tf::Node* op,
...
+
+  if (tf_data_format != "NHWC" && tf_data_format != "NCHW") {
+    return tf::errors::InvalidArgument(
+        "BiasAddGrad data format is neither NHWC nor NCHW");
+  }
+
+  auto ng_input_shape = ng_input->get_shape();
+  if (ng_input_shape.size() != 4) {
+    return tf::errors::InvalidArgument(
+        "BiasAddGrad argument does not have 4 dimensions");
```
aprocter: The `BiasAddGrad` docs at https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/bias-add-grad indicate that `out_backprop` can have any number of dimensions. If 4D is all we need for the moment then that's fine, but let's register that restriction in the confirmation pass. If we can support n-dimensional inputs, that's even better.

jianyinglang: We can support >=2D for `NHWC` data format but Tensorflow only supports 4D for `NCHW`.  Should we register this restriction in the confirmation pass? https://github.com/tensorflow/tensorflow/blob/f202958ee2d5177a474e3d107fdbf0c83174d099/tensorflow/core/kernels/bias_op.cc#L249

test/tf_exec.cpp
```diff
@@ -119,11 +119,26 @@ void AssertTensorEquals(tf::Tensor T1, tf::Tensor T2) {
   }
 }
 
-void AssignInputValues(tf::Tensor& A) {
+void AssertTensorValue_close(tf::Tensor T1, tf::Tensor T2, float tol) {
```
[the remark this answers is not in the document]
jianyinglang: Renamed the function to be CamelCase `ValidateTensorData`.

test/tf_exec.cpp
```diff
@@ -295,6 +306,42 @@ TEST(tf_exec, BatchMatMul_2D) {
   AssertTensorEquals(outputs[0], outputs_cpu[0]);
 }
 
+TEST(tf_exec, BiasAddGrad) {
+  tf::Scope root = tf::Scope::NewRootScope();
+  auto dev_scope = root.WithDevice("/device:NGRAPH:0");
+  tf::Tensor X(tf::DT_FLOAT, tf::TensorShape({2, 3, 4, 5}));
+  AssignInputValues(X, 0.86f);
```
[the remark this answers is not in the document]
jianyinglang: I checked the output graph and it seems run the right ops. But could you please point me one example for that case so I have a better understanding on the issue?

src/ngraph_builder.cc
```diff
@@ -628,21 +628,29 @@ static tf::Status TranslateBiasAddGradOp(const tf::Node* op,
...
+    }
+    for (size_t i = 0; i < ng_input_shape.size() - 1; i++) {
+      reduction_axes.insert(i);
+    }
   } else {
-    reduction_axes = {0, 2, 3};
+    // Tensorflow NCHW format supports only 4D input/output tensor
+    if (ng_input_shape.size() != 4) {
+      return tf::errors::InvalidArgument(
+          "BiasAddGrad only support 4d input/output for NCHW data format");
```
aprocter: Interesting... so it's supported at shape inference time, but not at actual runtime!

aprocter: (Just to clarify: I'm saying that this is indeed apparently the behavior on the CPU backend.)

PR #124, opened 2018-08-07
L2Loss Op

kbalka:
* l2loss implementation
* unit tests c++ and python
* some clang formatting noise in other sections


src/ngraph_builder.cc
```diff
@@ -1282,6 +1280,29 @@ static tf::Status TranslateIdentityOp(const tf::Node* op,
...
+
+  auto const_2 = make_shared<ng::op::Constant>(
+      ng_input->get_element_type(), ng::Shape{}, std::vector<std::string>{"2"});
+
+  size_t input_rank = ng_input->get_shape().size();
+  ng::AxisSet axes;
+  for (auto i = 0; i < input_rank; ++i) {
+    axes.insert(i);
+  }
+  std::shared_ptr<ng::Node> ng_pow = make_shared<ng::op::Power>(
```
aprocter: Could we elementwise multiply `ng_input` with itself instead?

PR #134, opened 2018-08-10
Enabled backend support

ransford2011:
* Enabled All backends
* Ran unit tests for CPU and INTERPRETER backends (Got a failure about cannot read protobuf on 2 of them)
* Tested `mnist_frop_only.py` on all except GPU

```
(ngraph-tf-dev) root@sc09lcralgo9:~/rmhyman/ngraph-tf/build/test# NGRAPH_TF_BACKEND=INTERPRETER ./gtest_ngtf 
[==========] Running 31 tests from 4 test cases.
[----------] Global test environment set-up.
[----------] 1 test from graph_exec
[ RUN      ] graph_exec.axpy
Calling nGraph function
Tensor<Result_5: 2, 3>{[[6, 6, 6], [6, 6, 6]]}
Tensor<Result_6: 2, 3>{[[5, 5, 5], [5, 5, 5]]}
[       OK ] graph_exec.axpy (1649 ms)
[----------] 1 test from graph_exec (1649 ms total)

[----------] 20 tests from tf_exec
[ RUN      ] tf_exec.hello_world
2018-08-15 16:33:03.838000: I /root/rmhyman/ngraph-tf/test/tf_exec.cpp:55] 19
-3
[       OK ] tf_exec.hello_world (23 ms)
[ RUN      ] tf_exec.axpy
5 5 5 5 5 5 
6 6 6 6 6 6 
[       OK ] tf_exec.axpy (7 ms)
[ RUN      ] tf_exec.BatchMatMul_0D
[       OK ] tf_exec.BatchMatMul_0D (13 ms)
[ RUN      ] tf_exec.BatchMatMul
[       OK ] tf_exec.BatchMatMul (49 ms)
[ RUN      ] tf_exec.BatchMatMul_3D
[       OK ] tf_exec.BatchMatMul_3D (24 ms)
[ RUN      ] tf_exec.BatchMatMul_2D
[       OK ] tf_exec.BatchMatMul_2D (7 ms)
[ RUN      ] tf_exec.BiasAddGrad
[       OK ] tf_exec.BiasAddGrad (34 ms)
[ RUN      ] tf_exec.FusedBatchNormGrad_NHWC
terminate called after throwing an instance of 'ngraph::ngraph_error'
  what():  unsupported op BatchNormBackprop
```

On CPU:
```
(ngraph-tf-dev) root@sc09lcralgo9:~/rmhyman/ngraph-tf/build/test# ./gtest_ngtf 
[==========] Running 31 tests from 4 test cases.
[----------] Global test environment set-up.
[----------] 1 test from graph_exec
[ RUN      ] graph_exec.axpy
Calling nGraph function
Tensor<Result_5: 2, 3>{[[6, 6, 6], [6, 6, 6]]}
Tensor<Result_6: 2, 3>{[[5, 5, 5], [5, 5, 5]]}
[       OK ] graph_exec.axpy (1690 ms)
[----------] 1 test from graph_exec (1690 ms total)

[----------] 20 tests from tf_exec
[ RUN      ] tf_exec.hello_world
2018-08-15 16:34:55.724822: I /root/rmhyman/ngraph-tf/test/tf_exec.cpp:55] 19
-3
[       OK ] tf_exec.hello_world (258 ms)
[ RUN      ] tf_exec.axpy
5 5 5 5 5 5 
6 6 6 6 6 6 
[       OK ] tf_exec.axpy (252 ms)
[ RUN      ] tf_exec.BatchMatMul_0D
[       OK ] tf_exec.BatchMatMul_0D (67 ms)
[ RUN      ] tf_exec.BatchMatMul
[       OK ] tf_exec.BatchMatMul (1680 ms)
[ RUN      ] tf_exec.BatchMatMul_3D
[       OK ] tf_exec.BatchMatMul_3D (1482 ms)
[ RUN      ] tf_exec.BatchMatMul_2D
[       OK ] tf_exec.BatchMatMul_2D (278 ms)
[ RUN      ] tf_exec.BiasAddGrad
[       OK ] tf_exec.BiasAddGrad (1443 ms)
[ RUN      ] tf_exec.FusedBatchNormGrad_NHWC
[       OK ] tf_exec.FusedBatchNormGrad_NHWC (363 ms)
[ RUN      ] tf_exec.Op_L2Loss
[       OK ] tf_exec.Op_L2Loss (615 ms)
[ RUN      ] tf_exec.Tile
[       OK ] tf_exec.Tile (635 ms)
[ RUN      ] tf_exec.Op_Conv2DBackpropFilter
[       OK ] tf_exec.Op_Conv2DBackpropFilter (1167 ms)
[ RUN      ] tf_exec.Op_RealDiv
[       OK ] tf_exec.Op_RealDiv (219 ms)
[ RUN      ] tf_exec.Op_Reciprocal
[       OK ] tf_exec.Op_Reciprocal (290 ms)
[ RUN      ] tf_exec.Op_Square
[       OK ] tf_exec.Op_Square (290 ms)
[ RUN      ] tf_exec.Op_SquaredDifference
[       OK ] tf_exec.Op_SquaredDifference (257 ms)
[ RUN      ] tf_exec.Op_Rsqrt
[       OK ] tf_exec.Op_Rsqrt (261 ms)
[ RUN      ] tf_exec.Op_Negate
[       OK ] tf_exec.Op_Negate (283 ms)
[ RUN      ] tf_exec.Op_FloorDiv
[       OK ] tf_exec.Op_FloorDiv (277 ms)
[ RUN      ] tf_exec.Op_FloorMod
[       OK ] tf_exec.Op_FloorMod (329 ms)
[ RUN      ] tf_exec.Op_AddN
[       OK ] tf_exec.Op_AddN (294 ms)
[----------] 20 tests from tf_exec (10741 ms total)

[----------] 3 tests from padding
[ RUN      ] padding.valid
[       OK ] padding.valid (0 ms)
[ RUN      ] padding.divisible
[       OK ] padding.divisible (0 ms)
[ RUN      ] padding.indivisible
[       OK ] padding.indivisible (0 ms)
[----------] 3 tests from padding (0 ms total)

[----------] 7 tests from conversions
[ RUN      ] conversions.reshape
[       OK ] conversions.reshape (1 ms)
[ RUN      ] conversions.batch_to_tensorflow_nchw
[       OK ] conversions.batch_to_tensorflow_nchw (0 ms)
[ RUN      ] conversions.batch_to_tensorflow_nhwc
[       OK ] conversions.batch_to_tensorflow_nhwc (0 ms)
[ RUN      ] conversions.batch_to_ngraph_nchw
[       OK ] conversions.batch_to_ngraph_nchw (0 ms)
[ RUN      ] conversions.param_to_ngraph_nchw
[       OK ] conversions.param_to_ngraph_nchw (0 ms)
[ RUN      ] conversions.batch_to_ngraph_nhwc
[       OK ] conversions.batch_to_ngraph_nhwc (0 ms)
[ RUN      ] conversions.param_to_ngraph_nhwc
[       OK ] conversions.param_to_ngraph_nhwc (0 ms)
[----------] 7 tests from conversions (1 ms total)

[----------] Global test environment tear-down
[==========] 31 tests from 4 test cases ran. (12432 ms total)
[  PASSED  ] 31 tests.
```


CMakeLists.txt
```diff
@@ -101,7 +101,7 @@ ExternalProject_Add(
     ext_ngraph
     GIT_REPOSITORY https://github.com/NervanaSystems/ngraph
     GIT_TAG v0.5.0
-    CMAKE_ARGS -DNGRAPH_INSTALL_PREFIX=${CMAKE_CURRENT_BINARY_DIR}/ngraph/ngraph_dist -DNGRAPH_USE_PREBUILT_LLVM=${NGRAPH_USE_PREBUILT_LLVM}
+    CMAKE_ARGS -DNGRAPH_INSTALL_PREFIX=${CMAKE_CURRENT_BINARY_DIR}/ngraph/ngraph_dist -DNGRAPH_USE_PREBUILT_LLVM=${NGRAPH_USE_PREBUILT_LLVM} -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
```
ransford2011: As a Fan of `gdb` I need to have this option available 😉

src/ngraph_encapsulate_op.cc
```diff
@@ -195,6 +207,16 @@ class NGraphEncapsulateOp : public tf::OpKernel {
       }
       last_used_src_ptrs[i] = src_ptr;
       ng_inputs.push_back(t);
+      };
+
+      if(env_value == "NNP") {
+      auto t = m_ng_backend->create_tensor(ng_element_type, ng_shape);
```
ransford2011: I'm not sure if GPU needs something similar to this.  @fengleitian can confirm.

src/ngraph_encapsulate_op.cc
```diff
@@ -229,9 +251,15 @@ class NGraphEncapsulateOp : public tf::OpKernel {
 
       // Create the nGraph output tensor
       void* dst_ptr = tf::DMAHelper::base(output_tensor);
-      auto t_result = m_ng_backend->create_tensor(elem_type, shape, dst_ptr);
+      if(env_value == "NNP") {
```
ransford2011: Same here. Not sure if GPU should guarded by this mechanism.

src/ngraph_encapsulate_op.cc
```diff
@@ -66,13 +67,22 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // TODO(amprocte): need to check status result here.
     OP_REQUIRES_OK(ctx, tf::ConvertGraphDefToGraph(opts, *graph_def, &m_graph));
 
+     auto env_value = std::string(std::getenv("NGRAPH_TF_BACKEND")); 
     // Create the backend
     if (m_ng_backend == nullptr) {
-#if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
```
avijit-nervana: Need to keep this - else you will be breaking the TF upstreamming version.

src/ngraph_encapsulate_op.cc
```diff
@@ -66,13 +67,22 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+	if( env_value  == "INTERPRETER") {
+      		m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
+	}
+	else if( env_value == "CPU") {
+      		m_ng_backend = ng::runtime::Backend::create("CPU");
+	}
+	else if( env_value == "NNP") {
+      		m_ng_backend = ng::runtime::Backend::create("NNP");
+	}
+	else if( env_value == "GPU") {
```
avijit-nervana: Right now the GPU won't work unless some changes are made in the data handling part. So please remove this option until that time.

src/ngraph_encapsulate_op.cc
```diff
@@ -66,13 +67,22 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
     if (m_ng_backend == nullptr) {
-#if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
-#else
-      m_ng_backend = ng::runtime::Backend::create("CPU");
-#endif
+	if( env_value  == "INTERPRETER") {
+      		m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
+	}
+	else if( env_value == "CPU") {
```
avijit-nervana: CPU option should be default i.e., put this under the `else` section.

src/ngraph_encapsulate_op.cc
avijit-nervana: This needs to change. We shouldn't rely on env variable to determine the backend everywhere - only once as it's done during the initialization time. In this class we should get this from the context.

CMakeLists.txt
avijit-nervana: Me too 👍 Can we split this long line to multiple lines?

src/ngraph_encapsulate_op.cc
fengleitian: Should be same. But let me double check.

fengleitian: Additional step needed for GPU. check this:
https://github.com/NervanaSystems/ngraph-tf/pull/74/files

ransford2011: @fengleitian Looks like I'm going to remove support for GPU as requested by @avijit-nervana  in this PR so I will let you take care of it.

src/ngraph_encapsulate_op.cc
```diff
@@ -229,9 +251,15 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
       void* dst_ptr = tf::DMAHelper::base(output_tensor);
-      auto t_result = m_ng_backend->create_tensor(elem_type, shape, dst_ptr);
+      if(env_value == "NNP") {
+      	auto t_result = m_ng_backend->create_tensor(elem_type, shape);
+        outputs.push_back(t_result);
+      }
+      else {
+      	auto t_result = m_ng_backend->create_tensor(elem_type, shape, dst_ptr);
+        outputs.push_back(t_result);
+      }
```
yxlao: There are missing parts:
1) before `m_ng_backend->call()`, do a tensor copy from host->device
2) after `m_ng_backend->call()`, do a tensor copy from device->host
Otherwise, it doesn't seem that there is a place where NNP is receiving the data.

src/ngraph_encapsulate_op.cc
```diff
@@ -195,6 +215,16 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+      };
+
+      if(m_ng_backend_name == "NNP") {
+      auto t = m_ng_backend->create_tensor(ng_element_type, ng_shape);
+	add_input_tensor(t);
+          }
+    else {
+      auto t = m_ng_backend->create_tensor(ng_element_type, ng_shape, src_ptr);
+	add_input_tensor(t);
+    }
```
yxlao: Run clang-format, use 2-space indent

src/ngraph_encapsulate_op.cc
```diff
@@ -69,9 +70,27 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // Create the backend
     if (m_ng_backend == nullptr) {
 #if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
+    	m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
+#elif defined(NGRAPH_TF_BACKEND)
+        auto env_value = std::string(std::getenv("NGRAPH_TF_BACKEND"));
```
yxlao: Check `nullptr` first before you can cast it to string
https://stackoverflow.com/questions/5867242/how-to-put-the-contents-of-getenv-into-a-string

src/ngraph_encapsulate_op.cc
```diff
@@ -69,9 +70,27 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // Create the backend
     if (m_ng_backend == nullptr) {
 #if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
+    	m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
+#elif defined(NGRAPH_TF_BACKEND)
+        auto env_value = std::string(std::getenv("NGRAPH_TF_BACKEND")); 
+	if( env_value.find("INTERPRETER") != string::npos) {
```
yxlao: If the `env_value` is empty, use CPU. Otherwise, directly pass-through the `env_value` to the `ng::runtime::Backend::create`. Since there will be newer backends not listed here.

src/ngraph_encapsulate_op.cc
```diff
@@ -195,6 +215,16 @@ class NGraphEncapsulateOp : public tf::OpKernel {
       }
       last_used_src_ptrs[i] = src_ptr;
       ng_inputs.push_back(t);
+      };
+
+      if(m_ng_backend_name == "NNP") {
```
yxlao: same as above, i would use CPU vs non-CPU case

src/ngraph_encapsulate_op.cc
```diff
@@ -229,9 +259,15 @@ class NGraphEncapsulateOp : public tf::OpKernel {
 
       // Create the nGraph output tensor
       void* dst_ptr = tf::DMAHelper::base(output_tensor);
-      auto t_result = m_ng_backend->create_tensor(elem_type, shape, dst_ptr);
+      if(m_ng_backend_name == "NNP") {
+      	auto t_result = m_ng_backend->create_tensor(elem_type, shape);
```
yxlao: For non-host backends, we shall only be calling `m_ng_backend->create_tensor(ng_element_type, ng_shape);` for the **first** time. Then we shall use the cached version of the created tensor. `m_ng_backend->create_tensor` is not free for non-CPU backends. We can afford to create tensors multiple times in CPU because the `src_ptr` is reused when we call `m_ng_backend->create_tensor(ng_element_type, ng_shape, src_ptr);`

We shall have a map of `<ng::Function, vector<ng::runtime::TensorView>> map_ng_function_to_inputs` to cache the input tensors.

src/ngraph_encapsulate_op.cc
```diff
@@ -195,6 +215,16 @@ class NGraphEncapsulateOp : public tf::OpKernel {
       }
       last_used_src_ptrs[i] = src_ptr;
       ng_inputs.push_back(t);
+      };
+
+      if(m_ng_backend_name == "NNP") {
+      auto t = m_ng_backend->create_tensor(ng_element_type, ng_shape);
```
yxlao: For non-host backends, we shall only be calling `m_ng_backend->create_tensor(ng_element_type, ng_shape);` for the first time. Then we shall use the cached version of the created tensor. `m_ng_backend->create_tensor` is not free for non-CPU backends. We can afford to create tensors multiple times in CPU because the `src_ptr` is reused when we call `m_ng_backend->create_tensor(ng_element_type, ng_shape, src_ptr);`

We shall have a map of `<ng::Function, vector<ng::runtime::TensorView>> map_ng_function_to_inputs` to cache the input tensors.

src/ngraph_encapsulate_op.cc
yxlao: Avoid `getenv` multiple times, save the "backend_string" value as a member of the class during class construction.

src/ngraph_encapsulate_op.cc
```diff
@@ -66,13 +67,22 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
-      m_ng_backend = ng::runtime::Backend::create("CPU");
-#endif
+	if( env_value  == "INTERPRETER") {
+      		m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
+	}
+	else if( env_value == "CPU") {
+      		m_ng_backend = ng::runtime::Backend::create("CPU");
+	}
+	else if( env_value == "NNP") {
+      		m_ng_backend = ng::runtime::Backend::create("NNP");
```
wenzhe-nrv: just wonder if we could use "NNP:0" as default? same may apply to GPU if we keep it there.

src/ngraph_encapsulate_op.cc
yxlao: I would also rename `env_value` to something more meaningful like `ng_backend_str`

src/ngraph_encapsulate_op.cc
fengleitian: GPU works same as NNP. 
1. Add data copy from CPU side to Device side when create input tensor.
2. Add data copy from Device to CPU side after get result.

src/ngraph_encapsulate_op.cc
fengleitian: Same on GPU.

src/ngraph_encapsulate_op.cc
ransford2011: I'm not sure if this is still relevant given https://github.com/NervanaSystems/ngraph-tf/blob/8c7c303852fba4382164842083235ded7bbbb17d/src/ngraph_encapsulate_op.cc#L304

yxlao: the fix you had exactly addressed this problem, nice fix

src/ngraph_encapsulate_op.cc
ransford2011: @yxlao .  I'm not so sure about blindly passing in a value that the user prevents.  This will generate an error during backend creation.  If the bridge developers are fine with that, then I'm ok.  I prefer to be more defensive and only allow what we can support.

yxlao: Say the user downloaded a `libnew_backend.so` and put it in the proper location. We want this to work out-of-the-box without compiling the bridge again. Ultimately nGraph backend creator will check whether the backend string is valid or not, here in the bridge, we don't have the information whether a particular backend is present or not. 

So it seems to me that passing the value to `backend::create` and letting the backend creator handle the value would be a better choice. E.g. we could catch an exception here if the `backend::create` throws an exception if there is an error code returned.

src/ngraph_encapsulate_op.cc
ransford2011: This seems like something that should be a part of Core since we are doing the same at both ends 🤔

src/ngraph_encapsulate_op.cc
yxlao: I would leave this decision to `ngraph-tf` devs

src/ngraph_encapsulate_op.cc
```diff
@@ -179,30 +185,47 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+        auto t =
+            m_ng_backend->create_tensor(ng_element_type, ng_shape, src_ptr);
+        add_input_tensor(t);
       } else {
-        t->set_stale(true);
+        auto t = m_ng_backend->create_tensor(ng_element_type, ng_shape);
+        auto element_type =
+            t->get_descriptor()->get_tensor_view_layout()->get_element_type();
+        t->write(src_ptr, 0,
+                 (t->get_element_count() * element_type.bitwidth() / 8.0));
```
ransford2011: @avijit-nervana @aprocter .  I will need your help in letting me know if you have a place that has this logic stored.  I'm not sure if this is the cleanest way to implement this at the Framework level.

src/ngraph_encapsulate_op.cc
```diff
@@ -228,12 +251,21 @@ class NGraphEncapsulateOp : public tf::OpKernel {
                                "the element type expected by TensorFlow"));
 
       // Create the nGraph output tensor
-      void* dst_ptr = tf::DMAHelper::base(output_tensor);
-      auto t_result = m_ng_backend->create_tensor(elem_type, shape, dst_ptr);
-
-      outputs.push_back(t_result);
+      dst_ptr = tf::DMAHelper::base(output_tensor);
```
fengleitian: could have multiple outputs.

src/ngraph_encapsulate_op.cc
```diff
@@ -69,10 +70,15 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
     if (m_ng_backend == nullptr) {
 #if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
-#else
-      m_ng_backend = ng::runtime::Backend::create("CPU");
+      m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
 #endif
+      if (std::getenv("NGRAPH_TF_BACKEND") != nullptr) {
+        m_ng_backend_name = std::string(std::getenv("NGRAPH_TF_BACKEND"));
+        cout << "Backend name : " << m_ng_backend_name << endl;
```
aprocter: Let's change to `NGRAPH_VLOG(2)`

src/ngraph_encapsulate_op.cc
```diff
@@ -179,30 +185,47 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+            t->get_descriptor()->get_tensor_view_layout()->get_element_type();
+        t->write(src_ptr, 0,
+                 (t->get_element_count() * element_type.bitwidth() / 8.0));
+        add_input_tensor(t);
       }
-      last_used_src_ptrs[i] = src_ptr;
-      ng_inputs.push_back(t);
     }
 
+    // m_ng_function_to_inputs.insert(ng_function , ng_inputs);
```
aprocter: Delete

src/ngraph_encapsulate_op.cc
```diff
@@ -228,12 +251,21 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+      } else if (m_ng_function_to_outputs.find(ng_function) ==
+                 end(m_ng_function_to_outputs)) {
+        auto t_result = m_ng_backend->create_tensor(elem_type, shape);
+        outputs.push_back(t_result);
+      }
     }
 
+    if (m_ng_function_to_outputs.find(ng_function) ==
+        end(m_ng_function_to_outputs)) {
+      // m_ng_function_to_inputs.insert(ng_function , outputs);
```
aprocter: Not sure what the conditional is doing since this is commented out

src/ngraph_encapsulate_op.cc
```diff
@@ -69,10 +70,18 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // Create the backend
     if (m_ng_backend == nullptr) {
 #if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
-#else
-      m_ng_backend = ng::runtime::Backend::create("CPU");
+      m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
 #endif
```
avijit-nervana: #else

src/ngraph_encapsulate_op.cc
```diff
@@ -262,9 +306,16 @@ class NGraphEncapsulateOp : public tf::OpKernel {
       m_ng_functions;
   std::map<std::shared_ptr<ngraph::Function>, std::vector<const void*>>
       m_last_used_src_ptrs_map;
+  std::unordered_map<std::shared_ptr<ngraph::Function>,
+                     std::vector<ng::runtime::TensorView>>
+      m_ng_function_to_inputs;
```
aprocter: (Didn't catch this til I got down here.) What are `m_ng_function_to_inputs` and `_outputs` for?

src/ngraph_encapsulate_op.cc
```diff
@@ -69,10 +70,18 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // Create the backend
     if (m_ng_backend == nullptr) {
 #if defined(NGRAPH_EMBEDDED_IN_TENSORFLOW)
-      m_ng_backend = std::make_shared<ng::runtime::interpreter::INTBackend>();
-#else
-      m_ng_backend = ng::runtime::Backend::create("CPU");
+      m_ng_backend = ng::runtime::Backend::create("INTERPRETER");
 #endif
+      if (std::getenv("NGRAPH_TF_BACKEND") != nullptr) {
```
avijit-nervana: Get it once

src/ngraph_encapsulate_op.cc
```diff
@@ -69,10 +70,18 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+        m_ng_backend_name = std::string(std::getenv("NGRAPH_TF_BACKEND"));
+        if (!m_ng_backend_name.empty()) {
+          m_ng_backend = ng::runtime::Backend::create(m_ng_backend_name);
+        }
+      }
+      else{ // env not defined.  Default to CPU backend
+          m_ng_backend = ng::runtime::Backend::create("CPU");
```
avijit-nervana: call it once

src/ngraph_encapsulate_op.cc
aprocter: yup

src/ngraph_encapsulate_op.cc
aprocter: :thisisfine:

ransford2011:
This should address majority of the comments.  I would like to refactor the caching logic into a single for both the inputs and outputs maps that I created, but I have to go on 🌴 sometime in my life 😉 .  I will let @yxlao address any final comments.  I think besides the duplicate code for the inputs/outputs, the logic should be ok.  Thanks for all the feedback! 💯

ransford2011:
Also note that I INTERPRETER does work, but since you don't have the nGraph version with the BatchNormBackProp implemented, it halts on `tf_exec.FusedBatchNormGrad_NHWC`

src/ngraph_encapsulate_op.cc
```diff
@@ -165,51 +185,82 @@ class NGraphEncapsulateOp : public tf::OpKernel {
     // Allocate tensors for arguments.
     vector<shared_ptr<ng::runtime::TensorView>> ng_inputs;
 
-    auto& last_used_src_ptrs = m_last_used_src_ptrs_map[ng_function];
-    last_used_src_ptrs.resize(input_shapes.size());
+    std::vector<std::pair<void*, std::shared_ptr<ng::runtime::TensorView>>>&
+        input_caches = m_ng_function_input_cache_map[ng_function];
+    input_caches.resize(input_shapes.size());
```
fengleitian: We should be able to check if it's cached and goes to copy directly for GPU and NNP if it is.

src/ngraph_encapsulate_op.cc
```diff
@@ -165,51 +185,82 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
                    << m_ngraph_cluster;
 
     // Allocate tensors for the results.
-    vector<shared_ptr<ng::runtime::TensorView>> outputs;
+    vector<shared_ptr<ng::runtime::TensorView>> ng_outputs;
+
+    std::vector<std::pair<void*, std::shared_ptr<ng::runtime::TensorView>>>&
+        output_caches = m_ng_function_output_cache_map[ng_function];
+    output_caches.resize(ng_function->get_output_size());
+
```
fengleitian: We should be able to check if it's cached and goes to copy directly for GPU and NNP if it is.

fengleitian:
I'm trying to test on GPU side.

fengleitian:
Looks it works on GPU side.  32 tests passed out of 34.
```
[==========] 34 tests from 4 test cases ran. (41927 ms total)
[  PASSED  ] 32 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] tf_exec.FusedBatchNormGrad_NHWC
[  FAILED  ] tf_exec.Op_SparseSoftmaxCrossEntropyWithLogits
```

src/ngraph_encapsulate_op.cc
aprocter: I don't understand what you mean here...

src/ngraph_encapsulate_op.cc
```diff
@@ -165,51 +185,82 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
-      //      the last time ng_function was called, and
-      //   2. we are using the same tensor in this argument position as
-      //      the one we used last time ng_function was called.
-      if (m_freshness_tracker->IsFresh(src_ptr, ng_function) &&
-          src_ptr == last_used_src_ptrs[i]) {
-        t->set_stale(false);
+      // At the first call of the ng_function, both last_src_ptr and
+      // last_tv shall point to null. Otherwise, they are retrived
+      // from cache.
+      void* last_src_ptr = input_caches[i].first;
```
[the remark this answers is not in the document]
yxlao: You're right, we only need `last_src_ptr` for `CPU`. However, we'll need to cache`ng::runtime::TensorView` for all backends, and currently they are stored in a pair. So it might be shorter to keep the current way. Otherwise, adding
```cpp
if (m_ng_backend_name == "CPU") {
  void* last_src_ptr = input_caches[i].first;
}
```
here doesn't seem to save us much.

src/ngraph_encapsulate_op.cc
yxlao: I think what @fengleitian means is that we can check that if all cached TensorViews are not `nullptr`, we can skip the shape and element type computation logic below, and go directly to CPU tensor freshness check or Device memory copy. The saving would be
```cpp
ng::Shape ng_shape(input_shapes[i].dims());
for (int j = 0; j < input_shapes[i].dims(); ++j) {
  ng_shape[j] = input_shapes[i].dim_size(j);
}
ng::element::Type ng_element_type;
OP_REQUIRES_OK(ctx, TFDataTypeToNGraphElementType(ctx->input(i).dtype(), &ng_element_type));
```
and the cost would be checking that all cached TensorViews are not `nullptr`

It seems like that would add additional complexity to the logic, while the saving is minimal.

src/ngraph_encapsulate_op.cc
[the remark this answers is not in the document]
fengleitian: Make sense.

src/ngraph_encapsulate_op.cc
```diff
@@ -165,51 +185,82 @@ class NGraphEncapsulateOp : public tf::OpKernel {
...
+
+      void* current_src_ptr = (void*)tf::DMAHelper::base(&ctx->input(i));
+      std::shared_ptr<ng::runtime::TensorView> current_tv;
+
+      if (m_ng_backend_name == "CPU") {
+        // We need to check last_tv != nullptr, since there are cases where at
+        // the first call to the ng_function, both the current_src_ptr (when the
+        // input is a 0-sized tensor) and last_src_ptr (uninitialized at the
+        // first call) are nullptr
+        if (current_src_ptr == last_src_ptr && last_tv != nullptr) {
```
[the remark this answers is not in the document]
fengleitian: OK. Got it, looks like that's means it's more than just set a pointer. In this case. the current way is better.

src/ngraph_encapsulate_op.cc
fengleitian: it will be like.
```
if(!cached)
{
  if(CPU)
  {
      create_tensorA();
  }
  else
  {
     create_tensorB();
  }
}
else
{
  if(CPU)
  {
      if(currnt_ptr != last_ptr)  create_tensorA();
  }
}

if(!CPU)  
{
   copy;
}
```
I'm not going to hold this PR for this.

PR #138, opened 2018-08-11
Make clang-format check required for CI

aprocter:
Porting this stuff over from the main `ngraph` repo. Had to make a few minor changes to the scripts there as follows:

* Different source dir names
* Have to take in `.cc` and `.h` as well
* Exclude `cpu_codegen` from search paths

I think we should merge to `master` rather than `r0.4` to minimize disruption for now. Note that the format check will actually _fail_ on CI for now. I don't want to actually run the format script until `r0.4` and `master` are back together.

### DIFF
--- a/maint/apply-code-format.sh
+++ b/maint/apply-code-format.sh
@@ -59,11 +59,11 @@ for ROOT_SUBDIR in ${SRC_DIRS}; do
         #
         # We also skip any dir named "cpu_codegen" in case there are
         # nGraph-generated files lying around from a test run.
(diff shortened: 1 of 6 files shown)


avijit-nervana:
@crlishka Mac does have `clang`. For Windows and other OSs - we should follow what TensorFlow does i.e., they may have some guidances for those OSs.

avijit-nervana:
@crlishka Worst case - we can disable the formatting check for those OSs - which should be fine.

PR #141, opened 2018-08-14
Dynamic shape support

aprocter:
Still needs a bit of doc/cleanup/internal refactoring (not happy with how the static_input_map is passed around) but going to go ahead and open


aprocter:
TESTNOW

src/ngraph_mark_for_clustering.cc
```diff
@@ -706,6 +471,25 @@ bool NodeIsMarkedForClustering(const Node* node) {
...
+  if (GetNodeAttr(node->attrs(), "_ngraph_static_inputs", inputs) != Status::OK()) {
+    *inputs = std::vector<int32>{};
+  }
+}
+
+bool InputIsStatic(const Node* node, int index) {
+  std::vector<int32> inputs;
+  GetStaticInputs(node,&inputs);
+
+  for (auto i : inputs) {
```
sayantan-nervana: The for loop can be replaced by a 'find' line"
return std::find(inputs.begin(), inputs.end(), index)!=inputs.end();

src/ngraph_mark_for_clustering.cc
```diff
@@ -266,20 +230,7 @@ Status MarkForClustering(Graph* graph) {
...
-                                &tf_orig_input_shape_vec) !=
-                Status::OK() ||
-            tf_orig_input_shape_vec.size() != 4) {
-          *result = false;
-          return Status::OK();
-        }
-
-        n->AddAttr("_ngraph_avgpoolgrad_static_input_shape",
-                   tf_orig_input_shape_vec);
+        SetStaticInputs(n, {0});
```
sayantan-nervana: It seems that the body of the static_input nodes are almost same as the always function:
How about this:

ConfirmationFunction CreateConfirmationFunction(vector<int> staticinps = vector<int>()){
    auto cf = [staticinps](Node* n, bool* result) {
      auto make_positive = [n](int x){return  (x>0? x : n->num_inputs() - x);};
      *result = true;
      if (staticinps.size()>0){
        vector<int> pos_idx(staticinps.size());
        for(auto idx : staticinps)
          pos_idx.push_back(make_positive(idx));
        SetStaticInputs(n, staticinps);
      }
      return Status::OK();
    };
    return cf;
}

auto always = CreateConfirmationFunction();

and then we can have:
confirmation_functions["AvgPool"] = CreateConfirmationFunction({0});
or even negative numbers:
confirmation_functions["ConcatV2"] = CreateConfirmationFunction({-1});
and for normal cases:
confirmation_functions["BiasAdd"] = always; or confirmation_functions["BiasAdd"] = CreateConfirmationFunction();

aprocter:
Moving this back to [WIP] state because of a confirmed bug with the contraction logic (failing test case to be checked in shortly). I think I know how to fix it.

aprocter:
During standup we decided that @sayantan-nervana can implement a fix for the bug, so I'm going to focus on review comments, take this out of WIP, and let him take it from here  :)

src/ngraph_mark_for_clustering.cc
aprocter: i likes it

test/graph_rewrites/assign_clusters.cc
```diff
@@ -0,0 +1,92 @@
...
+
+namespace tensorflow {
+
+namespace ngraph_bridge {
+
+#define ASSERT_OK(x) ASSERT_EQ((x), ::tensorflow::Status::OK());
+
+// Given a graph of this form:
+//
+//  Node1--->Node2
```
sayantan-nervana: The cluster assignment happens after POST_REWRITE_FOR_EXEC. So tensorflow has already run its optimizations at this point.

Suppose we have a graph like this. N1 is a node with a static input marked by *
C1          C2
  \            /
    \        /
        +
        |
        |
        v*
        N1 <---------N2

Then after TF's optimizations, the graph should look like this after constant folding:
C(=C1+C2)
   |
   |
   v*
   N1<------N2

At which point we can handle this a static input


Here is a suggested test:
Have a constant-folding scenario and check that with L1 optimization enabled tensorflow, we get C and N1 in the same cluster. With L0 optimization on tensorflow, they should be on different clusters.

aprocter: > (Trying out ascii art) :-D

Nice, I like it  ;)

We could test this, but I think what you're describing here has more to do with properties of the constant-folding pass than with properties of our cluster-assignment pass. So unless we have reason to believe that there might be some failure emerging from a weird interaction between CF and cluster assignment, a simple test like:

```
C
|
|
v*
N1<------ N2
```

would be sufficient in and of itself, I think.

PR #143, opened 2018-08-16
Igor/fusedbatchnorm fix

megawidget:
Fixed FusedBatchNorm to actually work.  It also now has a test.  Furthermore, tweaked `GetInputNode` to provide more detail on errors.

The test fails sometimes when rtol < 0.01, which doesn't seem right to me.  Also, I updated the epsilon to TensorFlow's default, let me know if that's the wrong thing to do.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -131,10 +131,19 @@ static tf::Status GetInputNode(const Builder::OpMap& ng_op_map,
 
   tf::Node* tf_input;
   TF_RETURN_IF_ERROR(op->input_node(input_idx, &tf_input));
+  const std::vector<shared_ptr<ng::Node>>* ng_op = nullptr;
   try {
-    *result = ng_op_map.at(tf_input->name()).at(src_output_idx);
+    ng_op = &ng_op_map.at(tf_input->name());
   } catch (const out_of_range&) {
-    return tf::Status(tensorflow::error::NOT_FOUND, "Input node not found");
+    return tf::Status(tensorflow::error::NOT_FOUND,
+                      string("Ngraph op not found for ") + tf_input->name());
+  }
+  try {
+    *result = ng_op->at(src_output_idx);
+  } catch (const out_of_range&) {
+    return tf::Status(
+        tensorflow::error::NOT_FOUND,
+        string("Input node not found at index ") + to_string(src_output_idx));
   }
   return tf::Status::OK();
 }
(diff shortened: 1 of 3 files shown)


test/python/test_fusedbatchnorm.py
```diff
@@ -0,0 +1,76 @@
...
+class TestFusedBatchNorm(NgraphTest):
+  x = np.random.rand(64, 3, 10, 8).astype('f')
+  scale = [1.0, 0.9, 1.1]
+  offset = [0.1, 0.2, -.3]
+
+  def test_fusedbatchnorm_nchw(self):
+    with self.device:
+      norm = tf.nn.fused_batch_norm(self.x, self.scale, self.offset,
+                                    data_format='NCHW')
+
```
sayantan-nervana: The TranslateOp function of batchnorm has 2 paths one for training one for testing.
By default the 'is_training' parameter of tf.nn.fused_batch_norm is true. We should have a testcase that activates the test path of the TranslateOp with is_training=false

sayantan-nervana: Not related to this change directly: tf_is_training in TranslateFusedBatchNormGradOp is being set, but never used. Maybe we can get rid of that?

sayantan-nervana: "is_training attribute not present, setting to true" ... Not sure if we prefer true or false. For grad ops we can clearly chose is_trainable to be true, not sure for non-grad ops.

aprocter: > "is_training attribute not present, setting to true" ... Not sure if we prefer true or false. For grad ops we can clearly chose is_trainable to be true, not sure for non-grad ops.

I think I wrote that, back when I wasn't sure whether we would need to set the default attribute values ourselves in the builder, or could rely on them to be inserted by TF itself. It appears to be the latter.

Same thing with the `epsilon` attribute.

src/ngraph_builder.cc
```diff
@@ -131,10 +131,19 @@ static tf::Status GetInputNode(const Builder::OpMap& ng_op_map,
 
   tf::Node* tf_input;
   TF_RETURN_IF_ERROR(op->input_node(input_idx, &tf_input));
+  const std::vector<shared_ptr<ng::Node>>* ng_op = nullptr;
```
aprocter: Why does `ng_op` need to be a pointer?

megawidget: because it cannot be a reference if it gets set later.

aprocter: Oooh okay, and that assignment has to be scoped inside the `try` so we can't set it at initialization time without makin' a mess. Works for me.

PR #149, opened 2018-08-20
Fix for static input bug

sayantan-nervana:
Create a shadow path in gc between 2 nodes
if we do NOT want them to be colocated

### DIFF
--- a/src/ngraph_assign_clusters.cc
+++ b/src/ngraph_assign_clusters.cc
@@ -123,6 +123,39 @@ Status AssignClusters(Graph* graph) {
     }
   }
 
+  // If we wish to add a constraint that 2 particular nodes not lie in the same
+  // cluster, then all we have to do is add a 2 'shadow' edges and 1 'shadow' node
+  // in the gc data structure between the 2 nodes. The shadow edges go from the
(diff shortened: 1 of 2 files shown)


aprocter:
Looks good overall. Thanks for this!

1. Will need to run `maint/apply-code-format.sh` and commit, to pass the formatting check in CI
2. Please re-enable the `assign_clusters.cone` test, which we had previously disabled (see https://github.com/NervanaSystems/ngraph-tf/blob/master/test/ci/run-premerge-ci-checks.sh#L40)

sayantan-nervana:
Made changes suggested by Adam.
maint/apply-code-format.sh is super useful!

aprocter:
> maint/apply-code-format.sh is super useful!

Yeah, in nGraph Core we've been doing it all along and it saves a lot of grief and diff noise :)

One thing you might want to do if you get a chance: it looks like `apply-code-format` did make a bit of a mess of your nicely formatted comment (it wrapped some long lines). If you feel like it, you could reformat it by hand (keeping within 78 columns, or 80, or whatever the limit is) and run `maint/check-code-format.sh` to see if it's still happy. Up to you.

PR #150, opened 2018-08-20
add resnet unit test

sharathns93:
Test that runs the same resnet graph on nGraph and CPU and compares the logits. 
Supports NHWC when using normal CPU and NCHW when built with MKL. 
Also supports different resnet sizes (--resnet_size) and training and inference phases (--phase_training=True)


aprocter:
TESTNOW

sharathns93:
Sounds good. Just let me know the location and I will move it there.

sharathns93:
@avijit-nervana yes it is self-contained. will move it to `test/`

sharathns93:
TESTNOW

sharathns93:
TESTNOW

aprocter:
Judging by the CI logs I think the script is getting caught up by the `pytest` run somehow. Maybe it needs to reside in a different directory since it is not a proper `pytest` test.

PR #156, opened 2018-08-21
Igor/op sqrt

megawidget:


src/ngraph_confirm_pass.cc
```diff
@@ -299,6 +299,7 @@ class NGraphConfirmPass : public tensorflow::GraphOptimizationPass {
         type_constraint_map["Split"]["T"] = NGraphDTypes();
         type_constraint_map["SplitV"]["T"] = NGraphDTypes();
         type_constraint_map["SplitV"]["Tlen"] = NGraphIndexDTypes();
+        type_constraint_map["Sqrt"]["T"] = NGraphDTypes();
```
aprocter: Should be `NGraphNumericDTypes`, I believe

megawidget: OK. I was wondering about that.  `Rsqrt` is also set to this. :/

PR #157, opened 2018-08-21
Upgraded to nGraph 0.6 version. Also added a version number for the P…

avijit-nervana:
…ython whl.


avijit-nervana:
Closed as we created a new PR with the same objective. This became problematic.

PR #158, opened 2018-08-21
Ngraph C/C++ APIs

megawidget:
Whoops, python bindings included. :|


python/ngraph/__init__.in.py
```diff
@@ -34,19 +34,40 @@
  
 import ctypes
 
-# TODO(amprocte): need some way to detect double-load of plugin.
```
aprocter: @megawidget pointed out to me in a Slack chat that this is not actually a problem. (Python modules themselves won't "double-load".)

megawidget: Caveat: They can, if they're being pulled from different paths (e.g. if the module is symlinked somewhere else and imported by a different path in another file).  If we do this we're architecturally in serious trouble however. :p

avijit-nervana: So then - is there a to prevent this?

megawidget: this is what the hack would look like, i can add it if you like:
```
import sys, ctypes, os
from platform import system


for name, module in sys.modules.items():
  if name == 'ngraph' or name.endswith('.ngraph'):
    try:
      if isinstance(module.ngraph, ctypes.CDLL):
        raise ImportError("ngraph has been imported twice")
    except AttributeError:
      pass

ext = 'dylib' if system() == 'Darwin' else 'so'
libpath = os.path.dirname(__file__)
ngraph = ctypes.cdll.LoadLibrary(os.path.join(libpath,
                                              'libngraph_device.' + ext))
```

PR #160, opened 2018-08-22
Langjian/batch norm bug fix

jianyinglang:


avijit-nervana:
The tests failed due to code format check. I will fix this and push it upstream.

jianyinglang:
Thanks for the fix!

PR #163, opened 2018-08-23
Backend API

IcarusMJ12:


megawidget:
Note PR is against my other branch for now

megawidget:
Another thought: should I return `tf::Status` instead in the C++ API?

aprocter:
> Another thought: should I return tf::Status instead in the C++ API?

Yup, I think that'd be a good idea.

python/ngraph/__init__.in.py
```diff
@@ -71,3 +73,20 @@ def disable():
...
+
+def backends_len():
+  return ngraph.ngraph_backends_len()
+
+
+def list_backends():
+  len_backends = backends_len()
+  result = (ctypes.c_string * len_backends)(*(None * len_backends))
+  if not ngraph.ngraph_build_backends(result, len_backends):
+    raise Exception("Backends fluctuated while listing")
```
aprocter: Do we expect this to happen?

test/python/test_api.py
```diff
@@ -32,3 +34,17 @@ def test_disable(self, ngraph_device):
...
     assert ngraph_device.ngraph_is_enabled() == 1
+
+  def test_backends_len(self, ngraph_device):
+    assert ngraph_device.ngraph_backends_len()
+
+  def test_set_backend(self, ngraph_device):
+    assert ngraph_device.ngraph_set_backend(b"CPU")
+
+  def test_set_backend_invalid(self, ngraph_device):
+    assert ngraph_device.ngraph_set_backend(b"POTATO") == 0
```
aprocter: You haven't heard about Project POTATO?

megawidget: I have not :(

python/ngraph/__init__.in.py
megawidget: it seems very, very unlikely, but who knows

PR #164, opened 2018-08-23
Fix for crash when inter_op_parallelism_threads>1

aprocter:
See NGRAPH-2631 and (maybe) NGRAPH-2099. Marking WIP because I want to evaluate whether the locks around the `create_tensor` calls are too conservative. (In theory they are not needed.)


aprocter:
It looks like the `create_tensor` locks were not necessary, which is nice, so I've removed them. Have also smuggled in the fix from #161.

PR #165, opened 2018-08-23
Fix default run-type for docker container command line runs

crlishka:
Fix for the default run-type for running commands (inference models) in a docker container.


crlishka:
TESTNOW

PR #167, opened 2018-08-24
Standardized version strings between python 2 and 3

megawidget:


python/ngraph/__init__.in.py
```diff
@@ -29,31 +29,49 @@
...
+TF_GIT_VERSION = tf.GIT_VERSION
+TF_VERSION_NEEDED = "${TensorFlow_GIT_VERSION}"
+
+try:
+  TF_VERSION = str(TF_VERSION, 'ascii')
+except TypeError:  # will happen for python 2 or  if already string
+  pass
+
+try:
+  if TF_GIT_VERSION.startswith("b'"):  # oh god...
```
avijit-nervana: Please do not use comments that do not mean anything.

python/ngraph/__init__.in.py
```diff
@@ -29,31 +29,49 @@
 from tensorflow.python.client import device_lib
 from tensorflow.python import pywrap_tensorflow as py_tf
 from tensorflow.python.framework import errors_impl
+
+TF_VERSION = tf.VERSION
```
avijit-nervana: Please add some comments explaining what's being done and why.

PR #169, opened 2018-08-27
Avijit/r0.4 finalize

avijit-nervana:
We are going to use this PR to finalize whatever is needed to lock r0.4 and make a release.


avijit-nervana:
Wrong base branch!!!!

PR #170, opened 2018-08-27
r0.4 finalize

avijit-nervana:


crlishka:
I checked the failure on this job (even though it is old), and I think it may be a false-negative.  I am going to start another round of testing.

crlishka:
TESTNOW

PR #172, opened 2018-08-27
log node placement

megawidget:
Not sure if this is the desired solution.  I considered using callbacks but adam thinks it's overkill.


src/ngraph_mark_for_clustering.cc
```diff
@@ -381,6 +381,11 @@ Status MarkForClustering(Graph* graph) {
         NGRAPH_VLOG(4) << "Rejecting: " << node->name() << "["
                        << node->type_string() << "]";
       }
+
+      if (config::is_logging_placement()) {
+        const char* status = confirmed ? "Accepting: " : "Rejecting: ";
```
aprocter: If we want to log the final cluster assignment, it might make more sense to do this in the `DeassignClusters` pass.

PR #173, opened 2018-08-28
Shrestha/add unit test infrastructure

shresthamalik:
Added Unit Test Infrastructure
- Supports ops that expect static inputs as well
- Review Comment changes are pending


test/opexecuter.h
```diff
@@ -0,0 +1,103 @@
...
+             const vector<DataType>& op_types, const vector<Output>& fetch_ops);
+
+  ~OpExecuter();
+
+  void ExecuteOnNGraph();
+  void ExecuteOnTF();
+  // To Do : Overload CompareNGraphAndTF() to take in tolerance
+  void CompareNGraphAndTF();
+  void CompareNGraphAndTF(float tolerance);
+  // Executes on NGraph, executes on TF, compares the results
```
avijit-nervana: We probably don't need the functions above separately - at least for now. Please make them private.

test/test_utilities.cpp
```diff
@@ -0,0 +1,91 @@
...
+using namespace std;
+
+namespace ng = ngraph;
+
+namespace tensorflow {
+
+namespace ngraph_bridge {
+
+void ActivateNGraph() {
+  setenv("NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS", "1", 1);
```
avijit-nervana: Once @igor checks in his API changes, the implementation of these functions will change. That would be a separate PR though.

test/opexecuter.cpp
```diff
@@ -0,0 +1,359 @@
...
+    } else if (node->type_string() == test_op_type_) {
+      // only one node of type test_op
+      ASSERT_FALSE(found_test_op);
+      found_test_op = true;
+    } else {
+      ASSERT_TRUE(node->type_string() == allowed_nodes[0])
+          << "Found Not allowed Op: " << node->type_string();
+    }
+  }
+
```
sayantan-nervana: The asserts made on 'found_test_op' can probably be made in GetNodeData. In that case the ValidateGraph function can be removed

test/test_math_ops.cpp
```diff
@@ -0,0 +1,104 @@
...
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 2;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+  Tensor B(DT_FLOAT, TensorShape({dim1, dim2}));
+
+  AssignInputValues(A, 2.0f);
+  AssignInputValues(B, 7.0f);
+
```
sayantan-nervana: This assigns constancy values to all locations in the tensor right? Can we have one of the tests with varying inputs as an example

test/test_nn_ops.cpp
```diff
@@ -0,0 +1,124 @@
...
+namespace testing {
+
+#define ASSERT_OK(x) ASSERT_EQ((x), ::tensorflow::Status::OK());
+
+// Test(TestCaseName, TestName)
+// Please ensure
+// Neither TestCaseName nor TestName should contain underscore
+// https://github.com/google/googletest/blob/master/googletest/docs/primer.md
+// Use only Tensors and ops::Const() to provide input to the test op
+
```
sayantan-nervana: Interesting! Thanks for going through the test documentation in so much detail.
Our earlier tests contains scores of underscores

test/test_utilities.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+void DeactivateNGraph() {
+  unsetenv("NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS");
+  setenv("NGRAPH_TF_DISABLE", "1", 1);
+}
+
+void AssertTensorEquals(Tensor& T1, Tensor& T2) {
+  auto T_size = T1.flat<float>().size();
+  auto T1_data = T1.flat<float>().data();
```
sayantan-nervana: We should also add an assert on the shape here.

test/test_utilities.cpp
```diff
@@ -0,0 +1,91 @@
...
+    auto b = T2_data[k];
+    if (a == 0) {
+      EXPECT_NEAR(a, b, tol);
+    } else {
+      auto rel = a - b;
+      auto rel_div = std::abs(rel / a);
+      EXPECT_TRUE(rel_div < tol);
+    }
+  }
+}
```
sayantan-nervana: AssertTensorEquals and ValidateTensorData look similar. Do they have different use cases? Can we remove one and make do with the other?

test/opexecuter.cpp
```diff
@@ -0,0 +1,359 @@
...
+    NGRAPH_VLOG(5) << "Edge between, Src: " << e->src()->name()
+                   << " Src op index " << e->src_output()
+                   << " ,Dst: " << e->dst()->name() << " dst ip index "
+                   << e->dst_input();
+    // update src's outedge metadata
+    node_outedge_md[e->src()].push_back({e->dst(), e->dst_input()});
+    node_inedge_md[e->dst()].push_back({e->src(), e->src_output()});
+    node_outedges[e->src()].push_back(e);
+  }
+}
```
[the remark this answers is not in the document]
shresthamalik: Yes correct, Thanks for asking. My first shot while implementing this was by iterating over the nodes. 

1. However, few apis node->in_edges() and node->out->edges() apis were not reliable. For e.g. their is an edge from SRC->SINK, which shows up as in_edge() for SINK but not as out_edge() of SRC. Similarly, there is an edge from test_Op->SINK , which does not show as out_edge() for test_op and many more. There was not a clear pattern regarding this behavior. In interest of time, i decided to switch to iterating over edges. 

2. To fetch the test_op_node, we could iterate again over the nodes. But then this function would involves two iterations. Once over the edges, another over the nodes.

test/opexecuter.cpp
shresthamalik: That's correct, it can be moved there. The primary reason for implementing like this was to have clear responsibility for each function.  Since, the class (unit test) has well defined constraints (Only "Const" nodes + 1 "test_op_type" node) on what kind of graph it can accept, I decided to have a separate function for validating, and only then process a validated graph.

test/test_math_ops.cpp
shresthamalik: Yes I agree, planning to add another utility function to generate random data. This is a borrowed function from the current unit tests.

test/test_nn_ops.cpp
shresthamalik: Thanks. Yup, we had underscores there. :)

test/test_utilities.cpp
shresthamalik: Yes, most review comments from crucible have not been addressed yet. Definitely in the next PR.

test/test_utilities.cpp
shresthamalik: One accepts tolerance as an additional input. For some ops, we might want to relax the tolerance by some margin. Ideally AssertTensorEquals() should be overloaded to take in tolerance. Planning in next PR.

test/test_utilities.cpp
shresthamalik: Will keep a note on this.

test/opexecuter.h
shresthamalik: For some ops, we need to run NHWC on tf and NCHW on NGraph and compare. For such tests, these functions would come in handy, so kept them public. Please see my comment here:
https://aipg-rancher.intel.com/crucible/cru/AIPG-11#c27

This is open for discussion and more ideas.

sayantan-nervana:
Yes, sounds good to me

test/test_utilities.cpp
```diff
@@ -0,0 +1,92 @@
...
+  unsetenv("NGRAPH_TF_DISABLE");
+}
+
+void DeactivateNGraph() {
+  unsetenv("NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS");
+  setenv("NGRAPH_TF_DISABLE", "1", 1);
+}
+
+void AssertTensorEquals(Tensor& T1, Tensor& T2) {
+  ASSERT_EQ(T1.shape(), T2.shape());
```
shresthamalik: Added the shape match

test/test_utilities.cpp
```diff
@@ -0,0 +1,92 @@
...
+    A_flat_data[i] = x;
+  }
+}
+
+void PrintTensor(const Tensor& T1) {
+  LOG(INFO) << "print tensor values" << T1.DebugString();
+}
+
+void ValidateTensorData(Tensor& T1, Tensor& T2, float tol) {
+  ASSERT_EQ(T1.shape(), T2.shape());
```
shresthamalik: Added shape match

PR #176, opened 2018-08-29
Sarkars/builder

sayantan-nervana:
Please take a look at the builder refactoring change.

1) Switching from old to new builder.
The killswitch is a primeval C style macro in encapsulate_op.cc. Maybe there is a better/preferable method. But I used it instead of if(flag) because, when we enable the newbuilder at a later stage, we can just delete the stuff in the #if-#else en masse and not have to modify any more code.
Currently, the old builder is in use, so nothing should fall apart.

#define BUILDER1 // (un)comment this line to (de)activate old builder

#ifdef BUILDER1
newbuilder
#else
oldbuilder
#endif

2) I will continue adding more TranslateOps functions in the meanwhile

3) Since its a big change, I am sure there are style issues, questionable design/syntax choices, outright unoptimised chunks and sneaky leaks. Your feedback is very important and appreciated.


src/ngraph_builder1.cc
```diff
@@ -0,0 +1,370 @@
...
+
+  VectNg ng_result_list;
+  TF_RETURN_IF_ERROR(GetOutputNodes(tf_ret_vals, ng_result_list));
+
+  // Create the nGraph function.
+  try {
+    ng_function = make_shared<ng::Function>(ng_result_list, ng_parameter_list);
+  } catch (...) {
+    return errors::Unimplemented("Unable to create nGraph function");
+  }
```
sayantan-nervana: Adding a try-catch here. Currently this is a catch-all, but in case ngraph has specific error classes, we can provide different messages for each case.

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+  Status GetInputNode(const Node*, size_t, shared_ptr<ng::Node>*);
+
+  // Classify a list of TF nodes into _Arg (input), _Retval (output) and other
+  // nodes. Used to populate tf_params, tf_ret_vals, tf_ops
+  Status ClassifyNodes(const vector<Node*>&);
+
+  // Given the input shapes and a list of TF _Arg nodes, create corresponding
+  // nGraph parameters. Also populate the ng_op_map
+  Status GetInputParams(const std::vector<TensorShape>&, vector<const Node*>,
+                        vector<shared_ptr<ng::op::Parameter>>&);
```
aprocter: I think TF style favors pointers over non-const references (not 100% sure though)

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+  // from 0 to num_inputs-1
+  const static std::map<const string, vector<int>> INPUT_INDEX_MAP;
+
+  // A map from Tf op type_string to a TranslateOp
+  const static std::map<const string, Builder1::TranslatorFn> TRANSLATE_OP_MAP;
+
+  // Given a TF node, return its corresponding TranslateOp function and required
+  // input indexes. A wrapper for TRANSLATE_OP_MAP and INPUT_INDEX_MAP
+  Status GetOpTranslationRequirements(const Node*, Builder1::TranslatorFn&,
+                                      vector<int>&);
```
aprocter: I think TF style favors pointers over non-const references (not 100% sure though)

src/ngraph_builder1.cc
```diff
@@ -0,0 +1,389 @@
...
+  return Status::OK();
+}
+
+Status Builder1::TranslateGraph(
+    OpKernelContext* ctx, std::shared_ptr<ngraph::Function>& ng_function) {
+  TF_RETURN_IF_ERROR(Initialize());
+
+  std::vector<const Tensor*> static_input_map;
+
+  std::vector<TensorShape> inputs(ctx->num_inputs());
```
aprocter: `input_shapes` might be a more descriptive name

src/ngraph_translateops.h
```diff
@@ -0,0 +1,269 @@
...
+      // https://www.tensorflow.org/api_guides/python/nn#Notes_on_SAME_Convolution_Padding
+      int64 out_shape = ceil(ng_image_shape[i] / ng_strides[i]);
+      int64 padding_needed = ng_strides[i] * (out_shape - 1) +
+                             ng_kernel_shape[i] - ng_image_shape[i];
+      padding_needed = padding_needed < 0 ? 0 : padding_needed;
+
+      size_t padding_lhs = padding_needed / 2;
+      size_t padding_rhs = padding_needed - padding_lhs;
+      ng_padding_below[i] = padding_lhs;
+      ng_padding_above[i] = padding_rhs;
```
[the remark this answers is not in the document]
aprocter: I have a dim recollection that I may have come across that document before, taken the formula as it was, and found out that it wasn't quite correct. If I'm wrong, let's definitely use the version from the docs.

src/ngraph_builder1.cc
```diff
@@ -0,0 +1,389 @@
...
+    int num_inputs = variadic_input ? op->num_inputs() : input_indexes.size();
+    VectNg ng_arg_vec(num_inputs);
+    if (op->type_string() != "Const") {
+      for (int idx = 0; idx < num_inputs; idx++) {
+        TF_RETURN_IF_ERROR(GetInputNode(
+            op, (variadic_input ? idx : input_indexes[idx]), &ng_arg_vec[idx]));
+      }
+    }
+    TF_RETURN_IF_ERROR(
+        translate_fn(op, ng_arg_vec, static_input_map, subgraph_out_nodes));
```
aprocter: I wonder if it would be possible to extract and pass just the static input tensors that are actually being used by this node, by examining `_ngraph_static_inputs` (or whatever we called it) here?

src/ngraph_translateops.h
```diff
@@ -0,0 +1,269 @@
...
+      std::make_shared<ng::op::Floor>(ng_arg_vec[0] / ng_arg_vec[1]);
+  subgraph_out_nodes[0] = ng_arg_vec[0] - (floordiv * ng_arg_vec[1]);
+
+  // Possible reuse of floordiv:
+  // TranslateFloorDivOp1(op, ng_arg_vec, static_input_map, subgraph_out_nodes);
+  // subgraph_out_nodes[0] = ng_arg_vec[0] - (subgraph_out_nodes[0] *
+  // ng_arg_vec[1]);
+  return Status::OK();
+}
+}
```
aprocter:
```
} // namespace ngraph_bridge
} // namespace tensorflow
```

src/ngraph_builder1.cc
aprocter: Try feeling this idea out as you implement some ops which have static inputs (like, say, convolution backprop)

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+
+  // Prevent Initialize from running twice
+  bool is_initialized = false;
+  const Graph& tf_graph;
+
+  // An array that will be useful in creating the static_input_map
+  std::vector<bool> m_input_is_static;
+
+  // Vectors containing TF nodes in 3 bins: inputs, outputs, actual ops
+  vector<const Node *> tf_params, tf_ret_vals, tf_ops;
```
aprocter: As you refactor this feel free to revisit the question of whether and where the `tf_` and `ng_` prefixes are necessary. I think they're really only needed where there's a need to distinguish between TensorFlow's notion of a `Foo` and nGraph's notion of a `Foo` (e.g., TF functions and nGraph functions).

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+// equivalent ngraph fucntion. To that end, it acts as an interface to
+// a library of 'TranslateOps' functions. These functions are defined
+// in ngraph_translateops.h. Builder just goes through each TF op and
+// calls the appropriate 'TranslatorFn'. It does not implement the per-op
+// translation, the heavy-lifting being done by ngraph_translateops.h;
+// It merely does some bookkeeping, such as process the nodes in correct
+// order, keep track of and retrieve parent nGraph nodes etc.
+
+class Builder1 {
+  using VectNg = std::vector<shared_ptr<ng::Node>>;
```
aprocter: Nitpick: I'm not a fan of this particular instance of `using`. (It obscures the fact that the elements are pointers, and that they are pointers to nodes, for one thing.)

sayantan-nervana: How about VectPtrNg?
This std::vector<shared_ptr<ng::Node>> gets used a lot, so wanted some typedef

src/ngraph_builder1.cc
sayantan-nervana: the padding issue: I ran the math on both implementations, they *seem* to be the same. But who knows what crazy corner case might show up. In case the current implementation works, lets keep that

src/ngraph_translateops.h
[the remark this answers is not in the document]
sayantan-nervana: the padding issue: I ran the math on both implementations, they seem to be the same. But who knows what crazy corner case might show up. In case the current implementation works, lets keep that

src/ngraph_translateops.h
sayantan-nervana: I thought clang-format automatically tags that namespace comment.

from what I see in clang-format's source here: 
https://github.com/llvm-mirror/clang/blob/42c483011b56fbd5d5aad7bad53135190b2d693a/lib/Format/Format.cpp#L729
Google style copies LLVM style and modifies some fields. It does not modify the 'FixNamespaceComments' field which is set to true. So clang-format *should* add these comment lines when we apply style/format. FixNamespaceComments's description can be found here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
It seems with the current combo of clang-format-3.9 + google-style does not do that automatically.

I tried adding "FixNamespaceComments: true" to graph-tf's .clang-format file, but that produced some errors indicating that clang-format-3.9 does not recognize FixNamespaceComments. That seems to indicate that FixNamespaceComments is a new field that is still not supported by clang-format-3.9

TL;DR: clang-format does not add this comment for now. we have to take care to add it on our own

src/ngraph_translateops.h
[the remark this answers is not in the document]
sayantan-nervana: added new commit keeping old formula

src/ngraph_translateops.h
sayantan-nervana: Added commit to add comments

src/ngraph_builder1.h
sayantan-nervana: removed the typedef

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+  Status TranslateGraph(const std::vector<TensorShape>&,
+                        const std::vector<const Tensor*>&,
+                        shared_ptr<ng::Function>&);
+
+ private:
+  // The op map holds a mapping from TensorFlow op names (strings) to
+  // vector of generated nGraph nodes. Since we process nodes in toposort
+  // order, it is guaranteed when processing a TF node, its parents
+  // will already have been processed and safely stowed in ng_op_map
+  std::unordered_map<std::string, VectNg> ng_op_map;
```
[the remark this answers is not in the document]
sayantan-nervana: added m_ to non-static members.
The 2 statics I have are:
  const static std::map<const string, vector<int>> INPUT_INDEX_MAP;
  const static std::map<const string, Builder1::TranslatorFn> TRANSLATE_OP_MAP;

TRANSLATE_OP_MAP was in all-caps in the old builder, copied that style. Not sure if adding s_ in front of them will look good. I could lowercase them and then add the s_

src/ngraph_builder1.cc
sayantan-nervana: Added this

src/ngraph_builder1.h
sayantan-nervana: The 'GetInputParams' and 'GetOutputNodes' works with ng and tf nodes, so I think the tf_ or ng_ prefix is useful here to distinguish between ng in/outs and tf in/outs.

src/ngraph_builder1.h
sayantan-nervana: https://google.github.io/styleguide/cppguide.html#Reference_Arguments

It seems to prefer a const T&, and never a raw T&. It prefer's T* over T& for 'return' values.

But, our code makes a lot of use of T& for returning. For example TranslateGraph itself uses std::shared_ptr<ngraph::Function>& to return the created ng_function. 

The 'cons' of not using T& as stated in the style guide is:
References can be confusing, as they have value syntax but pointer semantics.

I am not sure if addressing that con is worth the effort of overhauling all places where we have already used T& for returning values

src/ngraph_builder1.h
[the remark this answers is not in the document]
sayantan-nervana: lowercased in new commit

sayantan-nervana:
I think most of Adam's review comments are addressed or incorporated. The two things remaining are:
1) T& vs T* for return. TF style prefers T* for return/changeable variables. references are preferred with consts. But in out code, we use TranslateGraph that uses non-const T&.
2) Adding MakeNGraphNode as suggested by Adam in Slack. Should we do this in a new PR?

Any other thing that I missed in the above summary?

avijit-nervana:
Comments below.

sayantan-nervana:
Ok, will make the T& -> T* changes. Also tagged you in slack

src/ngraph_builder1.cc
```diff
@@ -0,0 +1,396 @@
...
+    //     index
+    // (2) for each _Arg node n, set m_input_is_static[n.index] to true if n
+    //     is driving any static input; else set it to false.
+    //
+
+    // Create the vector.
+    int32 max_arg_index = -1;
+    std::vector<const Node*> arg_nodes;
+
+    for (auto node : m_tf_graph.nodes()) {
```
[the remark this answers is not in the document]
shresthamalik: Going by the GetInputParams() function, max_arg_index = m_tf_params.size()-1 .. In that case we dont need this loop. And, If the _Arg indexes are not predictable, i.e. in sequence (0 to n-1) then GetInputParams() would need changes, ng_parameter_list[index] = ng_param; index might be greater than size.

src/ngraph_builder1.h
```diff
@@ -0,0 +1,138 @@
...
+  Status TranslateGraph(const std::vector<TensorShape>&,
+                        const std::vector<const Tensor*>&,
+                        shared_ptr<ng::Function>&);
+
+ private:
+  // The op map holds a mapping from TensorFlow op names (strings) to
+  // vector of generated nGraph nodes. Since we process nodes in toposort
+  // order, it is guaranteed when processing a TF node, its parents
+  // will already have been processed and safely stowed in ng_op_map
+  std::unordered_map<std::string, vector<shared_ptr<ng::Node>>> m_ng_op_map;
```
shresthamalik: Google Style Guide suggests adding trailing _ to all private data members of a class. nGraph core follows the style adopted here. 
https://google.github.io/styleguide/cppguide.html#Variable_Names
@avijit-nervana which one is recommended? Actually we can do both, by adding _ to all the variables.

sayantan-nervana: The m_ means data member, and s_ means static members. According to Adam, ngraph follows m_ and s_, so I have that format. Not sure if we stick to (m_, s_) or trailing _

avijit-nervana: @sayantan-nervana When you get a chance - please resolve the conflict.

PR #180, opened 2018-08-30
Shrestha/fix bug strided slice shrink mask

shresthamalik:
Treat ShrinkAxisMask as a mask and not a boolean. It's a very small fix and does not handle all cases. E.g., final shape does not account for the strided vector or the ordering. More work required for full implementation. PR to enable testing of models.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1300,9 +1300,8 @@ static Status TranslateFillOp(
     ng_output_shape[i] = dims_vec[i];
     ng_axis_set.insert(i);
   }
-  SaveNgOp(
-      ng_op_map, op->name(),
-      make_shared<ng::op::Broadcast>(ng_value, ng_output_shape, ng_axis_set));
+  SaveNgOp(ng_op_map, op->name(), make_shared<ng::op::Broadcast>(
+                                      ng_value, ng_output_shape, ng_axis_set));
   return Status::OK();
 }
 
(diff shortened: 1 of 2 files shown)


test/python/test_slice.py
```diff
@@ -61,26 +60,26 @@ def test_strided_slice(self):
...
     slice_ts.append(x[:, ...])
     slice_ts.append(x[1:, :-2])
     slice_ts.append(x[::2, :-2])
     slice_ts.append(x[1, :])
-    #slice_ts.append(x[:, 1])
+    slice_ts.append(x[:, 1])
     slice_ts.append(x[1, 1])
     slice_ts.append(x[0])
     slice_ts.append(x[0][1])
     slice_ts.append(x[-1])
```
sayantan-nervana: Are these all the slicing cases that we can possibly see. Even if we do not support them now, can we write them down and comment them out as future goals or known unsupported features.
For one, I think this is not supported: y = x[:, tf.newaxis]

src/ngraph_builder.cc
```diff
@@ -2437,19 +2437,42 @@ static Status TranslateStridedSliceOp(
...
+
+    NGRAPH_VLOG(3) << "Shrink axis mask " << tf_shrink_axis_mask;
+    NGRAPH_VLOG(3) << " Shrink axis " << ng::join(shrink_axis);
+
+    vector<size_t> output_shape;
+    for (int i = 0; i < lower_vec.size(); i++) {
+      if (shrink_axis.find(i) == shrink_axis.end()) {
+        output_shape.push_back(end_vec[i] - lower_vec[i]);
+      }
+    }
```
sayantan-nervana: The two loops can be combined:
for (int i = 0; i < lower_vec.size(); i++){
if ((shrink_axis_mask>>i & 1) == 1){
  output_shape.push_back(end_vec[i] - lower_vec[i]);
}
}

 If we feel 'shrink_axis' might be used later for other cases, then we might the 2 loop approach.

test/python/test_slice.py
shresthamalik: Yes, there are many such cases that are unsupported. It will actually take me a while to jot down all the unsupported test cases.

We can think of this PR to enable test cases where we are shrinking one or more axis.
slice_ts.append(x[:,1]) or 
for a (2,4,5) shaped tensor
slice_ts.append(x[1,:,:])

src/ngraph_builder.cc
shresthamalik: yes. Thanks. I will update this.

test/python/test_slice.py
shresthamalik: Added x[:, tf.newaxis] as currently unsupported

src/ngraph_builder.cc
```diff
@@ -2437,19 +2437,35 @@ static Status TranslateStridedSliceOp(
...
+    int64 shrink_axis_mask = tf_shrink_axis_mask;
+    vector<size_t> output_shape;
+
+    for (int i = 0; i < lower_vec.size(); i++) {
+      if ((shrink_axis_mask & 1) != 1) {
+        output_shape.push_back(end_vec[i] - lower_vec[i]);
+      }
+      shrink_axis_mask >>= 1;
+    }
+
```
shresthamalik: Using Single loop

PR #181, opened 2018-08-31
Rewrite nn op test using new test framework

mingshan-wang:
To support:
1. BiasAddGrad
2. Op_L2Loss
3. Op_RealDiv


test/test_nn_ops.cpp
```diff
@@ -52,31 +52,69 @@ namespace testing {
 // https://github.com/google/googletest/blob/master/googletest/docs/primer.md
 // Use only Tensors and ops::Const() to provide input to the test op
 
-TEST(NNOps, SparseSoftmaxCrossEntropyWithLogits) {
-  Scope root = Scope::NewRootScope();
-  int batch = 10;
-  int num_of_classes = 2;
+// The backword operation for "BiasAdd" on the bias tensor.
```
shresthamalik: small typo 'backward'. Nice comments! :)

test/test_nn_ops.cpp
```diff
@@ -118,6 +156,56 @@ TEST(NNOps, Conv2DBackpropFilter_NHWC) {
...
+  input_sizes.push_back({2, 3, 4});
+  input_sizes.push_back({0});
+
+  vector<int> static_input_indexes = {};
+
+  for (auto const& input_size : input_sizes) {
+    Scope root = Scope::NewRootScope();
+
+    Tensor input_data(DT_FLOAT, TensorShape(input_size));
+    AssignInputValues(input_data, 0.0);
```
shresthamalik: Please assign some other value, other than 0.0f.

test/test_nn_ops.cpp
mingshan-wang: Thanks for catching that.

avijit-nervana:
TESTNOW

test/test_nn_ops.cpp
```diff
@@ -118,7 +148,159 @@ TEST(NNOps, Conv2DBackpropFilter_NHWC) {
   }
 }
 
-}  // namespace testing
+// NCHW data format not supported on the test framework now
+TEST(NNOps, DISABLED_Conv2DBackpropFilterNCHW) {
```
shresthamalik: Can we move this disabled test to tf_Exec.cc temporarily? Let's not add any disabled test in the new framework, to avoid confusion. @avijit-nervana What do you think? We can add it back when the unit test framework supports it.

test/test_utilities.cpp
```diff
@@ -66,6 +68,32 @@ void AssignInputValues(Tensor& A, float x) {
...
+void AssignInputValuesAnchor(Tensor& A, float x) {
+  auto A_flat = A.flat<float>();
+  auto A_flat_data = A_flat.data();
+  for (int i = 0; i < A_flat.size(); i++) {
+    A_flat_data[i] = x * i;
+  }
+}
+
+// Randomly generate a float number between -0.0005 ~ 0.0005
+void AssignInputValuesRandom(Tensor& A) {
```
shresthamalik: The comment says this function generates a float between -0.0005 ~0.0005. Following the other comments, the range seems different.

shresthamalik: Also it would be helpful to have a more generic function that could generate numbers between any range

shresthamalik:
Nice, detailed comments. Minor change requested

test/test_nn_ops.cpp
avijit-nervana: That's a good suggestion. But this DISABLED test uses the new framework - so we probably should keep this here. We should retire the tf_exec.cpp once all the tests in there are converted to the new framework.

shresthamalik:
Yes, I thought I was closing my review.

test/test_utilities.cpp
mingshan-wang: Yes, I found out this discrepancy in comments yesterday, accidentally pushed to another branch. Will do a pull request with that branch today.
Sure, I am thinking about that too. I will do another pull request regarding adding another function to random generate numbers in range. Thanks @shresthamalik .

mingshan-wang:
TESTNOW

PR #185, opened 2018-09-05
Langjian/fused batch nor grad training only

jianyinglang:
Add two additional outputs in FusedBatchNormGrad op to match the Tensorflow output.
Mark `is_training=False` as a rejection for FusedBatchNormGrad op.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1489,11 +1489,13 @@ static Status TranslateFusedBatchNormGradOp(
   SaveNgOp(ng_op_map, op->name(), ng_beta_delta_op);
   // Output reserve_space_3: Unused placeholder to match the mean input 
   // in FusedBatchNorm.
-  std::shared_ptr<ng::Node> output_mean;
+  std::shared_ptr<ng::Node> output_mean = make_shared<ngraph::op::Constant>(
+      ng_mean->get_element_type(), ng::Shape{}, std::vector<std::string>{""});
   SaveNgOp(ng_op_map, op->name(), output_mean);
   // Output reserve_space_4: Unused placeholder to match the variance input 
   // in FusedBatchNorm.
-  std::shared_ptr<ng::Node> output_variance;
+  std::shared_ptr<ng::Node> output_variance = make_shared<ngraph::op::Constant>(
+      ng_variance->get_element_type(), ng::Shape{}, std::vector<std::string>{""});
   SaveNgOp(ng_op_map, op->name(), output_variance);
 
   return Status::OK();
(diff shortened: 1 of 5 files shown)


avijit-nervana:
The reason for the failure is clang-format issues. The way to fix this is to run the following script from the top level directory:
maint/apply-code-format.sh

I have made this change and pushed it.

mingshan-wang:
Looks good to me too. Feel free to merge it.

src/ngraph_builder.cc
```diff
@@ -1487,6 +1487,17 @@ static Status TranslateFusedBatchNormGradOp(
   SaveNgOp(ng_op_map, op->name(), ng_input_delta_op);
   SaveNgOp(ng_op_map, op->name(), ng_scale_delta_op);
   SaveNgOp(ng_op_map, op->name(), ng_beta_delta_op);
+  // Output reserve_space_3: Unused placeholder to match the mean input
+  // in FusedBatchNorm.
+  std::shared_ptr<ng::Node> output_mean = make_shared<ngraph::op::Constant>(
+      ng_mean->get_element_type(), ng::Shape{}, std::vector<std::string>{""});
```
aprocter: Are you sure scalar is the right shape here?

src/ngraph_mark_for_clustering.cc
```diff
@@ -272,8 +272,17 @@ Status MarkForClustering(Graph* graph) {
       confirmation_functions["FloorDiv"] = SimpleConfirmationFunction();
       confirmation_functions["FloorMod"] = SimpleConfirmationFunction();
       confirmation_functions["FusedBatchNorm"] = SimpleConfirmationFunction();
-      confirmation_functions["FusedBatchNormGrad"] =
-          SimpleConfirmationFunction();
+      confirmation_functions["FusedBatchNormGrad"] = [](Node* n, bool* result) {
+        // Reject if "is_training" is set false.
+        bool tf_is_training;
+        TF_RETURN_IF_ERROR(
+            GetNodeAttr(n->attrs(), "new_axis_mask", &tf_is_training));
```
aprocter: `"new_axis_mask"` -> `"is_training"`

src/ngraph_builder.cc
jianyinglang: We keep it a scalar for now to match the empty tensor output in tensorflow.

src/ngraph_mark_for_clustering.cc
jianyinglang: Changed. Thanks for catching this!

PR #190, opened 2018-09-08
using weakptr for backend

fengleitian:
Use weak_ptr to hold ng_backend. The static shared_ptr cause segfault during destruction.


src/ngraph_encapsulate_op.cc
```diff
@@ -479,15 +478,13 @@ class NGraphEncapsulateOp : public OpKernel {
   int m_ngraph_cluster;
   std::vector<bool> m_input_is_static;
 
-  static std::shared_ptr<ng::runtime::Backend> s_ng_backend
-      GUARDED_BY(s_ng_backend_mutex);
+  static std::weak_ptr<ng::runtime::Backend> s_ng_backend_wptr;
+  std::shared_ptr<ng::runtime::Backend> s_ng_backend;
```
rkimballn1: Change to m_ng_backend

src/ngraph_encapsulate_op.cc
```diff
@@ -123,24 +123,23 @@ class NGraphEncapsulateOp : public OpKernel {
...
+      if (s_ng_backend_name.empty()) {
+        s_ng_backend_name = "CPU";
       }
+    } else {
+      s_ng_backend_name = "CPU";
     }
+    m_ng_backend = ng::runtime::Backend::create(s_ng_backend_name);
+    OP_REQUIRES(ctx, m_ng_backend != nullptr,
+                errors::InvalidArgument("Cannot create nGraph backend"));
+    s_ng_backend_wptr = m_ng_backend;
```
aprocter: I think there's a possibility of a race condition with the initialization of `s_ng_backend_wptr` here:

```
t1 calls Backend::create and assigns m_ng_backend
t2 calls Backend::create and assigns m_ng_backend
t2 assigns s_ng_backend_wptr
t1 assigns s_ng_backend_wptr
```

fengleitian: yes, you are right, I was think it will be fine since the assign for weak_ptr is atomic, but I realize the it will be create and assign multiple times. Let me add a mutex.

fengleitian: fixed.

CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ ExternalProject_Add(
         -DNGRAPH_TOOLS_ENABLE=FALSE 
         -DNGRAPH_DEX_ONLY=TRUE         
         -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+        -DNGRAPH_GPU_ENABLE=TRUE
```
aprocter: Can we change this to `-DNGRAPH_GPU_ENABLE=${NGRAPH_GPU_ENABLE}`? Unfortunately not all of our systems are set up to build the GPU backend.

fengleitian: fixed.

src/ngraph_encapsulate_op.cc
```diff
@@ -520,16 +520,16 @@ class NGraphEncapsulateOp : public OpKernel {
   int m_ngraph_cluster;
   std::vector<bool> m_input_is_static;
 
-  static std::shared_ptr<ng::runtime::Backend> s_ng_backend
+  static std::weak_ptr<ng::runtime::Backend> s_ng_backend_wptr;
+  std::shared_ptr<ng::runtime::Backend> m_ng_backend
       GUARDED_BY(s_ng_backend_mutex);
```
csullivan: Just a note: I read up on GUARDED_BY and wondering if we should be locking `s_ng_backend_mutex` prior to tensor creation via `m_ng_backend->create_tensor`?
```
// GUARDED_BY allows the user to specify a particular mutex that should be
// held when accessing the annotated variable. 
```

fengleitian: m_ng_backend will be copy to s_ng_backend_wptr; so the s_ng_backend_wptr, so it will be point to the same resource. That's why s_ng_backend_wptr need to be inside the mutex.

csullivan: Right, I'm wondering something a little different. `m_ng_backend` is `GUARDED_BY(s_ng_backend_mutex)`. The above comment indicates that a lock should be held whenever the annotated variable (`m_ng_backend` in this case) is accessed - for example, when `create_tensor` is called. My guess is that it isn't critical, but I'm not familiar enough with this TF macro to be sure, and so just wanted to point it out.

fengleitian: to my understanding, `create_tensor` is thread safe. Since it does not modify non-local data.

csullivan: Yeah, since we default to device 0, it should be fine. Since we assume a MT environment in the TF bridge, we should probably add `cudaSetDevice(0)` in the GPU_Backend::create_tensor, or the GPU_TensorView to avoid future bugs when we add multi-gpu support and are creating tensors from different threads.

Regarding `GUARDED_BY`, from here: https://clang.llvm.org/docs/ThreadSafetyAnalysis.html it suggests that use of this without the lock will emit a compile-time warning -- but only for clang, I believe.

PR #191, opened 2018-09-08
Sarkars/viewer small fix

sayantan-nervana:


sayantan-nervana:
Sorry my bad. Forgot to update the file names earlier. Tested this one, it works. Thanks to @mingshan-wang for pointing it out.

mingshan-wang:
In ngtf_graph_viewer.py, the bottom sample usages need to rename as well.

PR #192, opened 2018-09-09
Updating some links, removing stale links

sayantan-nervana:


sayantan-nervana:
some links were pointing to ngraph pages, redirected them to ngraph-tf pages.
removed some stale links
updated bazel link

PR #193, opened 2018-09-10
Kanvi/convert python to c++ unit tests

kanvi-nervana:
Adding more C++ unit tests - converting from python

### DIFF
--- a/test/opexecuter.cpp
+++ b/test/opexecuter.cpp
@@ -135,6 +135,9 @@ void OpExecuter::ExecuteOnTF() {
   DeactivateNGraph();
   ClientSession session(tf_scope_);
   ASSERT_EQ(Status::OK(), session.Run(sess_run_fetchoutputs_, &tf_outputs_));
+  for(int i=0; i<tf_outputs_.size(); i++){
+    NGRAPH_VLOG(5) << " TF op " << i << tf_outputs_[i].DebugString();
+  }
 }
 
 // Compares tf_outputs_ with ngraph_outputs_
@@ -349,6 +352,7 @@ void OpExecuter::ExecuteOnNGraph() {
     void* dst_ptr = DMAHelper::base(&output_tensor);
     ng_op_tensors[i]->read(dst_ptr, 0, output_tensor.TotalBytes());
     ngraph_outputs_.push_back(output_tensor);
+    NGRAPH_VLOG(5) << " NGRAPH op " << i << ngraph_outputs_[i].DebugString();
   }
 
 }  // ExecuteOnNGraph
--- a/test/opexecuter.cpp
+++ b/test/opexecuter.cpp
@@ -135,7 +135,7 @@ void OpExecuter::ExecuteOnTF() {
   DeactivateNGraph();
   ClientSession session(tf_scope_);
   ASSERT_EQ(Status::OK(), session.Run(sess_run_fetchoutputs_, &tf_outputs_));
-  for(int i=0; i<tf_outputs_.size(); i++){
+  for (int i = 0; i < tf_outputs_.size(); i++) {
     NGRAPH_VLOG(5) << " TF op " << i << tf_outputs_[i].DebugString();
   }
 }
@@ -144,8 +144,7 @@ void OpExecuter::ExecuteOnTF() {
 void OpExecuter::CompareNGraphAndTF() {
   ASSERT_EQ(tf_outputs_.size(), ngraph_outputs_.size());
   for (int i = 0; i < tf_outputs_.size(); i++) {
-    switch(expected_output_datatypes_[0])
-    {
+    switch (expected_output_datatypes_[0]) {
       case DT_FLOAT:
         AssertTensorEqualsFloat(tf_outputs_[i], ngraph_outputs_[i]);
         break;
@@ -153,7 +152,8 @@ void OpExecuter::CompareNGraphAndTF() {
         AssertTensorEqualsInt32(tf_outputs_[i], ngraph_outputs_[i]);
         break;
       default:
-        NGRAPH_VLOG(5) << "Could not find the corresponding function for the expected output datatype.";
+        NGRAPH_VLOG(5) << "Could not find the corresponding function for the "
+                          "expected output datatype.";
     }
   }
 }
--- a/test/opexecuter.cpp
+++ b/test/opexecuter.cpp
@@ -144,7 +144,17 @@ void OpExecuter::ExecuteOnTF() {
 void OpExecuter::CompareNGraphAndTF() {
   ASSERT_EQ(tf_outputs_.size(), ngraph_outputs_.size());
   for (int i = 0; i < tf_outputs_.size(); i++) {
-    AssertTensorEquals(tf_outputs_[i], ngraph_outputs_[i]);
+    switch(expected_output_datatypes_[0])
+    {
+      case DT_FLOAT:
+        AssertTensorEqualsFloat(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      case DT_INT32:
+        AssertTensorEqualsInt32(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      default:
+        NGRAPH_VLOG(5) << "Could not find the corresponding function for the expected output datatype.";
+    }
   }
 }
 
--- a/test/test_math_ops.cpp
+++ b/test/test_math_ops.cpp
@@ -389,6 +389,70 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
   opexecuter.RunTest();
 }
 
+TEST(MathOps, Abs) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 2;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+
+  AssignInputValues(A, -3.5f);
+
+  vector<int> static_input_indexes = {};
+  auto R = ops::Abs(root, A);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Abs", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+  
+  opexecuter.RunTest();
+}
+
+// Cast float to int
+TEST(MathOps, Cast1D) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1}));
+
+  AssignInputValues(A, 3.5f);
+
+  vector<int> static_input_indexes = {};
+  auto R = ops::Cast(root, A, DT_INT32);
+
+  vector<DataType> output_datatypes = {DT_INT32};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Cast", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}
+
+// Cast float to int
+TEST(MathOps, Cast2D) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 2;
+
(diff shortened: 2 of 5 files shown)


test/test_utilities.cpp
```diff
@@ -48,6 +48,18 @@ void AssertTensorEquals(Tensor& T1, Tensor& T2) {
...
+  auto T_size = T1.flat<int32>().size();
+  auto T1_data = T1.flat<int32>().data();
+  auto T2_data = T2.flat<int32>().data();
+  for (int k = 0; k < T_size; k++) {
+    auto a = T1_data[k];
+    auto b = T2_data[k];
+    EXPECT_EQ(a, b);
+  }
+}
+
```
sayantan-nervana: We can (somewhat) templatize AssertTensorEqual as follows:
We pass the definition of equality to AssertTensorEqual and  use EXPECT_TRUE, instead of relying on EXPECT_EQUAL or EXPECT_FLOAT_EQ.
```
template <typename T>
void AssertTensorEqual(Tensor& T1, Tensor& T2, std::function<bool(T, T)> eq){
    ASSERT_EQ(T1.shape(), T2.shape());
    auto T_size = T1.flat<T>().size();
   auto T1_data = T1.flat<T>().data();
   auto T2_data = T2.flat<T>().data();
  for (int k = 0; k < T_size; k++) {
    auto a = T1_data[k];
    auto b = T2_data[k];
    EXPECT_TRUE(eq(a, b));
  }
}
```

sayantan-nervana: ```eq``` can have a default value:
```void AssertTensorEqual(Tensor& T1, Tensor& T2, std::function<bool(T, T)> eq = [](T arg0, T arg1){return arg0==arg1})```

This would take care of bool and int. so the switchcase will be much reduced now:
```
switch (data_type_enum){
case dt_int:
case dt_bool:
case others_where_vanilla_equality_suffices:
  AssertTensorEqual(a, b)
  break;
case dt_float:
  AssertTensorEqual(a, b, custom_eq_op)
default:
  fail here
}
```

or, we could have the default case as vanilla equality

File: test/test_math_ops.cpp [hunk 2]
```diff
@@ -389,6 +389,69 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+
+// Cast float to int
+TEST(MathOps, Cast2D) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 2;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+
+  AssignInputValues(A, 3.5f);
```
shresthamalik: We have random initializers which @mingshan-wang  added in utility functions, in case you want to try those. Completely optional.

test/test_utilities.cpp
shresthamalik: Thanks @sayantan-nervan.  @kanvi-nervana  and I had been discussing how to templatize/overload this function. This is a good option. Most of our ops have float outputs.  So , need to figure out how the float comparison would work out with this, we might have to change the function to take in tolerance std::function<bool(T, T, tolerance)> eq

sayantan-nervana: here's another possibility:
we have the AssertTensorEqual templated on T
```
template <typename T>
AssertTensorEqual(Tensor& T1, Tensor& T2){
}
```

Then we have:
```
EXPECT_TRUE(eq<T,T>(a, b))
```

Finally we have a template specialization:
```
template <typename T>
bool eq(T arg0, T arg1){ return arg0==arg1;}

template <>
bool eq(float arg0, float arg1){return arg0 == arg1;}
```

Now this removes the requirement of even the switch case

shresthamalik: Yes, this is even better! 👍

shresthamalik: Another consideration, here would be the error messages. On using EXPECT_EQ()/ EXPECT_FLOAT_EQ() When the test case fails, the message prints both the values we are comparing (a and b) .  EXPECT_TRUE() would print out the boolean value returned by the function eq and compare it with true. We wont see the actual values being compared, I think. So there is a case for using EXPECT_EQ and EXPECT_FLOAT_EQ .

kanvi-nervana: I haven't implemented the latest suggestion but if we want to consider that we could add debug logs.

test/opexecuter.cpp
```diff
@@ -135,13 +135,26 @@ void OpExecuter::ExecuteOnTF() {
...
+    NGRAPH_VLOG(5) << " TF op " << i << tf_outputs_[i].DebugString();
+  }
 }
 
 // Compares tf_outputs_ with ngraph_outputs_
 void OpExecuter::CompareNGraphAndTF() {
   ASSERT_EQ(tf_outputs_.size(), ngraph_outputs_.size());
   for (int i = 0; i < tf_outputs_.size(); i++) {
-    AssertTensorEquals(tf_outputs_[i], ngraph_outputs_[i]);
+    switch (expected_output_datatypes_[0]) {
```
shresthamalik: It should be switch (expected_output_datatypes_[i])

kanvi-nervana: Thanks :)

test/opexecuter.cpp
```diff
@@ -135,13 +135,41 @@ void OpExecuter::ExecuteOnTF() {
...
+                                          } else if(arg0 == 0 && arg1 == 0) {
+                                            return true;
+                                          } else if(arg0 == 0 && arg1 != 0) {
+                                            return false; // because the abs(x/arg1) = 1 > 0.001
+                                          } else {
+                                            return false; // because the abs(x/arg0) = 1 > 0.001
+                                          }
+                                      
+                                        });
+        break;
```
sayantan-nervana: for the last 2 cases (one of the args is 0), we cannot always return false. 

Lets pick something from here:
https://en.wikipedia.org/wiki/Relative_change_and_difference

we want to: 
```
measure = |x-y|/f(x,y)
```

we can choose ```f = max``` or ```f(x,y) = |x+y|/2``` or ```f(x,y) = (|x|+|y|)/2```

I prefer the first one as [suggested](http://c-faq.com/fp/fpequal.html) by Doug Gwyn of ANSI C fame. In this case we have only 2 cases (both zero, everything_else):

```
bool both_are_zero = (x==0 && y==0);
return both_are_zero || ((|x-y|/max(|x|, |y|)) < tol) 
```

test/test_utilities.cpp
```diff
@@ -116,5 +104,14 @@ void ValidateTensorData(Tensor& T1, Tensor& T2, float tol) {
   }
 }
 
+template <>
+bool eq(float arg0, float arg1) {
+  if (arg0 == 0 && arg1 == 0) {
+    return true;
+  } else {
+    return (abs(arg0 - arg1) / max(abs(arg0), abs(arg1)) <= 0.001);
```
kanvi-nervana: Should I add a cout here which would print the values in case of false(failing test) rather than enabling the log levels?

test/opexecuter.cpp
```diff
@@ -135,13 +135,26 @@ void OpExecuter::ExecuteOnTF() {
...
+      case DT_FLOAT:
+        AssertTensorEquals<float>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      case DT_INT32:
+        AssertTensorEquals<int>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      default:
+        NGRAPH_VLOG(5) << "Could not find the corresponding function for the "
+                          "expected output datatype.";
+    }
```
sayantan-nervana: We should have an assert or fail (something like ```EXPECT_TRUE(false)```) it at this point, so that we take heed of the warning.
Otherwise given some other datatype CompareNGraphAndTF might happily exit without comparing, and we would not know (unless log-5 is enabled and someone is peering into it).

test/test_utilities.cpp
shresthamalik: I am not able to see the AssertTensorEquals() function on the git. Another way would be to add a switch(T) inside the AssertTensorEquals() and let GTEST do all the heavy lifting of validating float, int outputs and generating clean failure logs. @sayantan-nervana  @kanvi-nervana what do you think?

template<T>
void AssertTensorEquals(Tensor& T1, Tensor& T2) {
  ASSERT_EQ(T1.shape(), T2.shape());
  auto T_size = T1.flat<T>().size();
  auto T1_data = T1.flat<T>().data();
  auto T2_data = T2.flat<T>().data();
  for (int k = 0; k < T_size; k++) {
    auto a = T1_data[k];
    auto b = T2_data[k];
    switch(T){
       case float:
         EXPECT_FLOAT_EQ(a, b);
         break;
       case int:
         EXPECT_EQ(a, b);
         break;
       case ....
     }
  }
}

shresthamalik: Need to validate if switch works like this on template : switch(T)

shresthamalik: Validated from @sayantan-nervana .. tf datatypes DT_FLOAT, DT_INT32 are defined as enum .. so this might work.

sayantan-nervana: it would be:
```template<int T>```
or with better types:
```template<tensorflow_enum_type T>```
But then, why template it (except maybe to get _slightly_ faster runtimes at the expense of _slightly_ slower compile times. We might as well pass the tensorflow datatype as a param to AssertTensorEquals.

Not sure this line will not work: T1.flat<T>().data();
current usage: ```T1.flat<float>().data()``` . ```flat``` seems to take in real data types, so tensorflow enum datatypes may not work.

Comment (sayantan-nervana) [reply test/test_utilities.cpp]: Following up on discussion from slack:

```

template <DataType T>  //an enum/int
void AssertTensorEquals(Tensor& T1, Tensor& T2) {
  ASSERT_EQ(T1.shape(), T2.shape());
  auto T_size = T1.flat<EnumToDataType<T>::Type>().size();
  auto T1_data = T1.flat<EnumToDataType<T>::Type>().data();
  auto T2_data = T2.flat<EnumToDataType<T>::Type>().data();
  for (int k = 0; k < T_size; k++) {
    auto a = T1_data[k];
    auto b = T2_data[k];
    switch (T){
      case DT_FLOAT:
      case DT_DOUBLE:
        EXPECT_FLOAT_EQ(a, b);
        break;
      case DT_INT32:
      case DT_OTHER_INTS:
        EXPECT_EQ(a, b);
    }
  }
}
```

Comment (shresthamalik) [reply test/test_utilities.cpp]: Thanks .. This is super! 👍 Addresses all our concerns.

File: test/opexecuter.cpp (hunk 6 above)
Comment (avijit-nervana) [reply test/opexecuter.cpp]: Agree with @sayantan-nervana - need an ASSERT here to catch this condition.

File: test/test_utilities.h [hunk 7]
```diff
@@ -32,14 +32,38 @@ namespace ngraph_bridge {
...
+  auto T2_data = T2.flat<T>().data();
+  for (int k = 0; k < T_size; k++) {
+    auto a = T1_data[k];
+    auto b = T2_data[k];
+    bool rt = eq<T>(a, b);
+    if (!rt) {
+      cout << " TF output " << a << endl;
+      cout << " NG output " << b << endl;
+    }
+    EXPECT_TRUE(rt);
```
avijit-nervana: Try this: 
```
EXPECT_TRUE(...) << 
   " TF output " << a << endl
  ...
```

test/opexecuter.cpp
```diff
@@ -135,13 +135,31 @@ void OpExecuter::ExecuteOnTF() {
...
+        AssertTensorEquals<float>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      case DT_INT8:
+      case DT_INT16:
+      case DT_INT32:
+      case DT_INT64:
+        AssertTensorEquals<int>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
+      default:
+        EXPECT_TRUE(false);
```
avijit-nervana: Try this:
```
EXPECT_TRUE(false) << "Could not find the corresponding function for the " ...
```

test/test_utilities.cpp
sayantan-nervana: This is getting into a template hell with difficult-to-parse compilation errors, debugging which would need time/effort. So suggest we not use BKM and switch to second best BKM discussed earlier.

shresthamalik: Thanks for trying! 👍

PR #194, opened 2018-09-10
Remove the rejection mark for is_training=false case in FusedBatchNor…

jianyinglang:
…mGrad to avoid the segfault in resnet50 training run.
The rejection label for `is_training=False` case causes a segment fault for resnet50 training runs because resnet50 training run command does not have an `is_training` flag. For a quick fix, just remove the mark in file  `ngraph_mark_for_clustering.cc` for now. We need to add some more solid implementation for `is_training=False` case in `FusedBatchNormGrad` op implementation (it might not be a priority for now).

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1422,6 +1422,8 @@ static Status TranslateFusedBatchNormGradOp(
   TF_RETURN_IF_ERROR(ValidateInputCount(op, 5));
 
   bool tf_is_training;
+  // Now we only support is_training=true case. We will either mark
+  // or implement is_training=false case later.
   if (GetNodeAttr(op->attrs(), "is_training", &tf_is_training) !=
       Status::OK()) {
     NGRAPH_VLOG(3) << "is_training attribute not present, setting to true";
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1422,8 +1422,8 @@ static Status TranslateFusedBatchNormGradOp(
   TF_RETURN_IF_ERROR(ValidateInputCount(op, 5));
 
   bool tf_is_training;
-  // Now we only support is_training=true case. We will either mark
-  // or implement is_training=false case later.
+  // Now we only support is_training=true case. We marked rejection for the case
+  // is_training=false.
   if (GetNodeAttr(op->attrs(), "is_training", &tf_is_training) !=
       Status::OK()) {
     NGRAPH_VLOG(3) << "is_training attribute not present, setting to true";
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1422,7 +1422,7 @@ static Status TranslateFusedBatchNormGradOp(
   TF_RETURN_IF_ERROR(ValidateInputCount(op, 5));
 
   bool tf_is_training;
-  // Now we only support is_training=true case. We marked rejection for the case
+  // We only support is_training=true case. We marked rejection for the case
   // is_training=false.
   if (GetNodeAttr(op->attrs(), "is_training", &tf_is_training) !=
       Status::OK()) {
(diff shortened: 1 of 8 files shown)


sayantan-nervana:
here is the description of is_training:
is_training: A bool value to indicate the operation is for training (default) or inference.

this line confuses me: operation is for training (default)
does it mean, if not present it is in training mode? (which makes sense since it is a grad op)
so we infer if batchnormgrad is in terianing if the flag is not present or if it is explicitly true?

sayantan-nervana:
but does the absence of the optional flag 'is_training' mean that the node is in training mode?
that is if one attempts to extract the is_training attribute and fails then we infer it is training?
or, do you mean that if is_training is not present we cannot decide that it is training or not?

jianyinglang:
> but does the absence of the optional flag 'is_training' mean that the node is in training mode?
> that is if one attempts to extract the is_training attribute and fails then we infer it is training?
> or, do you mean that if is_training is not present we cannot decide that it is training or not?

I just checked the code and confirmed that `is_training` is true by default. Digging into the code to figure out more.

src/ngraph_mark_for_clustering.cc
```diff
@@ -272,17 +272,8 @@ Status MarkForClustering(Graph* graph) {
...
-        // Reject if "is_training" is set false.
-        bool tf_is_training;
-        TF_RETURN_IF_ERROR(
-            GetNodeAttr(n->attrs(), "is_training", &tf_is_training));
-        if (!tf_is_training) {
-          *result = false;
-          return Status::OK();
-        }
-        SimpleConfirmationFunction()(n, result);
-      };
```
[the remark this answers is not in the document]
jianyinglang: Now I understand why it was causing the segfault. For the `is_training` case, it did not return `true` in confirmation function because I coded `SimpleConfirmationFunction()(n, result);` instead of `return SimpleConfirmationFunction()(n, result);`. Fixed that bug in the updated code.

sayantan-nervana:
Yes, it did warn when compiling 'return value not used'.

jianyinglang:
Thanks @sayantan-nervana ! Just changed the code according to your suggestions.

src/ngraph_mark_for_clustering.cc
```diff
@@ -272,8 +272,12 @@ Status MarkForClustering(Graph* graph) {
       confirmation_functions["FloorDiv"] = SimpleConfirmationFunction();
       confirmation_functions["FloorMod"] = SimpleConfirmationFunction();
       confirmation_functions["FusedBatchNorm"] = SimpleConfirmationFunction();
-      confirmation_functions["FusedBatchNormGrad"] =
-          SimpleConfirmationFunction();
+      confirmation_functions["FusedBatchNormGrad"] = [](Node* n, bool* result) {
+        //bool tf_is_training;
```
avijit-nervana: It's good to not have commented out code (aka `dead code`). Please remove this line.

src/ngraph_builder.cc
```diff
@@ -1422,8 +1422,8 @@ static Status TranslateFusedBatchNormGradOp(
   TF_RETURN_IF_ERROR(ValidateInputCount(op, 5));
 
   bool tf_is_training;
-  // Now we only support is_training=true case. We will either mark
-  // or implement is_training=false case later.
+  // Now we only support is_training=true case. We marked rejection for the case
```
avijit-nervana: Is this comment accurate i.e., is it that `we only support is_training=true case` now or is this permanent?

src/ngraph_mark_for_clustering.cc
jianyinglang: Removed.

src/ngraph_builder.cc
jianyinglang: Removed `Now` in the comment

PR #195, opened 2018-09-11
fixed version checking compatible for python 2 & 3

avijit-nervana:


crlishka:
How to reproduce the issue I am seeing:
1. Build the ngraph wheel using Python 2.  You must use Python 2 for this step.
2. Create and activate a new virtual environment for Python 3.  (You must use Python 3 for this step.)
3. pip-install a tensorflow wheel for Python 3 (in the virtual environment activated in step 2).
4. pip-install the ngraph wheel from step 1 (in the virtual environment activated in step 2).
3. In the virtual environment activated in step 2, start python.  "import tensorflow", followed by "import ngraph".  You should see the version mismatch error, which is really a comparison error between different types of Python strings.

crlishka:
I created an ngraph wheel and tensorflow wheels using the ngraph-tf-daily-build job and Igor's latest commit (163fbe7).  The resulting wheels work in my testing with both python2 and python3 now, which is excellent news.

PR #196, opened 2018-09-12
Mingshan/Support UINT16 and enable the NCF to run with NGRAPH

mingshan-wang:
Added small change to support UINT16 data type on NGRAPH.

Tested with the NCF model, both the training and inference runs with NGRAPH.

### DIFF
--- a/src/ngraph_encapsulate_op.cc
+++ b/src/ngraph_encapsulate_op.cc
@@ -191,6 +191,8 @@ class NGraphEncapsulateOp : public OpKernel {
         TensorDataToStream<uint8>(ostream, n_elements, data);
         break;
       case DT_UINT16:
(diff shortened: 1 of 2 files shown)


src/ngraph_encapsulate_op.cc
```diff
@@ -191,6 +191,8 @@ class NGraphEncapsulateOp : public OpKernel {
         TensorDataToStream<uint8>(ostream, n_elements, data);
         break;
       case DT_UINT16:
+        TensorDataToStream<uint16>(ostream, n_elements, data);
+        break;
```
sayantan-nervana: since DT_UINT16 and DT_QUINT16 seem to map to uint16 only, we do not need these 2 lines. DT_UINT16's case should fall through to DT_QUINT16

PR #197, opened 2018-09-12
Sindhu/Conversion of the below pytests to C++ tests

sindhu-nervana:
1) Math Op test_abs.py
2) Array Op test_fill.py

### DIFF
--- a/test/test_array_ops.cpp
+++ b/test/test_array_ops.cpp
@@ -33,6 +33,11 @@
 #include "tensorflow/cc/client/client_session.h"
 #include "tensorflow/cc/ops/standard_ops.h"
 #include "tensorflow/core/framework/tensor.h"
+#include "tensorflow/core/framework/tensor_testutil.h"
+#include "tensorflow/core/kernels/ops_testutil.h"
+#include "tensorflow/core/kernels/ops_util.h"
+#include "tensorflow/core/lib/core/status_test_util.h"
+#include "tensorflow/core/platform/test.h"
 #include "tensorflow/core/public/session.h"
 
 using namespace std;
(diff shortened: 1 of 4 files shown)


test/test_math_ops.cpp
```diff
@@ -389,6 +389,48 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
   opexecuter.RunTest();
 }
 
+TEST(MathOps, Abs1d) {
```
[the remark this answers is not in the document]
sindhu-nervana: I made the changes.

test/test_math_ops.cpp
```diff
@@ -389,6 +389,47 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+  Scope root = Scope::NewRootScope();
+  int dim1 = 1;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1}));
```
sayantan-nervana: By abs1d, we mean a abs test on a rank1 tensor (or a vector).
If my understanding is right this is how you might define a vector of length 5
```
TensorShape({dim1})
```

The fact that there is only 1 number in the braced list means it is rank 1. however that number itself can be non-one.

In this test you have dim1 set to 1. So '1' becomes a bit overloaded (rank is 1 and dim is also 1). Also we are testing a degenerate/trivial vector since it is basically a single number

TL;DR: set dim1 to some integer > 1

test/test_array_ops.cpp
```diff
@@ -157,6 +157,40 @@ TEST(ArrayOps, Unpack) {
...
+  Tensor input_data(DT_FLOAT, TensorShape({}));
+  AssignInputValuesRandom(input_data);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  // Fill creates a tensor filled with scalar value
+  // 1-D shape of the output tensor
+  auto shape1 =
+      ops::Const(root1, {int32(2), int32(1), int32(3)}, TensorShape({3}));
+  auto R1 = ops::Fill(root1, shape1, input_data);
```
sayantan-nervana: So earlier we ran into the problem of incorrect types and casting. You seem to have figured out the right types to use here ```int32(2)```:

So now you can have a loop over different sizes.
```
vector<vector<i32>> shape_vector= {{2,1,3}, {1,2}, {4,5}};
...
for (auto shape : shape_vector)
{
  auto shape_const = ops::Const(root1, shape, TensorShape({shape.size()});
  ...
}
```

test/test_math_ops.cpp
sindhu-nervana: Changed.

test/test_array_ops.cpp
sayantan-nervana: As discussed, you are right, we cannot use vectors. Please checkout slack for a workaround

sindhu-nervana: I did try to use the input_sizes vector and loop for different sizes but that did not work as Fill op is not taking pre-defined vectors. So used constant and assigned int32 values to the tensor.

test/test_array_ops.cpp
```diff
@@ -157,6 +157,40 @@ TEST(ArrayOps, Unpack) {
...
+
+  vector<int> static_input_indexes = {0};  // has static input
+
+  for (auto const& input_size : input_sizes) {
+    Scope root = Scope::NewRootScope();
+
+    int input_dim = input_size.size();
+    Tensor shape(DT_INT32, TensorShape({input_dim}));
+    AssignInputIntValues(shape, 5);
+
```
sayantan-nervana: I think this is not doing what we want.
Lets say: ```input_size = {2,3,4}```
Then, ```input_dim = 3```
Then ```shape``` is a int32 tensor of dims ```1x3``` (or just 3, not exactly sure)
Then ```AssignInputIntValues(shape, 5);``` is saying fill up this ```1x3``` tensor names ```shape``` with values starting from 0 and increasing by int steps. So ```shape``` tensor would now be ```{0,1,2}```.

Which means we will create a tensor of size ```0,1,2``` and not ```{2,3,4}``` as originally intended.

Can you please check and confirm this.

sayantan-nervana:
Maybe we add a new utility? (unless it already exists?)
```
template <typeclass T>
void AssignInputValuesFromVector(Tensor& A, vector<T> x) {
  auto A_flat = A.flat<float>();
  auto A_flat_data = A_flat.data();
  assert A_flat.size()==x.size()
  for (int i = 0; i < A_flat.size(); i++) {
    A_flat_data[i] = x[i];
  }
}
```

test/test_utilities.h
```diff
@@ -41,6 +41,16 @@ void AssignInputValuesRandom(Tensor& A);
...
+void AssignInputValuesFromVector(Tensor& A, vector<T> x) {
+  auto A_flat = A.flat<T>();
+  auto A_flat_data = A_flat.data();
+  assert(A_flat.size() == x.size());
+  for (int i = 0; i < A_flat.size(); i++) {
+    A_flat_data[i] = x[i];
+  }
+}
+
 template <class T>
```
sayantan-nervana: Could you please remive the function declarations, if they are not being used anywhere
AssertTensorEqualsFloat
AssertTensorEqualsIntInt32

PR #199, opened 2018-09-12
Sindhu/more pytest to c++

sindhu-nervana:


test/test_math_ops.cpp
```diff
@@ -485,6 +485,81 @@ TEST(MathOps, Abs2d) {
...
+  AssignInputValues(A, -2.5f);
+  AssignInputValues(B, 4.0f);
+  vector<int> static_input_indexes = {};
+  auto R = ops::Pow(root, A, B);
+  vector<DataType> output_datatypes = {DT_FLOAT};
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Pow", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+  opexecuter.RunTest();
+}
```
sayantan-nervana: For both pow and log, 1D and 2D tests could be folded into one test in a loop (like the one you did earlier for fill).

PR #200, opened 2018-09-12
Kanvi/Implement ShapeOp

kanvi-nervana:
The test currently fails because the framework needs the output datatype to be float where as the output datatype for this op is int. The fix for that is in the other branch.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -2031,14 +2031,15 @@ static Status TranslateShapeOp(
   // the rank of the input tensor which will be the shape to the Constant Op
   auto rank = input_shape.size();
 
-  // the inputs to the Constant Op
-  // type
-  ng::element::Type type;
-  TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(DataType::DT_INT32, &type));
+  DataType dtype;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "out_type", &dtype));
 
-  // shape
+  // the inputs to the Constant Op
+  ng::element::Type type;
+  TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(dtype, &type));
+
   auto shape = ng::Shape(1, rank);
-  // values
+
   std::vector<int> values(rank);
   for (int i = 0; i < rank; i++) {
     values[i] = input_shape[i];
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1422,6 +1422,8 @@ static Status TranslateFusedBatchNormGradOp(
   TF_RETURN_IF_ERROR(ValidateInputCount(op, 5));
 
   bool tf_is_training;
+  // We only support is_training=true case. We marked rejection for the case
+  // is_training=false.
   if (GetNodeAttr(op->attrs(), "is_training", &tf_is_training) !=
       Status::OK()) {
     NGRAPH_VLOG(3) << "is_training attribute not present, setting to true";
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -2030,6 +2030,34 @@ static Status TranslateRsqrtOp(
       });
 }
 
+static Status TranslateShapeOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+  
+  // the shape of the input tensor which will be the value to the Constant Op
+  auto input_shape = ng_input->get_shape();
+  
+  // the rank of the input tensor which will be the shape to the Constant Op
+  auto rank = input_shape.size();
+
+  // the inputs to the Constant Op
+  // type
+  auto type = Builder::TF_NGRAPH_TYPE_MAP().at(DataType::DT_FLOAT);
+
+  // shape
+  auto shape = ng::Shape(1,rank);
+  // values
+  std::vector<int> values(rank);
+  for(int i = 0; i < rank; i++) {
+    values[i] = input_shape[i];
+  }
+  SaveNgOp(ng_op_map, op->name(),
+           make_shared<ng::op::Constant>(type, shape, values));
+  return Status::OK();
+}
+
 static Status TranslateSigmoidOp(
     const Node* op, const std::vector<const Tensor*>& static_input_map,
     Builder::OpMap& ng_op_map) {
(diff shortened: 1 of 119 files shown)


src/ngraph_builder.cc
```diff
@@ -2030,6 +2030,34 @@ static Status TranslateRsqrtOp(
...
+  
+  // the shape of the input tensor which will be the value to the Constant Op
+  auto input_shape = ng_input->get_shape();
+  
+  // the rank of the input tensor which will be the shape to the Constant Op
+  auto rank = input_shape.size();
+
+  // the inputs to the Constant Op
+  // type
+  auto type = Builder::TF_NGRAPH_TYPE_MAP().at(DataType::DT_FLOAT);
```
sayantan-nervana: the // type could go in the above line, and format-check probably will still be happy

sayantan-nervana: DT_FLOAT -> DT_INT

sayantan-nervana: Oh, its fixed :-D

src/ngraph_builder.cc
```diff
@@ -2019,6 +2019,35 @@ static Status TranslateRsqrtOp(
...
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+
+  // the shape of the input tensor which will be the value to the Constant Op
+  auto input_shape = ng_input->get_shape();
+
+  // the rank of the input tensor which will be the shape to the Constant Op
+  auto rank = input_shape.size();
+
+  // the inputs to the Constant Op
+  // type
```
sayantan-nervana: could you try shifting the // type to the line above

src/ngraph_builder.cc
```diff
@@ -2019,6 +2019,35 @@ static Status TranslateRsqrtOp(
...
+  ng::element::Type type;
+  TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(DataType::DT_INT32, &type));
+
+  // shape
+  auto shape = ng::Shape(1, rank);
+  // values
+  std::vector<int> values(rank);
```
sayantan-nervana: //shape and //values comment is probably not needed, as the variables themselves are names shapes and values

src/ngraph_builder.cc
```diff
@@ -2019,6 +2019,35 @@ static Status TranslateRsqrtOp(
...
+  for (int i = 0; i < rank; i++) {
+    values[i] = input_shape[i];
+  }
+  SaveNgOp(ng_op_map, op->name(),
+           make_shared<ng::op::Constant>(type, shape, values));
+  return Status::OK();
```
sayantan-nervana: instead of type, lets use ```ng::element::i32``` directly. Then we do not need the lines above:
```
ng::element::Type type;
TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(DataType::DT_INT32, &type));
```

test/test_array_ops.cpp
```diff
@@ -81,6 +81,50 @@ TEST(ArrayOps, PreventGradient) {
...
+  vector<int> static_input_indexes = {};
+  auto R = ops::Shape(root, A);
+
+  vector<DataType> output_datatypes = {DT_INT32};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Shape", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
```
sayantan-nervana: Shape2D and Shape3D functions could have been fused into one single function looping over a vector of possible shapes (like Sindhu's fill).

src/ngraph_mark_for_clustering.cc
```diff
@@ -213,6 +213,7 @@ Status MarkForClustering(Graph* graph) {
       type_constraint_map["Reshape"]["T"] = NGraphDTypes();
       type_constraint_map["Reshape"]["Tshape"] = NGraphIndexDTypes();
       type_constraint_map["Rsqrt"]["T"] = NGraphDTypes();
+      type_constraint_map["Shape"]["T"] = NGraphNumericDTypes();
```
aprocter: `T` is the attribute for the input type, so any nGraph-able datatype (`NGraphDTypes()`) should be fine.

In addition to this, `Shape` also has a type variable called `out_type`, so we also will need something like:

```
type_constraint_map["Shape"]["out_type"] = NGraphIndexDTypes();
```

(I think `NGraphIndexDTypes` includes int32 and int64, but I'm not sure.)

src/ngraph_builder.cc
```diff
@@ -2019,6 +2019,35 @@ static Status TranslateRsqrtOp(
...
+  // the shape of the input tensor which will be the value to the Constant Op
+  auto input_shape = ng_input->get_shape();
+
+  // the rank of the input tensor which will be the shape to the Constant Op
+  auto rank = input_shape.size();
+
+  // the inputs to the Constant Op
+  // type
+  ng::element::Type type;
+  TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(DataType::DT_INT32, &type));
```
aprocter: This may need to be `DT_INT32` or `DT_INT64`, depending on the value of the `out_type` attribute.

test/test_array_ops.cpp
kanvi-nervana: I like to keep them separate since that gives more readability. For this test though, I'll only have the 3D - forgot to get rid of the 2D.

src/ngraph_mark_for_clustering.cc
kanvi-nervana: Can the shape op take bool type? I wasn't sure so I used NGraphNumericDTypes.

And yes, NgraphIndexTypes includes int32 and int64.

test/test_array_ops.cpp
kanvi-nervana: Keeping both 2D and 3D since 2D tests for Int64 and 3D for Int32

test/opexecuter.cpp
```diff
@@ -146,15 +146,20 @@ void OpExecuter::CompareNGraphAndTF() {
...
+        AssertTensorEquals<int16>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
       case DT_INT32:
-      case DT_INT64:
         AssertTensorEquals<int>(tf_outputs_[i], ngraph_outputs_[i]);
         break;
+      case DT_INT64:
+        AssertTensorEquals<int64>(tf_outputs_[i], ngraph_outputs_[i]);
+        break;
       default:
```
sayantan-nervana: The only way to avoid the switch-case, is to know expected_output_datatypes_[i] at compile time
(in which case we would do something like: 
```AssertTensorEquals<EnumToDataType<expected_output_datatypes_[i]>::Type>```

but to know expected_output_datatypes_[i] at runtime, OpExecutor must accept it as template arguments, and not as normal arguments. And since expected_output_datatypes_ is of variable length, it would have to be variadic template inputs to Opexecutor constructor. And variadic templated classes might seriously hurt readability.

Therefore this is currently the best way to do it.

I suggest doing this in each case X:
```EnumToDataType<X>::Type```
to remove the possibility of any error or to future-proof against any type.

test/test_array_ops.cpp
sayantan-nervana: looks good here

src/ngraph_mark_for_clustering.cc
aprocter: Yup, it can take any type as input.

PR #201, opened 2018-09-13
Sarkars/pyformat

sayantan-nervana:
Adding only changes in ./maint folder for now. When review is done will add the python format noise in this or another PR.


sayantan-nervana:
CI failed because it was unable to find yapf.
Also which python does CI use? Py2 or Py3. Currently I assert Py3 when running yapf to autoformat.

PR #204, opened 2018-09-13
Enabling the ability to build nbench at command line

ransford2011:
For backend releases, it is convenient to have the nbench tool to be built as well. As matter of convenience, I would like to add this line as a command line option.


aprocter:
Going ahead with the `approve` since I think @avijit-nervana has already caught the only issue I can think of.

ransford2011:
Thanks @avijit-nervana and @aprocter for the feedback.  I will try to close this one by tomorrow.  Had a few things I needed to get in place. 👍

ransford2011:
@avijit-nervana .  I've added the logic for the `install` path.  In regards to setting the `Rpath` logic for nBench, it's not set in nGraph unless is doing so: https://github.com/NervanaSystems/ngraph/blob/edd3aecaca9ce1ddd7f34fc976a36195c8322b63/src/tools/nbench/CMakeLists.txt#L38

PR #206, opened 2018-09-14
Mingshan/Added model_test scripts in diagnositics folder

mingshan-wang:
Follow @sayantan-nervana 's idea, combine the model level testing with the tensorboard utilities in the diagonistics folder. Then we can encourage users to check out their network with that all the utilities in the diagnostics folder and report discrepancies to us. Also maybe can help users debug their custom models.


diagnostics/model_test/verify_model.py
```diff
@@ -0,0 +1,161 @@
...
+    Returns:
+        The output vector obtained from running the input_example through the graph.
+    """
+    frozen_graph_filename = param_dict["frozen_graph_location"]
+    input_tensor_name = param_dict["input_tensor_name"]
+    output_tensor_name = param_dict["output_tensor_name"]
+
+    with tf.gfile.GFile(frozen_graph_filename, "rb") as f:
+        graph_def = tf.GraphDef()
+        graph_def.ParseFromString(f.read())
```
sayantan-nervana: This currently only loads .pb files and not .pbtxt.

You can use ```load_file(graph_file, input_binary, modifier_function_list=[])``` function in ngtf_graph_viewer.py.
for input_binary u can pass: frozen_graph_filename.split('.')[1]=='pb'

diagnostics/model_test/verify_model.py
```diff
@@ -0,0 +1,161 @@
...
+
+    with tf.Graph().as_default() as graph:
+        with tf.device('/job:localhost/replica:0/task:0/device:' + select_device + ':0'):
+            tf.import_graph_def(graph_def)
+
```
sayantan-nervana: tf.device setting is not required now. In C++ we use ActivateNgraph, DeactivateNgraph. In python need to figure it out?

diagnostics/model_test/verify_model.py
```diff
@@ -0,0 +1,161 @@
...
+    ngraph_output_flatten = ngraph_output_squeezed.flatten()
+    tf_output_flatten = tf_output_squeezed.flatten()
+
+    factor = np.prod(ngraph_output_squeezed.shape)
+
+    if desired_norm == 'l1_norm':
+        return np.sum(np.abs(ngraph_output_flatten - tf_output_flatten), axis=0) / factor
+    elif desired_norm == 'l2_norm':
+        return np.sum(np.dot(np.abs(ngraph_output_flatten - tf_output_flatten), np.abs(ngraph_output_flatten - tf_output_flatten)))/factor
+    elif desired_norm == 'inf_norm':
```
sayantan-nervana: you could just use np.linalg.norm for the other norms too.

in that case desired_norm can be np.inf, or 1 or 2. In which case we just call it like below for all cases:
```
if desired_norm not in [1,2,np.inf]:
    Exception("Only 1, 2 and inf norms supported")
np.linalg.norm((ngraph_output_flatten - tf_output_flatten), desired_norm)```

File: diagnostics/model_test/mnist_cnn.json [hunk 4]
```diff
@@ -0,0 +1,10 @@
+{
+    "model_name": "mnist_cnn",
+    "frozen_graph_location": "/nfs/site/home/mingshan/frozen_graphs/frozen_mnist.pb",
+    "input_tensor_name": "import/Placeholder:0",
+    "output_tensor_name":"import/fc2/yconv:0",
+    "l1_norm_threshold": 0.01,
+    "l2_norm_threshold": 0.01,
+    "inf_norm_threshold": 0.01,
+    "input_dimension" : 784
+}
```
sayantan-nervana: The graph can have multiple inputs, and each input can have multiple dimensions.
Ideally we want to support something like this:
[[28,28], [5,5], [3,3,10]]
to indicate 3 input dimensions

diagnostics/model_test/verify_model.py
mingshan-wang: Oh yes, we need to discuss with @avijit-nervana , since I remember previously Chris is doing some python functions to deactivate ngraph in python files. We can check the status of this.

diagnostics/model_test/verify_model.py
mingshan-wang: The script now only supports loading frozen graph .pb file, I think .pbtxt will only contain graph structure but won't contain weights information right, unless you also supply a checkpoint file.

I am plan also to support .pbtxt + checkpoint combination, what do you think?

avijit-nervana: General question - what topologies we tested this on?

sayantan-nervana: frozen graphs come in 2 varieties. Both are 'executable':
.pb are binary files and not very human readable
.pbtxt are human readable. .pbtxts can contain weights and constants though.

checkout the 'text or binary' section here
https://www.tensorflow.org/extend/tool_developers/#text_or_binary

However there may also be stripped .pbtxts, which do not have weights etc. Those we wouldn't be able to process I think.

mingshan-wang:
TESTNOW

diagnostics/model_test/mnist_cnn.json
```diff
@@ -0,0 +1,10 @@
+{
+    "model_name": "mnist_cnn",
+    "frozen_graph_location": "/nfs/site/home/mingshan/frozen_graphs/frozen_mnist.pb",
+    "input_tensor_name": "import/Placeholder:0",
+    "output_tensor_name":"import/fc2/yconv:0",
+    "l1_norm_threshold": 0.01,
+    "l2_norm_threshold": 0.01,
+    "inf_norm_threshold": 0.01,
+    "input_dimension" : [784]
+}
```
sayantan-nervana: Networks could have >1 inputs. So input dims might need to be specified as list of list. same with input_tensor_name. In this case, the input dim should be [[1,784]]. note if user should specify batch size here, instead of us adding [1] as we currently have in this line:
```random_input = np.random.random_sample(tuple([1] + input_dimension))```
Also we can assert length of this list equals length of list of input_tensor_names.

there might also be multiple outputs, but for now we can assume 1 out.

**Comment by sayantan-nervana** [comment]
I still see that ci tests are running.
Need to check if it is because of the new python formatter?

**Comment by avijit-nervana** [comment]
TESTNOW

File: diagnostics/model_test/mnist_cnn.json (hunk 5 above)
Comment (mingshan-wang) [reply diagnostics/model_test/mnist_cnn.json]: Hi @sayantan-nervana , I just updated the script to support multiple inputs, and also added one batch_size field, so for input_dimension, user should specify [[784]] in this case, and batch_size =1. 

Also, FYI, I didn't format the code this time as well, maybe you can use this to check whether the formatter is working or not.

**Comment by mingshan-wang** [comment]
I updated both the diagnostics/README and model_test/README.

For the diagnostics/README, I include how to disable and enable ngraph in python scripts.
For the model_test/README, I include how to use the model_test scripts. 

Feel free to comment if you see anything needs to be improved. Thanks!

**Comment by sayantan-nervana** [comment]
funfact: @mingshan-wang 's python file was perfectly formatted without applying auto format XD. Hence ci passes happily.

**Comment by sayantan-nervana** [comment]
TESTNOW

**Comment by sayantan-nervana** [comment]
some ci changes were just added, so triggering a test

**Comment by sayantan-nervana** [comment]
TESTNOW

File: diagnostics/model_test/verify_model.py [hunk 6]
```diff
@@ -0,0 +1,179 @@
...
+    Calculate the L1/L2/inf norm between the NGRAPH and tensorflow output vectors.
+
+    Args:
+        ngraph_output: The output vector generated from NGRAPH graph.
+        tf_output: The output vector generated from tensorflow graph.
+        desired_norm: L1/L2/inf norm. 
+
+    Returns:
+        Calcualted norm between the vectors.
+
```
sayantan-nervana: Calcualted -> calculated

mingshan-wang:
TESTNOW

diagnostics/model_test/README.md
```diff
@@ -0,0 +1,21 @@
...
+## To prepare the required json file:"
+* You will need the names of the input/output tensors of the model. Currently we are supporting
+multiple input tensors and one output tensor. Put the input tensor names as a list in the ```input_tensor_name``` field of the json file, and the output tensor name as a string in the ```output_tensor_name``` field of the json file
+* You will need the input dimensions for all the input tensors provided. Put the dimensions information as a list in the ```input_dimension``` field of the json file, and the corresponding order of ```input_tensor_name``` list should match the ```input_dimension``` list. Therfore, the length of ```input_tensor_name``` list should match the length of ```input_dimension``` list
+* Specify the the location of the frozen graph in the ```frozen_graph_location``` of the json file
+* Specify the ```batch_size``` field in the json file to the desired batch size for inference
+* Specify the tolerance between the TF and NGraph outputs at ```l1_norm_threshold```, ```l2_norm_threshold``` and ```inf_norm_threshold``` in the json file 
+
+# To run the model test tool:
+	python verify_model.py --json_file="./mnist_cnn.json"
```
kanvi-nervana: May you can change this to python verify_model.py --json_file="<path to json file>"

diagnostics/model_test/verify_model.py
```diff
@@ -0,0 +1,185 @@
...
+        parsed_json = json.load(f)
+        frozen_graph_location = parsed_json['frozen_graph_location']
+        input_tensor_name = parsed_json['input_tensor_name']
+        output_tensor_name = parsed_json['output_tensor_name']
+        l1_norm_threshold = parsed_json['l1_norm_threshold']
+        l2_norm_threshold = parsed_json['l2_norm_threshold']
+        inf_norm_threshold = parsed_json['inf_norm_threshold']
+        input_dimension = parsed_json['input_dimension']
+        batch_size = parsed_json['batch_size']
+
```
kanvi-nervana: I don't think you need the variables above, you can directly parse the json and assign to the dictionary.

diagnostics/model_test/README.md
sayantan-nervana: I feel we should not have a default for ```--json_file``` option, as it is _the_ key input to the code.
In any case, the current mnist.json will not work for an external user (since the .pb in the son is internal).

Therefore without a default, we would force users to supply their own .json so that they do not use our .json by mistake and find that it fails.

diagnostics/model_test/mnist_cnn.json
```diff
@@ -0,0 +1,11 @@
+{
+    "model_name": "mnist_cnn",
+    "frozen_graph_location": "/nfs/site/home/mingshan/frozen_graphs/frozen_mnist.pb",
+    "input_tensor_name": ["import/Placeholder:0"],
```
sayantan-nervana: the .pb is in an internal website. This is fine I suppose, since people are supposed to use their own .pb for debugging. But if we wanted someone to try this script, it would not run (unless we checkin the mnist .pb in our repo).

sayantan-nervana: No this is fine. In fact when we use this framework for ci, we could checkin the sons for other networks like resnet. They may not be runnable by external folks, but could still serve as an example.

diagnostics/README.md
```diff
@@ -28,6 +28,14 @@ To get a **full** dump use the following set of flags
 * Modify the graphdef and dump TB file in ```./vis``` using encapsulate information in ```nodemap.pkl```: ```python ngtf_graph_viewer.py -c nodemap.pkl ./path/to/original_network_pbtxtfile.pbtxt ./vis```. If you do not have the pbtxt of the original tensorflow graph, you can dump it from your script using [write_graph](https://www.tensorflow.org/api_docs/python/tf/train/write_graph)
 *  View the original network with encapsulate information by running tensorboard, using the files created in ```./vis```.
 
+### Disable NGRAPH in python
+* In your script, import ngraph by having: ```import ngraph```
```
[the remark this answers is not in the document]
mingshan-wang: I will add this change to this PR. Thanks for pointing this out!

mingshan-wang:
Updates:

1. Provide a template.json file for user to start with. They can modify the template.json directly to match their models.
2. Make a check in the script to require the user to supply a json file.
3. Update the diagnostic README to warn user disable ngraph after session run has no effect.
4. Minor change in the script and model_test README.

Thanks for all the comments!

diagnostics/README.md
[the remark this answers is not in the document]
sayantan-nervana: And while we are adding stuff to READMEs could you please also add something that says "use NCHW for good performance (dont use NHWC)" in some README.

@avijit-nervana : Should we add this? Where should we add this?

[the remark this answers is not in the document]
avijit-nervana: @sayantan-nervana This information should be in the examples page

diagnostics/model_test/verify_model.py
```diff
@@ -106,35 +106,30 @@ def parse_json():
...
+        param_dict["frozen_graph_location"] = parsed_json[
+            'frozen_graph_location']
+        param_dict["input_tensor_name"] = parsed_json['input_tensor_name']
+        param_dict["output_tensor_name"] = parsed_json['output_tensor_name']
+        param_dict["l1_norm_threshold"] = parsed_json['l1_norm_threshold']
+        param_dict["l2_norm_threshold"] = parsed_json['l2_norm_threshold']
+        param_dict["inf_norm_threshold"] = parsed_json['inf_norm_threshold']
+        param_dict["input_dimension"] = parsed_json['input_dimension']
+        param_dict["batch_size"] = parsed_json['batch_size']
```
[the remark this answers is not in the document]
mingshan-wang: Oh yes, you are right... sorry for this overengineering

PR #208, opened 2018-09-15
Shrestha/ Updated the tf_graphcycles files 

shresthamalik:
Updated the tf_graphcycles files with Revision: a768f270c15ded657c30fe9ef873251de3556e58

This is second to last revision for these files. The latest revision uses absl library
Added headers with github reference

### DIFF
--- a/src/tf_graphcycles.h
+++ b/src/tf_graphcycles.h
@@ -18,8 +18,8 @@ File: tensorflow/tensorflow/compiler/jit/graphcycles/graphcycles.h
 
 *******************************************************************************/
 
-#ifndef TENSORFLOW_COMPILER_JIT_GRAPHCYCLES_GRAPHCYCLES_H_
-#define TENSORFLOW_COMPILER_JIT_GRAPHCYCLES_GRAPHCYCLES_H_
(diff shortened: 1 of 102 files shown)


src/tf_graphcycles.h
```diff
@@ -1,20 +1,25 @@
...
-#ifndef NGRAPH_TF_GRAPHCYCLES_H_
-#define NGRAPH_TF_GRAPHCYCLES_H_
+This file is a copy of
+Github repository: https://github.com/tensorflow/tensorflow
+Revision: a768f270c15ded657c30fe9ef873251de3556e58
+File: tensorflow/tensorflow/compiler/jit/graphcycles/graphcycles.h
+
+*******************************************************************************/
+
+#ifndef TENSORFLOW_COMPILER_JIT_GRAPHCYCLES_GRAPHCYCLES_H_
```
shresthamalik: Do we want to use name "NGRAPH_TF_GRAPHCYCLES_H_" like before or stick to the name in the original file?

avijit-nervana: It's good to change the macro as otherwise in future there may be some inclusion conflicts. Also we should "ngraph_" prefix to the filename due to the same reason.

PR #210, opened 2018-09-17
Kanvi/Add capability to get serialized graphs and other debug graphs

kanvi-nervana:
Add capability to get serialized graphs and other debug graphs while using unit test framework for debugging.

### DIFF
--- a/src/ngraph_encapsulate_op.cc
+++ b/src/ngraph_encapsulate_op.cc
@@ -276,8 +276,15 @@ class NGraphEncapsulateOp : public OpKernel {
         NGRAPH_VLOG(0) << "Serializing graph to: " << file_name;
         std::string js = ngraph::serialize(ng_function, 4);
         {
-          std::ofstream f(file_name);
-          f << js;
+          try{
+              std::ofstream f(file_name);
+              f << js;
+          }
+          catch(std::ofstream::failure e)
+          {
+            std::cerr << "Exception opening/closing file\n";
+          }
+
         }
       }
 
--- a/src/ngraph_encapsulate_op.cc
+++ b/src/ngraph_encapsulate_op.cc
@@ -329,38 +329,50 @@ class NGraphEncapsulateOp : public OpKernel {
       void* current_src_ptr = (void*)DMAHelper::base(&ctx->input(i));
       std::shared_ptr<ng::runtime::TensorView> current_tv;
 
-      if (s_ng_backend_name == "CPU") {
-        // We need to check last_tv != nullptr, since there are cases where at
-        // the first call to the ng_function, both the current_src_ptr (when the
-        // input is a 0-sized tensor) and last_src_ptr (uninitialized at the
-        // first call) are nullptr
-        if (current_src_ptr == last_src_ptr && last_tv != nullptr) {
-          // Mark each tensor as non-stale if:
-          //   1. the freshness tracker says the tensor has not changed since
-          //      the last time ng_function was called, and
-          //   2. we are using the same tensor in this argument position as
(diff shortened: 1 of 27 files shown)


test/opexecuter.cpp
```diff
@@ -312,6 +317,18 @@ void OpExecuter::ExecuteOnNGraph() {
   // ng function should get same number of outputs
   ASSERT_EQ(expected_output_datatypes_.size(), ng_function->get_output_size());
 
+  // For debug
+  // Serialize to nGraph if needed
+  if (std::getenv("NGRAPH_ENABLE_SERIALIZE") != nullptr) {
+    std::string file_name = "unit_test" + test_op_type_ + ".json";
+    NGRAPH_VLOG(0) << "Serializing graph to: " << file_name;
```
sayantan-nervana: currently the file name string is:
```unit_testAdd.json```
we could add an underscore and make it:
```unit_test_Add.json```

File: test/opexecuter.cpp [hunk 2]
```diff
@@ -312,6 +317,18 @@ void OpExecuter::ExecuteOnNGraph() {
   // ng function should get same number of outputs
   ASSERT_EQ(expected_output_datatypes_.size(), ng_function->get_output_size());
 
+  // For debug
+  // Serialize to nGraph if needed
+  if (std::getenv("NGRAPH_ENABLE_SERIALIZE") != nullptr) {
+    std::string file_name = "unit_test_" + test_op_type_ + ".json";
+    NGRAPH_VLOG(0) << "Serializing graph to: " << file_name;
+    std::string js = ngraph::serialize(ng_function, 4);
+    {
```
avijit-nervana: Why do you need this extra scope? Also - you may consider a try catch here i.e., to catch potential exceptions thrown by the ofstream <<. It's a good practice to catch exceptions especially during I/O.

sayantan-nervana: In which case, lets make similar changes here too:
https://github.com/NervanaSystems/ngraph-tf/blob/55f9f7ee133a94e05f41fedc4518b81a4c50563f/src/ngraph_encapsulate_op.cc#L277

avijit-nervana: Agree with @sayantan-nervana - please add this change in this PR.

src/ngraph_encapsulate_op.cc
```diff
@@ -275,9 +275,14 @@ class NGraphEncapsulateOp : public OpKernel {
...
         std::string js = ngraph::serialize(ng_function, 4);
-        {
-          std::ofstream f(file_name);
+        std::ofstream f;
+        f.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+        try {
+          f.open(file_name);
           f << js;
+          f.close();
+        } catch (std::ofstream::failure e) {
```
avijit-nervana: 1. May consider adding `&` 
2. Add e.what() in the error message display.

test/opexecuter.cpp
```diff
@@ -312,6 +317,23 @@ void OpExecuter::ExecuteOnNGraph() {
...
+    std::string file_name = "unit_test_" + test_op_type_ + ".json";
+    NGRAPH_VLOG(0) << "Serializing graph to: " << file_name << endl;
+    std::string js = ngraph::serialize(ng_function, 4);
+    std::ofstream f;
+    f.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+    try {
+      f.open(file_name);
+      f << js;
+      f.close();
+    } catch (std::ofstream::failure e) {
```
avijit-nervana: Same as above.

PR #211, opened 2018-09-17
sarkars/Continue cpp formatting even if pyformat requirements are not met

sayantan-nervana:
1) Earlier, when applying ./apply-code-format.sh, if pyformat's criteria were not met (Python 3), then the script would break and not format cpp. Fixed that, now cpp format proceeds even if pyformat fails. So, if one has only cpp changes, this will not need one to switch virtual env to py3 to apply formatting.

2) Python version was not displaying correctly before. Fixed that

Thanks @shresthamalik for pointing these out.

Please don't merge yet. Will test it out with a badly formatted python file and make sure it fails ci checks.


sayantan-nervana:
TESTNOW

aprocter:
If you need a `badly formatted python file`, let me know---pretty good at making those

sayantan-nervana:
It seems ci checks happen in Py2 environment.

https://aipg-rancher.intel.com/jenkins/algo/job/ngraph-tf-unittest/592/console
and search for ```About to check formatting```

If Py2 is what we do ci on, I'll change the pyformat version requirement to Py2

@avijit-nervana : Any thoughts/comments?

avijit-nervana:
@sayantan-nervana I think they use Py2. Try changing to py2 based one and see if that fails the test.

sayantan-nervana:
This change is ready @avijit-nervana. Tested with bad-format py file and it fails ci. Changed to Py2, since that is what is used in ci

PR #213, opened 2018-09-18
Fix Python3 daily-build, plus balance Dockerfile additions

crlishka:
ngraph-tf PR#211 made additions to the Dockerfile used for Python 2 builds in CI.  A small bug in this PR caused the Python 3 builds (in ngraph-tf-daily-build) to break when the Python package "yapf" could not be found.

This PR does two specific actions:

1. Fix the bug, as per @sayantan-nervana 's suggestion

1. Balance the Dockerfiles for both Py2 and Py3, by having both "yapf" and "futures" added to both.  In addition, include a comment in the Dockerfiles explaining how these additions are used.

Successful testing of this patch:
ngraph-tf-daily-build: https://aipg-rancher.intel.com/jenkins/algo/job/ngraph-tf-daily-build/263/

### DIFF
--- a/test/ci/docker/Dockerfile.ngraph-tf-ci-py3
+++ b/test/ci/docker/Dockerfile.ngraph-tf-ci-py3
@@ -54,11 +54,14 @@ RUN updatedb
 # keras_applications is needed for modern TensorFlow builds
 # matplotlib are needed to run inference models
 # opencv-python is used by some inference models (for import cv2)
+# yapf and futures are needed for code-format checks (ngraph-tf PR#211)
 RUN pip3 install --upgrade pip
 RUN pip3 install six enum34 mock
 RUN pip3 install scipy portpicker sklearn
 RUN pip3 install keras_applications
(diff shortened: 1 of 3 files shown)


test/ci/docker/Dockerfile.ngraph-tf-ci-py3
```diff
@@ -54,11 +54,14 @@ RUN updatedb
...
 # matplotlib are needed to run inference models
 # opencv-python is used by some inference models (for import cv2)
+# yapf and futures are needed for code-format checks (ngraph-tf PR#211)
 RUN pip3 install --upgrade pip
 RUN pip3 install six enum34 mock
 RUN pip3 install scipy portpicker sklearn
 RUN pip3 install keras_applications
 RUN pip3 install matplotlib opencv-python
+RUN pip3 install yapf
+RUN pip3 install futures
```
sayantan-nervana: not needed for python 3, but lets keep them in in case we switch to py3 for yap formatting in the future

PR #214, opened 2018-09-18
Kanvi/more c++ tests

kanvi-nervana:


kanvi-nervana:
TESTNOW

test/python/test_abs.py
```diff
@@ -1,52 +0,0 @@
...
-    def test_abs_2d(self):
-        test_input = ((1.5, -2.5, 0.0, -3.5), (-4.5, -5.5, 6.5, 1.0))
-        val = tf.placeholder(tf.float32, shape=(2, 4))
-        out = tf.abs(val)
-
-        def run_test(sess):
-            return sess.run(out, feed_dict={val: test_input})
-
-        assert (
-            self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
```
sayantan-nervana: abs, cast and conv2dbackpropinput are the only pytests that would currently run in the deviceless setting. Let us not remove these tests, as it is an example of how to write python tests. Today I had jianying asking me how might one write a pytest in deviceless setting and I pointed her to this. We might need to implement a few ops as pytest (as cpp api might not support it). So lets keep the abs test as an example.

note, the conv2dbackpropinput test exists and uses the ```self.with_ngraph(run_test_ngraph), self.without_ngraph(run_test_tf)``` line, but it is a complex testcase. For a hello-world, this is a good example.

kanvi-nervana: ok I'll keep the above mentioned 3 tests

test/test_math_ops.cpp
```diff
@@ -319,246 +316,307 @@ TEST(MathOps, FloorModBroadcasting) {
...
-  auto R = ops::SquaredDifference(root, A, B);
+  auto R = ops::Minimum(root, A, B);
 
   vector<DataType> output_datatypes = {DT_FLOAT};
 
   std::vector<Output> sess_run_fetchoutputs = {R};
-  OpExecuter opexecuter(root, "SquaredDifference", static_input_indexes,
-                        output_datatypes, sess_run_fetchoutputs);
+  OpExecuter opexecuter(root, "Minimum", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
```
sayantan-nervana: Minimum also supports broadcasting. Lets have a test for that mode too.

test/test_array_ops.cpp
```diff
@@ -51,9 +51,71 @@ namespace testing {
...
+  int dim1 = 2;
+  int dim2 = 3;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+  AssignInputValuesRandom(A);
+
+  // axis at which the dimension will be inserted
+  // should be -rank-1 <= axis <= rank
+  auto axis = 0;
+
```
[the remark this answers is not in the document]
kanvi-nervana: I tested with axis = -1 and it passed. Is there a need for a separate test?

test/test_math_ops.cpp
kanvi-nervana: Added

test/test_array_ops.cpp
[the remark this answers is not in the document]
sayantan-nervana: I meant in a loop. you can have the same 'A' tensor of size 2x3, just varying the 'axis' variable is enough.

we should add it, to protect against regressions. even if it passes now, it could fail in the future.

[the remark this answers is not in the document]
kanvi-nervana: Done

PR #215, opened 2018-09-18
Mingshan/unit test update

mingshan-wang:
1. Add AssignInputValuesRandom() function to generate random values to populate the tensor with specified data type: int32, int64 and float. The function also takes a (min,max) range. 

2. Change some of the existing unit tests to use the new utility function.


test/test_utilities.cpp
```diff
@@ -85,6 +85,40 @@ void AssignInputValuesRandom(Tensor& A) {
...
+    value = value + min;  // range from min to max
+    // value =
+    //     roundf(value * 100) / 100.0;  // change the precision of the float to
+    //                                   // 2 number after the decimal
+    A_flat_data[i] = value;
+  }
+}
+
+template <>
+void AssignInputValuesRandom(Tensor& A, int64_t min, int64_t max) {
```
shresthamalik: Trying to understand the use-cases for specialized template.

1. We want to round the precision to two decimals for float data types? In which case it might be better to add the specialized template for float. As it is doing something more than the regular template.

2.  Do we expect explicit function calls for int32_t and int64_t?
And map both int32 and int64 to int , here ? 
``` auto A_flat = A.flat<int>();```

3. For explicitly assigning, int32 for DT_INT32 and int64 for DT_INT64? Which might not happen the way we have defined the template. I think

File: test/test_utilities.cpp [hunk 2]
```diff
@@ -85,6 +85,40 @@ void AssignInputValuesRandom(Tensor& A) {
...
+void AssignInputValuesRandom(Tensor& A, int32_t min, int32_t max) {
+  auto A_flat = A.flat<int>();
+  auto A_flat_data = A_flat.data();
+  srand(static_cast<unsigned>(time(0)));
+  for (int i = 0; i < A_flat.size(); i++) {
+    // give a number between 0 and max
+    int32_t value = static_cast<int32_t>(rand()) /
+                    static_cast<int32_t>(RAND_MAX / (int32_t)(max - min));
+    value = value + min;  // range from min to max
+    // value =
```
shresthamalik: Please remove the dead code.

test/test_utilities.h
```diff
@@ -51,6 +51,24 @@ void AssignInputValuesFromVector(Tensor& A, vector<T> x) {
   }
 }
 
+template <typename T>
+// Randomly generate data with specified type to populate the Tensor
+void AssignInputValuesRandom(Tensor& A, T min, T max) {
```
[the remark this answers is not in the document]
avijit-nervana: Agree. We should.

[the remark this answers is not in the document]
mingshan-wang: If the user might define a tensor A of data type DT_FLOAT and then call AssignInputValuesRandom(A, 5, 6) without specifying the data type, you will get a data mismatch error from our testing framework complaining expect float, but got int32. 

So I don't think I need to add one more check here.

test/test_utilities.cpp
mingshan-wang: Thanks for the detailed comments. Actually this is the old version of my change, I am sorry it lands here. I will push the latest change. 

1. You are right, we can create a special template for float and put those two lines there. 

2. I did separate int32 and int64 because for some inputs, in the documentation, it might say it can have type int32 or int64, so in that case, we can test different situation for different int types.

test/test_utilities.h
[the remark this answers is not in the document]
shresthamalik: The failure message comes from GTEST()? Or the unit test framework (Op Executor)?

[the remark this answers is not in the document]
mingshan-wang: I think it is due to some bug in our unit test framework. Kanvi fixed this and the PR just get merged.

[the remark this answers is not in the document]
mingshan-wang: Sorry, that failure msg is actually from TF complaining assigning a different data type value to a Tensor with certain data type. In this case, TF is expecting float, but we supplied int.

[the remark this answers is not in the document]
shresthamalik: So, it fails on call to AssignInputRandomValues() itself? Then, we are good. 👍  Thanks for testing.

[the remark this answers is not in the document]
mingshan-wang: Yes, I think that's the case. No problem, thanks for all the useful comments 👍

test/test_utilities.h
```diff
@@ -51,6 +51,21 @@ void AssignInputValuesFromVector(Tensor& A, vector<T> x) {
...
+  auto A_flat = A.flat<T>();
+  auto A_flat_data = A_flat.data();
+  srand(static_cast<unsigned>(time(0)));
+  for (int i = 0; i < A_flat.size(); i++) {
+    T value =
+        // randomly generate a number between 0 and (max-min)
+        static_cast<T>(rand()) / static_cast<T>(RAND_MAX / (max - min));
```
shresthamalik: For int , the current function generates a value between [min, max-1] and not [min, max]. Is that what we want? 

If we want [min, max] range, we would want to do change to below. This will be more intuitive. 

```
static_cast<T>(rand()) / static_cast<T>(RAND_MAX / (max+1 - min));
```

It would be helpful, if we can add a comment describing the behavior of the function, so that there is no confusion for the user. 

For e.g. in the SparseCrossEntropyTest(), we need to provide labels. So the tensor should be assigned an int between [0, num_of_classes -1] . 

Right now, if we call AssignInputValues(B, 0, num_of_classes) it does just that.

For a moment, I was confused whether it should be AssignInputValues(B, 0, num_of_classes-1)

test/test_utilities.cpp
```diff
@@ -85,6 +85,25 @@ void AssignInputValuesRandom(Tensor& A) {
...
+        // randomly generate a number between 0 and (max-min)
+        static_cast<float>(rand()) / static_cast<float>(RAND_MAX / (max - min));
+    value = value + min;  // transform the range to min and max
+    value =
+        roundf(value * 100) / 100.0;  // change the precision of the float to
+                                      // 2 number after the decimal
+    A_flat_data[i] = value;
+  }
+}
+
```
sayantan-nervana: Not sure why we are specializing for float.
Compared to the unspecialized function this seems to be an extra line
```value = roundf(value * 100) / 100.0```
Why is this needed, and wouldn't the unspecialized one just work?

I might have missed some discussion why this is needed

shresthamalik: Precision adds some readability when debugging, the numbers are easier to comprehend. Easier to add 2.46 + 3.23 than 2.46343 + 3.23455 . This is one of the use cases  I can think of.

sayantan-nervana: but the main reason for adding this was the expectation that fusedbatchnormgrad test would pass, which it doesn't with this change.
Also we can always print stuff with reduced precision, rather than curtailing precision at the source.

shresthamalik: I meant when you are doing the math in your head to verify the expected results. Useful when the ops are simple. We can use the other utility functions for this purpose, i guess. 👍

mingshan-wang: Hi @shresthamalik , @sayantan-nervana , some updates:

As I mentioned before, I added this restriction to solve the precision issue when testing FusedBatchNormGrad op.  

Now, @kanvi-nervana has a change for AssertTensorEquals() function where when comparing floats, she normalized the difference, and this exactly solves the precision problem we have seen before, so I also make a change for the FusedBatchNormGrad unit test to use the AssignInputValuesRandom() function and no decimal restriction is there.

In conclusion, the fundamental problem is solved by Kanvi's PR, and for my utility function, specialized template is not needed.

PR #217, opened 2018-09-19
Additions to .gitignore

aprocter:


.gitignore
```diff
@@ -7,3 +7,6 @@ cpu_codegen/
 
 *.pbtxt
 *.dot
+
+.*.swp
+.nfs*
```
sayantan-nervana: how about adding one for python?
*.py[cod]

PR #219, opened 2018-09-20
kanvi/Implement Max and Min op

kanvi-nervana:


test/test_math_ops.cpp
```diff
@@ -468,6 +468,33 @@ TEST(MathOps, Log2d) {
...
+  AssignInputValuesRandom(A);
+
+  // axis at which the dimension will be inserted
+  // should be -rank-1 <= axis <= rank
+  vector<int> axis_ = {0,-1};
+  
+  vector<int> static_input_indexes = {1};
+  vector<DataType> output_datatypes = {DT_FLOAT};
+  
+  for(auto const& axis:axis_){
```
shresthamalik: We decided to have a separate test for each case for unit tests. Please add separate tests for axis 0 and -1.

kanvi-nervana: All other tests with a similar inputs(axis to be specific) are done in this manner.

src/ngraph_builder.cc
```diff
@@ -1550,6 +1550,63 @@ static Status TranslateMatMulOp(
...
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_max_op));
+
+  bool tf_keep_dims;
+  if (GetNodeAttr(op->attrs(), "keep_dims", &tf_keep_dims) != Status::OK()) {
+    tf_keep_dims = false;
+  }
+
+  std::vector<int64> max_axes;
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &max_axes));
+
```
shresthamalik: We can add a check here to make sure the max_axes dimensions are within [-rank(input), rank(input))

src/ngraph_builder.cc
```diff
@@ -1550,6 +1550,63 @@ static Status TranslateMatMulOp(
   return Status::OK();
 }
 
+static Status TranslateMaxOp(const Node* op,
+                             const std::vector<const Tensor*>& static_input_map,
+                             Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input, ng_max_op;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_max_op));
```
shresthamalik: We do not need to fetch ng_max_op, not used anywhere

src/ngraph_builder.cc
```diff
@@ -1550,6 +1550,63 @@ static Status TranslateMatMulOp(
...
+        ng_result_shape_with_keep[i] = 1;
+      }
+    }
+
+    ng::AxisVector ng_axis_order(ng_max->get_shape().size());
+
+    for (size_t i = 0; i < ng_max->get_shape().size(); i++) {
+      ng_axis_order[i] = i;
+    }
+
```
sayantan-nervana: can be replaced by:
```
ng::AxisVector ng_axis_order(ng_max->get_shape().size());
std::iota(ng_axis_order.begin(), ng_axis_order.end(), 0);
```

sayantan-nervana:
Also, removing wip, so that ci tests it.

sayantan-nervana:
TESTNOW

test/test_math_ops.cpp
shresthamalik: That's fine. We decided to use this approach for the new tests that we implement.

src/ngraph_builder.cc
kanvi-nervana: The testing framework throws an error if I try to give a value that is out of bounds. So when I added the check, I have no way of testing it.

src/ngraph_builder.cc
```diff
@@ -1550,6 +1550,63 @@ static Status TranslateMatMulOp(
...
+  ng::AxisSet ng_reduction_axes;
+
+  for (auto i : max_axes) {
+    if (i < 0) {
+      ng_reduction_axes.insert(input_rank + i);
+    } else {
+      ng_reduction_axes.insert(i);
+    }
+  }
+
```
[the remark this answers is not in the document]
sayantan-nervana: transform might be a c++17 thing. There's for_each though:
```for_each (max_axes.begin(), max_axes.end(), [&ng_reduction_axes_vect, input_rank](int idx){ng_reduction_axes_vect.push_back(idx + (idx<0 ? input_rank : 0));});```

File: src/ngraph_utils.cc [hunk 6]
```diff
@@ -164,6 +164,17 @@ const gtl::ArraySlice<DataType>& NGraphIndexDTypes() {
   return result;
 }
 
+Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
+  for (auto i : axes) {
+    if (i >= -(rank) && i < rank) {
+      return errors::InvalidArgument(
+          "Axis Dimension is out of range.Should be in range [-rank(input), "
+          "rank(input))");
```
sayantan-nervana: `range.Should` -> `range. Should`

How about printing the actual numbers in the message, something like:
```
for (auto i = 0; i < axes.size(); i++) {
     if (axes[i] >= -rank && axes[i] < rank) {
       return errors::InvalidArgument(
           "Axis Dimension is out of range. Got " + axes[i] + " at position " + i +" should have be in range [-rank(input), rank(input)), that is [" + -rank + ", " + rank + ")");   //I am using string concat operator + loosely here
```

src/ngraph_utils.cc
```diff
@@ -164,6 +164,17 @@ const gtl::ArraySlice<DataType>& NGraphIndexDTypes() {
   return result;
 }
 
+Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
+  for (auto i : axes) {
+    if (i >= -(rank) && i < rank) {
```
shresthamalik: Shouldn't it be ?
```
if (i < -(rank) || i >= rank)
```

kanvi-nervana:
TESTNOW

src/ngraph_utils.cc
```diff
@@ -164,6 +164,17 @@ const gtl::ArraySlice<DataType>& NGraphIndexDTypes() {
   return result;
 }
 
+Status CheckAxisDimInRange(std::vector<int64> axes, size_t rank) {
+  for (auto i : axes) {
+    if (i < (int)-rank || i >= (int)rank) {
+      return errors::InvalidArgument("Axis Dimension is out of range. Got ", i,
```
sayantan-nervana: `rank` is of type `size_t`, so negating it will cause an unsigned wraparound. Not sure what `(int)-rank` will actually  mean.
How about casting first before negating:
`-(static_cast<int>(rank))`

kanvi-nervana: That is what was happening earlier when there was no cast - negating was causing the wraparound and the tests were failing. With the cast I am getting the correct values for the range and the tests are passing.

shresthamalik: Cast it before using inside if condition, for clarity? completely optional.

src/ngraph_builder.cc
kanvi-nervana: Added the check anyway

kanvi-nervana:
TESTNOW

PR #220, opened 2018-09-20
sarkars/Use NCHW

sayantan-nervana:
Adding Sharath's comment


sayantan-nervana:
Incorporating following comments from Sharath:
1) use NCHW (in benchmark as well as a general advice)
2) use 128 in inference instead of 32
3) shifted useful/variable options (like batchsize and model) to the end and more fixed ones (NCHW) to the beginnign

PR #224, opened 2018-09-21
Add more packages to Dockerfile for additions to daily inference model runs

crlishka:
The YoloV2 inference model requires an additional "librosa" python package to run.  I tried adding this as a "pip install librosa" in the command for the daily inference run, but encountered errors.  Thus, I am adding the package in the standard Dockerfiles for ngraph-tf.


crlishka:
TESTNOW

PR #227, opened 2018-09-25
Sindhu/implement argmaxop

sindhu-nervana:
I still have to make SimpleConfirmationFunction changes to run the op on CPU if dimension is more than 1 dimension. So WIP.


sindhu-nervana:
I verified that tensorflow ArgMax takes an integer as input for dimension so it will be only 1 dimension. Please review the PR

sindhu-nervana:
TESTNOW

src/ngraph_builder.cc
```diff
@@ -446,6 +447,37 @@ static Status TranslateAddNOp(
...
+  shared_ptr<ng::Node> ng_input, ng_dim;
+
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_dim));
+
+  std::vector<int64> tf_dim;
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &tf_dim));
+
+  ng::Shape input_shape = ng_input->get_shape();
+  size_t input_rank = ng_input->get_shape().size();
+
```
sayantan-nervana: `ng_input->get_shape().size()`
to
`input_shape.size()`

src/ngraph_builder.cc
```diff
@@ -446,6 +447,37 @@ static Status TranslateAddNOp(
...
+  // If input dimension is negative, make it positive
+  if (tf_dim[0] < 0) {
+    tf_dim[0] = input_rank + tf_dim[0];
+  }
```
sayantan-nervana: An assert that `tf_dim.size()==1`

src/ngraph_builder.cc
```diff
@@ -446,6 +447,37 @@ static Status TranslateAddNOp(
...
+
+  std::vector<int64> tf_dim;
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &tf_dim));
+
+  ng::Shape input_shape = ng_input->get_shape();
+  size_t input_rank = ng_input->get_shape().size();
+
+  // If input dimension is negative, make it positive
+  if (tf_dim[0] < 0) {
+    tf_dim[0] = input_rank + tf_dim[0];
```
aprocter: Might cast `input_rank` to `int64` to be certain.

src/ngraph_builder.cc
```diff
@@ -446,6 +447,37 @@ static Status TranslateAddNOp(
...
+  }
+  size_t input_dims = tf_dim[0];
+
+  DataType dtype = DT_INT64;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "output_type", &dtype));
```
sayantan-nervana: We need to read the datatype from the op. It could be DT_INT32 too. 

Something like:
```
DataType dtype;
TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "dtype", &dtype));
```

src/ngraph_builder.cc
```diff
@@ -22,6 +22,7 @@
 #include "ngraph/builder/autobroadcast.hpp"
 #include "ngraph/builder/numpy_transpose.hpp"
 
+#include "ngraph/op/argmax.hpp"
```
aprocter: Should probably move this above the preceding blank line, so there's one `ngraph` group and this is separate from the `tensorflow` group.

test/test_math_ops.cpp
```diff
@@ -853,6 +853,52 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+  AssignInputValuesRandom(A);
+
+  vector<int64> dim = {-1};
+
+  vector<int> static_input_indexes = {1};
+
+  auto R = ops::ArgMax(root, A, dim[0]);
+
+  vector<DataType> output_datatypes = {DT_INT64};
+
```
sayantan-nervana: Since we have DT_INT64 in the earlier, lets have DT_INT32 here

test/test_math_ops.cpp
```diff
@@ -853,6 +853,52 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+// ArgMax test for positive dimension
+TEST(MathOps, ArgMax_Pos) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 3;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+  AssignInputValuesRandom(A);
+
+  vector<int64> dim = {1};
```
shresthamalik: We don't need dims to be a vector as you have verified that the TF op takes a single dimension. Please make dim a scalar int.
Would be good to have a test setting the "output_type" attribute to INT_32. Please modify any of the test to add that.

src/ngraph_builder.cc
shresthamalik: This is the expected data type of the output set by the user as an attribute. Not sure why we need to read datatype of the op.

sayantan-nervana: Right. By 'datatype of the op', I meant 'datatype of the output set by the user'

sindhu-nervana:
@sayantan-nervana @shresthamalik Made the requested changes - added INT32 check, passed attribute to one of the tests and some more minor changes. Please review

src/ngraph_builder.cc
sindhu-nervana: This is the output data type attr used in TF '''DataType tensorflow::ops::ArgMax::Attrs::output_type_''' and the function '''TF_MUST_USE_RESULT Attrs tensorflow::ops::ArgMax::Attrs::OutputType(
  DataType x
)'''

sayantan-nervana: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L896

DT_INT64 is the default if nothing is specified. But it could be DT_INT32, hence we need to read that.

So this looks good.

src/ngraph_builder.cc
```diff
@@ -458,15 +458,17 @@ static Status TranslateArgMaxOp(
   TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &tf_dim));
 
   ng::Shape input_shape = ng_input->get_shape();
-  size_t input_rank = ng_input->get_shape().size();
+  size_t input_rank = input_shape.size();
+
+  tf_dim.size() == 1;
```
sayantan-nervana: This generates a boolean, but does not assert.
We need to return tf error status if the size != 1

shresthamalik: 👍

test/test_math_ops.cpp
```diff
@@ -853,6 +853,56 @@ TEST(MathOps, SquaredDifferenceBroadcasting) {
...
+// ArgMax test for positive dimension
+TEST(MathOps, ArgMax_Pos) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 3;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2}));
+  AssignInputValuesRandom(A);
+
+  int dim = {1};
```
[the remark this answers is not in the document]
shresthamalik: Interestingly, this does :) @sayantan-nervana

PR #228, opened 2018-09-25
Mingshan/sigmoid grad op

mingshan-wang:
Implement the SigmoidGrad op to be used for GNMT model


src/ngraph_builder.cc
```diff
@@ -1491,6 +1491,25 @@ static Status TranslateFusedBatchNormGradOp(
...
 
+static Status TranslateSigmoidGradOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input;
+  shared_ptr<ng::Node> ng_delta;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_delta));
+
+  auto ng_mul = std::make_shared<ng::op::Multiply>(ng_input, ng_delta);
+  ng::Shape input_shape = ng_input->get_shape();
```
sayantan-nervana: ng nodes have support for basic arithmetic ops, so one can use: `ng_input * ng_delta`

src/ngraph_builder.cc
```diff
@@ -1491,6 +1491,25 @@ static Status TranslateFusedBatchNormGradOp(
...
+  std::vector<std::string> constant_values(ng::shape_size(input_shape), "1");
+  auto c = make_shared<ng::op::Constant>(ng_input->get_element_type(),
+                                         input_shape, constant_values);
+  auto ng_subtract = std::make_shared<ng::op::Subtract>(c, ng_input);
```
sayantan-nervana: `c-ng_input` should work.

If broadcast is supported (I think it is), this might just work:
`auto ng_subtract = make_shared<ng::op::Constant>(ng_input->get_element_type(), ng::Shape(1), std::vector<int>({1})) - ng_input;`

src/ngraph_builder.cc
```diff
@@ -1491,6 +1491,25 @@ static Status TranslateFusedBatchNormGradOp(
...
+  auto ng_result = std::make_shared<ng::op::Multiply>(ng_mul, ng_subtract);
+
```
sayantan-nervana: `ng_mul * ng_subtract`

src/ngraph_mark_for_clustering.cc
```diff
@@ -266,6 +266,7 @@ Status MarkForClustering(Graph* graph) {
       type_constraint_map["Shape"]["T"] = NGraphDTypes();
       type_constraint_map["Shape"]["out_type"] = NGraphIndexDTypes();
       type_constraint_map["Sigmoid"]["T"] = NGraphNumericDTypes();
+      type_constraint_map["SigmoidGrad"]["T"] = NGraphNumericDTypes();
       type_constraint_map["Sign"]["T"] = NGraphNumericDTypes();
```
sayantan-nervana: Not sure if SigmoidGrad would work or make sense for int types. Can you please double check what `T` ought to be

If it works for int types, lets make one of the test cases with int numbers

src/ngraph_builder.cc
mingshan-wang: Oh, good to know. I will make all the changes accordingly. Thanks!

src/ngraph_mark_for_clustering.cc
mingshan-wang: I verified in the gen_math_ops.py file, the input to SigmoidGrad only supports `bfloat16`, `half`, `float32`, `float64`, `complex64`, `complex128`.

So there is no int type, then I need to change to `NgraphDTypes()` for non-int type constraints?

shresthamalik: For most ops, we are using NGraphNumericDTypes() as a superset, expecting tf to complain about wrong datatypes, before it reaches us. Either tf will complain, or tf (without nGraph) would fail too, worth checking.

sayantan-nervana: I think we should have "DT_FLOAT, DT_DOUBLE" in a separate ArraySlice, but that is a change not related to this PR I suppose. So no NGraphNumericDTypes() looks fine for now.

mingshan-wang: Sounds good. I will still use `NGraphNumericDTypes()` for SigmoidGrad op.

PR #232, opened 2018-09-27
Sarkars/Minor correction in instruction

sayantan-nervana:


sayantan-nervana:
Thanks Vinod for using it and pointing out the mistake

PR #239, opened 2018-09-28
Shrestha/BugFix_Variable_Capture_Pass_Manifested_In_DRAW_Model

shresthamalik:
In ngraph_capture_variable.cc we replace the tf Variable V2 nodes with nGraphVariable nodes. While we were replacing all the outgoing edges from V2, with the new source, we were not replacing the incoming edges to V2, with the new destination. The bug manifested in DRAW model. This commit fixes it.

- Tested draw model on tf-ng and ng-tf both


shresthamalik:
TEST_NOW

shresthamalik:
TESTNOW

PR #240, opened 2018-09-28
Implement All op for GNMT training

kanvi-nervana:


src/ngraph_mark_for_clustering.cc
```diff
@@ -209,6 +209,8 @@ Status MarkForClustering(Graph* graph) {
       type_constraint_map["Abs"]["T"] = NGraphNumericDTypes();
       type_constraint_map["Add"]["T"] = NGraphNumericDTypes();
       type_constraint_map["AddN"]["T"] = NGraphNumericDTypes();
+      type_constraint_map["All"]["T"] = NGraphDTypes();
```
sayantan-nervana: According to the op registration, input can only be bool
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/math_ops.cc#L1219

since it is only 1 type, we can remove this line:
`type_constraint_map["All"]["T"] = NGraphDTypes();`

test/test_math_ops.cpp
```diff
@@ -140,6 +140,109 @@ TEST(MathOps, AddN) {
...
+  int dim2 = 3;
+
+  std::vector<bool> v;
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(false);
+  v.push_back(false);
+  Tensor A(DT_BOOL, TensorShape({dim1, dim2}));
```
sayantan-nervana: minor suggest:
```std::vector<bool> v = {true, true, true, true, false, false};```

File: test/test_math_ops.cpp [hunk 3]
```diff
@@ -140,6 +140,109 @@ TEST(MathOps, AddN) {
...
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(false);
+  v.push_back(false);
+  v.push_back(true);
+  v.push_back(false);
+  v.push_back(false);
+  Tensor A(DT_BOOL, TensorShape({dim1, dim2}));
```
sayantan-nervana: minor suggest:
```std::vector<bool> v = {true, true, true, true, false, false, true, false, false};```

File: test/test_math_ops.cpp [hunk 4]
```diff
@@ -140,6 +140,109 @@ TEST(MathOps, AddN) {
...
+TEST(MathOps, AllKeepDims) {
+  int dim1 = 2;
+  int dim2 = 2;
+
+  std::vector<bool> v;
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  v.push_back(true);
+  Tensor A(DT_BOOL, TensorShape({dim1, dim2}));
```
sayantan-nervana: minor suggest:
std::vector<bool> v = {true, true, true, true};

kanvi-nervana:
TESTNOW

src/ngraph_mark_for_clustering.cc
```diff
@@ -307,6 +308,7 @@ Status MarkForClustering(Graph* graph) {
       confirmation_functions["Abs"] = SimpleConfirmationFunction();
       confirmation_functions["Add"] = SimpleConfirmationFunction();
       confirmation_functions["AddN"] = SimpleConfirmationFunction();
+      confirmation_functions["All"] = SimpleConfirmationFunction({1});
       confirmation_functions["AvgPool"] = SimpleConfirmationFunction();
```
sayantan-nervana: good catch @kanvi-nervana

kanvi-nervana:
TESTNOW

src/ngraph_builder.cc
```diff
@@ -446,6 +446,62 @@ static Status TranslateAddNOp(
...
+
+  ng::Shape input_shape = ng_input->get_shape();
+  size_t input_rank = ng_input->get_shape().size();
+
+  TF_RETURN_IF_ERROR(CheckAxisDimInRange(all_axes, input_rank));
+
+  std::vector<size_t> ng_reduction_axes_vect(all_axes.size());
+  std::transform(
+      all_axes.begin(), all_axes.end(), ng_reduction_axes_vect.begin(),
+      [input_rank](int idx) { return idx + (idx < 0 ? input_rank : 0); });
```
shresthamalik: We might want to typecast input_rank to int

test/test_math_ops.cpp
```diff
@@ -140,6 +140,91 @@ TEST(MathOps, AddN) {
...
+  AssignInputValuesFromVector<bool>(A, v);
+
+  // axis at which the dimension will be inserted
+  // should be -rank <= axis < rank
+  vector<int> axis_ = {0};
+
+  vector<int> static_input_indexes = {1};
+  vector<DataType> output_datatypes = {DT_BOOL};
+
+  for (auto const& axis : axis_) {
```
shresthamalik: Do we need the for loop?

src/ngraph_builder.cc
```diff
@@ -446,6 +446,62 @@ static Status TranslateAddNOp(
   return Status::OK();
 }
 
+static Status TranslateAllOp(const Node* op,
+                             const std::vector<const Tensor*>& static_input_map,
+                             Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input, ng_axes_op;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input, &ng_axes_op));
+
```
shresthamalik: We can add a check for invalid data type for input, since it can have only DT.BOOL. It would be also interesting to see, where the op fails(if it fails at all) if we pass input tensor of data type other than bool. 

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt

kanvi-nervana: After talking to @sayantan-nervana my understanding is that tf will throw an error in case of an invalid datatype and it will not reach nGraph.

kanvi-nervana:
TESTNOW

src/ngraph_builder.cc
shresthamalik: If we can confirm TF is doing this check for us, we can close this. Sometimes the TF checks are done in kernel, in which case we might not reach the check.

kanvi-nervana:
TESTNOW

PR #245, opened 2018-10-02
Kanvi/implement zeroslike

kanvi-nervana:


src/ngraph_builder.cc
```diff
@@ -2960,6 +2960,22 @@ static Status TranslateUnpackOp(
...
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ng::Node> ng_input;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+
+  ng::Shape input_shape = ng_input->get_shape();
+  std::vector<std::string> const_values(ng::shape_size(input_shape), "0");
+  auto ng_const = make_shared<ng::op::Constant>(ng_input->get_element_type(),
+                                                input_shape, const_values);
+  auto ng_result = std::make_shared<ng::op::Multiply>(ng_input, ng_const);
+
```
sayantan-nervana: we don't need to multiply it right?
`ng_const` is our final output.

src/ngraph_mark_for_clustering.cc
```diff
@@ -302,6 +302,7 @@ Status MarkForClustering(Graph* graph) {
       type_constraint_map["Transpose"]["T"] = NGraphDTypes();
       type_constraint_map["Transpose"]["Tperm"] = NGraphIndexDTypes();
       type_constraint_map["Unpack"]["T"] = NGraphDTypes();
+      type_constraint_map["ZerosLike"]["T"] = NGraphNumericDTypes();
```
sayantan-nervana: The registration of zeros like: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/array_ops.cc#L736

It accepts all types of data even booleans. Hence we should not use NGraphNumericDTypes();
We can skip this line totally, as there is no type constraint.

PR #247, opened 2018-10-02
Mingshan/Size op

mingshan-wang:
`Size` op is required for GNMT training.


mingshan-wang:
TESTNOW

mingshan-wang:
TESTNOW

src/ngraph_builder.cc
```diff
@@ -2309,6 +2309,34 @@ static Status TranslateSigmoidOp(
...
+  // Size has an attribute to specify output, int32 or int64
+  ng::element::Type type;
+  TF_RETURN_IF_ERROR(TFDataTypeToNGraphElementType(dtype, &type));
+
+  auto ng_input_shape = ng_input->get_shape();
+
+  int32 result = 1;
+  for (auto dim : ng_input_shape) {
+    result *= dim;
+  }
```
sayantan-nervana: We could use int64/long here.
tensor sizes might be very big, eg: 128*64*64*512 etc

mingshan-wang: Yes, I will make this change now.

PR #248, opened 2018-10-02
Implement ArgMin Op

sindhu-nervana:


test/test_math_ops.cpp
```diff
@@ -336,6 +336,55 @@ TEST(MathOps, ArgMax_Pos) {
   opexecuter.RunTest();
 }  // end of test op ArgMax
 
+// ArgMin test for negative dimension
+TEST(MathOps, ArgMin_Neg) {
```
shresthamalik: Please do not use '_' in the name. Neither TestCaseName nor TestName should contain underscore. Please change it in the ArgMax test also. Didnt notice that.

PR #249, opened 2018-10-04
Shrestha/Unit Test FW : Add APIs to allow execution only on nGraph or on TF

shresthamalik:
For some ops like Conv2DBackPropFilter TF does not have supporting kernels for NCHW data format, while nGraph does. In such cases, we execute the graph on nGraph, but before running on TF we reshape ips to NHWC format and then run on TF. Finally we can compare the outputs.

There might be other cases when one wants to execute only on nGraph or on TF and get outputs. 

- Added test cases for Conv2DBackPropFilterNCHW as example


test/opexecuter.h
```diff
@@ -70,14 +70,29 @@ class OpExecuter {
...
+  // Returns outputs
+  void ExecuteOnTF(vector<Tensor>& outputs);
+
+  // Executes on NGraph, then executes on TF, and compares the results
   void RunTest();
 
+  // Compares NGraph and TF outputs
+  static void CompareNGraphAndTF(const vector<Tensor>& tf_outputs,
+                                 const vector<Tensor>& ngraph_outputs);
+
```
shresthamalik: Contemplating moving CompareNGraphAndTF function to utilities. It is comparing two vectors of Tensor (it is the context in which it is used that makes one vector tf outputs and the other ngraph outputs), and is quite independent of the OpExecuter class. Any thoughts?

```
static void CompareNGraphAndTF(const vector<Tensor>& tf_outputs,
                                 const vector<Tensor>& ngraph_outputs);
```

sayantan-nervana: Sounds like a good idea

shresthamalik:
TESTNOW

test/opexecuter.h
mingshan-wang: Yes, the functionality fits more like a utility function.

test/test_nn_ops.cpp
```diff
@@ -106,46 +106,198 @@ TEST(NNOps, BiasAddGrad) {
...
+  initializer_list<int> filter_size_HWIO = {3, 3, 2, 2};
+  vector<DataType> output_datatypes = {DT_FLOAT};
+  vector<int> static_input_indexes = {1};
+
+  // Define scope for nGraph
+  // Data Format : NCHW
+  Scope ngraph_scope = Scope::NewRootScope();
+  vector<int64> input_size_NCHW = {1, 2, 7, 6};
+  Tensor input_data_NCHW(DT_FLOAT, TensorShape(input_size_NCHW));
+  AssignInputValuesAnchor(input_data_NCHW, -1.1f);
```
mingshan-wang: Hi @shresthamalik , is there a reason you are using `AssignInputValuesAnchor()` here? I think it will be good to just use random values via AssignInputValuesRandom() and gives a (min,max) range.

test/opexecuter.cpp
```diff
@@ -141,36 +149,44 @@ void OpExecuter::ExecuteOnTF() {
...
+               "expected output datatype."
+            << expected_dtype;
     }
   }
 }
 
+void OpExecuter::ExecuteOnNGraph(vector<Tensor>& outputs) {
+  ExecuteOnNGraph();
+  outputs = ngraph_outputs_;
+}
```
sayantan-nervana: So ExecuteOnNGraph is now overloaded. one is private and one is public.
The public version, just calls the private version and returns the ngraph_outputs_

How about we just have a getter function for ngraph_outputs_? Then we have only 1 private ExecuteOnNGraph()

test/opexecuter.cpp
```diff
@@ -127,7 +130,12 @@ OpExecuter::~OpExecuter() {}
 void OpExecuter::RunTest() {
   ExecuteOnNGraph();
   ExecuteOnTF();
-  CompareNGraphAndTF();
+  CompareNGraphAndTF(tf_outputs_, ngraph_outputs_);
+}
```
sayantan-nervana: It is a good idea to have CompareNGraphAndTF in the utility

however with the slight change (using getter), we don't have to change CompareNGraphAndTF's interface (which earlier was accepting nothing, but now accepts 2 inputs)

test/test_nn_ops.cpp
```diff
@@ -106,46 +106,198 @@ TEST(NNOps, BiasAddGrad) {
...
+  OpExecuter opexecuter_tf(tf_scope, "Conv2DBackpropFilter",
+                           static_input_indexes, output_datatypes,
+                           sess_run_fetchoutputs_tf);
+
+  vector<Tensor> tf_outputs;
+  opexecuter_tf.ExecuteOnTF(tf_outputs);
+
+  // Compare NGraph and TF Outputs
+  OpExecuter::CompareNGraphAndTF(tf_outputs, ngraph_outputs);
+}
```
sayantan-nervana: so we have:
```
stuff
vector<Tensor> ngraph_outputs;
opexecuter_ngraph.ExecuteOnNGraph(ngraph_outputs);
stuff
vector<Tensor> tf_outputs;
opexecuter_tf.ExecuteOnTF(tf_outputs);
OpExecuter::CompareNGraphAndTF(tf_outputs, ngraph_outputs);
```

but, ngraph_outputs is just ngraph_outputs_
and tf_outputs is just tf_outputs_
so we pull them out of the class and then send them into CompareNGraphAndTF. That sounds roundabout.

So we are creating 2 opexecutors. instead how about:
```
create opexp
stuff
opexp.ExecuteOnNGraph();
opexp.clear_graph()
opexp.set_new_graph() . // a setter for graph or whatever needs to change
stuff
opexp.ExecuteOnTF();
OpExecuter::CompareNGraphAndTF();
```

Even if we do not do the above  rearraggement (setter for new graph/scope), I think the (getter for tf_outputs_ and ngraph_outputs_ + no overloading) would help.

test/opexecuter.cpp
kanvi-nervana: I like the idea of the getter function as well.

test/test_nn_ops.cpp
shresthamalik: Yes, i forgot to update it to use Random. Will update this.

test/opexecuter.h
avijit-nervana: 👍

shresthamalik:
TESTNOW

test/opexecuter.cpp
```diff
@@ -125,49 +130,20 @@ OpExecuter::OpExecuter(const Scope sc, const string test_op,
...
 
 void OpExecuter::RunTest() {
-  ExecuteOnNGraph();
-  ExecuteOnTF();
-  CompareNGraphAndTF();
+  vector<Tensor> ngraph_outputs;
+  ExecuteOnNGraph(ngraph_outputs);
+  vector<Tensor> tf_outputs;
+  ExecuteOnTF(tf_outputs);
+  Compare(tf_outputs, ngraph_outputs);
```
avijit-nervana: This is much better

test/graph_exec.cpp
```diff
@@ -107,6 +109,8 @@ TEST(graph_exec, axpy) {
   // TODO
 }
 
+}  // namespace testing
+
 }  // namespace ngraph_bridge
 
-}  // namespace tensorflwo
+}  // namespace tensorflow
```
sayantan-nervana: good catch :D

test/test_utilities.cpp
```diff
@@ -107,14 +90,49 @@ void ValidateTensorData(Tensor& T1, Tensor& T2, float tol) {
...
+  ASSERT_EQ(v1.size(), v2.size());
+  for (int i = 0; i < v1.size(); i++) {
+    auto expected_dtype = v1[i].dtype();
+    switch (expected_dtype) {
+      case DT_FLOAT:
+        Compare<float>(v1[i], v2[i]);
+        break;
+      case DT_INT8:
+        Compare<int8>(v1[i], v2[i]);
+        break;
```
sayantan-nervana: Slight confusion here:
Compare is an untemplated function that compares 2 vectors of tensors

Compare is also templated function that compares individual tensors?

Maybe we should have different and more descriptive names for them? especially because in a situation when someone wants to compare 2 tensors, one should do this
```
Compare({t1}, {t1}) . //pass singleton vectors
```

But by mistake if you write:
```
Compare(t1, t2)
```

Then the compiler will understand it as 
```
template <class T>
bool Compare(T arg0, T arg1) {
  return arg0 == arg1;
}
```

which was not our intention




Compare is also overloaded to compare POD:
```Compare<int8>(v1[i], v2[i]);```

Comment (shresthamalik) [reply test/test_utilities.cpp]: There is a Compare function for tensors too, so when someone calls 
```
Compare(t1, t2)
```
It calls
```
// Compares two Tensors
template <typename T>
void Compare(const Tensor& T1, const Tensor& T2) {
```
and not the
```
// Compares two arguments
template <class T>
bool Compare(T arg0, T arg1) {
  return arg0 == arg1;
}
```
I tested this with logs. So, we might be safe here.

However, I see this being a problem for unsupported types.

For eg some new class A
```
Compare(a1, a2) 
```
might get us into trouble.

Are there any other corner cases where the Compare function overload might be an issue?

PR #251, opened 2018-10-04
Avijit/klockwork scan fixes

avijit-nervana:


src/ngraph_api.cc
```diff
@@ -51,7 +51,7 @@ vector<string> ListBackends() {
 tensorflow::Status SetBackend(const string& type) {
   try {
     ngraph::runtime::Backend::create(type);
-  } catch (const runtime_error& e) {
+  } catch (const runtime_error&) {
     return tensorflow::errors::Unavailable("Backend unavailable: ", type);
   }
```
sayantan-nervana: Should we print e.what() too, in case ngraph returns some useful error messages?

PR #253, opened 2018-10-08
Shrestha/unit test fw support multiple backends

shresthamalik:
Added Functionality to support multiple backends for Unit Test FW.
We can either set the backend via environment variable NGRAPH_TF_BACKEND or pass the backend name to the function calls RunTest() and ExecuteOnNGraph().
Environment Variable takes precedence over the parameters sent via Function Calls


test/opexecuter.cpp
```diff
@@ -320,8 +321,17 @@ void OpExecuter::ExecuteOnNGraph(vector<Tensor>& ngraph_outputs) {
...
-  // Create the nGraph backend
-  auto backend = ng::runtime::Backend::create("CPU");
+  // If NGRAPH_TF_BACKEND is set create that backend
+  // Else create backend of type ng_backend_name
+  const char* ng_backend_env_value = std::getenv("NGRAPH_TF_BACKEND");
+  if (ng_backend_env_value != nullptr) {
+    string backend_env = std::string(ng_backend_env_value);
+    if (!backend_env.empty()) {
+      ng_backend_name = backend_env;
+    }
```
sayantan-nervana: before setting a backend, we could check if that string is supported using ngraph_backends_len 
 + ngraph_list_backends
```
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_api.cc#L18
```

test/opexecuter.cpp
```diff
@@ -320,8 +321,17 @@ void OpExecuter::ExecuteOnNGraph(vector<Tensor>& ngraph_outputs) {
...
+  }
+  NGRAPH_VLOG(5) << " Creating NG Backend " << ng_backend_name;
+  auto backend = ng::runtime::Backend::create(ng_backend_name);
```
sayantan-nervana: and in case we do not apply the check i mention above, we can try-catch the backend creation, to know that incorrect backends were not requested

test/opexecuter.cpp
shresthamalik: NGraph throws an error if we pass an invalid string. Had the same concern, so checked this scenario.
The error looks like this with NGRAPH_TF_BACKEND=U
```
unknown file: Failure
C++ exception with description "Backend 'U' not registered. Error:/localdisk/shrestha/tf-1.11/ngraph-tf/build/ngraph/ngraph_dist/lib/libu_backend.so: cannot open shared object file: No such file or directory" thrown in the test body.
```

test/opexecuter.cpp
```diff
@@ -129,9 +129,9 @@ OpExecuter::OpExecuter(const Scope sc, const string test_op,
 // Destructor
 OpExecuter::~OpExecuter() {}
 
-void OpExecuter::RunTest() {
+void OpExecuter::RunTest(string ng_backend_name) {
```
avijit-nervana: You could use `const string&` instead of `string`

test/opexecuter.h
```diff
@@ -72,20 +72,23 @@ class OpExecuter {
...
-  // Executes on NGraph, then executes on TF, and compares the results
-  void RunTest();
+  // Executes on NGraph backend, then executes on TF, and compares the results
+  // NOTE: Env Variable NGRAPH_TF_BACKEND if set, overrides ng_backend_name
+  void RunTest(string ng_backend_name = "CPU");
 
   // TODO(malikshr) : Overload RunTest() to take in tolerance
-  static void RunTest(float tolerance);
+  // NOTE: Env Variable NGRAPH_TF_BACKEND if set, overrides ng_backend_name
+  void RunTest(float tolerance, string ng_backend_name = "CPU");
```
avijit-nervana: Is this version used currently or for future?

shresthamalik: Future, right now we do not have support for tolerance while running test

avijit-nervana: In that case you may want to remove this declaration.

test/opexecuter.cpp
shresthamalik: @avijit-nervana Incorporated the change.

PR #254, opened 2018-10-08
[WIP] Automate Mac builds

crlishka:
[TBD]


crlishka:
TESTNOW

crlishka:
Added comment that the script was developed for and tested with High Sierra.

PR #256, opened 2018-10-09
Mingshan/conv2dbackpropinput c++ test

mingshan-wang:
Since the pytest for the conv2DBackpropInput tests are failing right now, I added this C++ unit test for Conv2DBackpropInput to ensure the correctness of the op.


test/test_nn_ops.cpp
```diff
@@ -300,51 +300,238 @@ TEST(NNOps, Conv2DBackpropFilterNHWCValid) {
...
+TEST(NNOps, Conv2DBackpropInputNCHWSame) {
+  // Input NHWC :[batch, in_channels, in_height, in_width]
+  string padding_type = "SAME";
+  initializer_list<int32> input_size_NCHW = {1, 2, 7, 6};  // an integer vector
+  initializer_list<int32> input_size_NHWC = {1, 7, 6, 2};
+
+  // Filter :[filter_height, filter_width, in_channels, out_channels]
+  vector<int64> filter_size_HWIO = {3, 3, 2, 2};
+  // Out_delta :[batch, out_channels, out_height, out_width]
+  vector<int64> output_del_size_valid_NCHW = {1, 2, 4, 3};
```
shresthamalik: Please rename this to refer to the padding type "SAME"

mingshan-wang:
@sayantan-nervana Right now the pytest version is not running in CI, and maybe we want to keep the test file since previously I encountered an error (TF output and Ngraph output differs by one large number) so I added this C++ test to ensure the correctness of this op translation. And we still didn't know why the pytest is giving such discrepancy. May need to debug later.

mingshan-wang:
Hi @sayantan-nervana, currently TF CPU doesn't support dilation rate greater than 1. So I couldn't pass other numbers to TF CPU case.

mingshan-wang:
Ok. I will try on a GPU machine(maybe my mac) and get the result from Tensorflow, and set the value in the unit test file and compare with ngraph results.

test/test_nn_ops.cpp
```diff
@@ -300,51 +300,436 @@ TEST(NNOps, Conv2DBackpropFilterNHWCValid) {
...
+
+  Tensor filter(DT_FLOAT, TensorShape(filter_size_HWIO));
+  size = 1;
+  for (auto num : filter_size_HWIO) {
+    size *= num;
+  }
+  std::list<float> filter_list(size);
+  std::iota(filter_list.begin(), filter_list.end(), 0);
+  std::vector<float> filter_vector{std::begin(filter_list),
+                                   std::end(filter_list)};
```
sayantan-nervana: so `filter_vector` is just a copy of `filter_list` right? Why have 2 variables?


more minor comments:
```
int size_vect = std::accumulate(filter_size_HWIO .begin(), filter_size_HWIO .end(), 1, std::multiplies<float>());
std::vector<float> filter_vector();
std::iota(filter_vector.begin(), filter_vector.end(), 0);
```
the above snippet should take care of lines 404 to 411

similarly for lines 393 to 400

mingshan-wang: Yes, `filter_vector` and `filter_list` are the same, was trying to convert a list to a vector, and then pass into the `AssignInputValues()` function.

Thanks for the code snippet, this is much simpler, I will rewrite all the similar parts using the `accumulate function`.

mingshan-wang: Hey @sayantan-nervana , I eliminate the redundant `filter_list`, so there is only one variables instead of two. And I changed the `std::multiples<float>` to `std::multiple<int64>` in my latest change.

test/test_nn_ops.cpp
```diff
@@ -132,7 +132,7 @@ TEST(NNOps, Conv2DBackpropFilterNCHWSame) {
   auto filter_sizes = ops::Const(ngraph_scope, filter_size_HWIO);
   vector<int> stride_NCHW = {1, 1, 2, 2};
 
-  // Dialtion rates > 1 not supported by TF
+  // Dialtion rates > 1 not supported by TF on CPU
```
sayantan-nervana: `Dialtion`->`Dilation`

PR #257, opened 2018-10-10
Added Cmake changes related to Centos build. 

ransford2011:
CentOs creates `lib64` directory whereas Ubuntu creates `lib` directory.  Added changes for that distinction


ransford2011:
I've added you @wenhaoli-aipg so that you are aware of the changes I'm making related to CentOs builds

PR #259, opened 2018-10-15
Shrestha/fix for tf master changes

shresthamalik:
In the latest TF Master, the API for node->AddAttr() has changed causing unexpected behavior in deadness analysis. This PR
    1.  Separates the mark_for_clustering_checks and setting node attributes.
    2.  Renames the TF version check macro, to TF_ENABLE_DEADNESS_CHECK, so that we can enable/disable deadness as required using any criteria (version check or any other)


src/ngraph_mark_for_clustering.cc
```diff
@@ -165,44 +176,174 @@ Status MarkForClustering(Graph* graph) {
...
+
+  //
+  // A map of op types (e.g. "Add") to set_attribute functions. These can be
+  // used to set any additional attributes. For example:
+  //
+  //    confirmation_function_map["MyOp"] = [](Node* n) {
+  //     if(n->condition()){
+  //        T dummy=5;
+  //        n->AddAttr("_ngraph_dummy_attr", T);
+  //      }
```
sayantan-nervana:
```
int dummy = 5;
n->AddAttr("_ngraph_dummy_attr", dummy);
```

src/ngraph_mark_for_clustering.cc
```diff
@@ -98,36 +98,47 @@ static Status TypeConstraintOk(Node* node,
...
 //
-// Marks the input indices in "inputs" as static, i.e., inputs that must be
-// driven either by an _Arg or by a Const in the encapsulated graph.
+// Marks the input indices in "inputs" as static (meaning
+// that its value must be known at translation-to-nGraph time)
+//
+// Otherwise, it marks the zeroth input to the node as static (meaning
+// that its value must be known at translation-to-nGraph time, and accepts
+// placement.
 //
```
sayantan-nervana: ```marks the zeroth input to the node as static ``` : Did not understand the marking of zeroth input

```accepts placement``` : but this function just sets static inputs and does nothing about accepting or rejecting nodes.

Comment (sayantan-nervana) [reply src/ngraph_mark_for_clustering.cc]: Also the earlier comment was a defining characteristic of static input. Maybe we should retain that comment:
```
// Marks the input indices in "inputs" as static, i.e., inputs that must be
 // driven either by an _Arg or by a Const in the encapsulated graph.
```

Comment (shresthamalik) [reply src/ngraph_mark_for_clustering.cc]: I did not understand the 'zeroth input' comment either. This comment was in our current version, just moved it around. I thought i had changed this comment. Forgot to save it, I guess.

File: src/ngraph_mark_for_clustering.cc [hunk 3]
```diff
@@ -439,7 +492,7 @@ Status MarkForClustering(Graph* graph) {
         break;
       }
 
-#if (TF_VERSION_GEQ_1_11)
+#if (TF_ENABLE_DEADNESS_CHECK)
```
avijit-nervana: I would recommend renaming this to NGRAPH_TF_DISABLE_DEADNESS_CHECK and reverse the logic i.e., `#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)`

src/ngraph_version_utils.h
```diff
@@ -23,5 +23,5 @@
    ((TF_MAJOR_VERSION == REQ_TF_MAJ_VER) &&            \
     (TF_MINOR_VERSION >= REQ_TF_MIN_VER)))
 
-#define TF_VERSION_GEQ_1_11 TF_VERSION_GEQ(1, 11)
+#define TF_ENABLE_DEADNESS_CHECK TF_VERSION_GEQ(1, 11)
```
avijit-nervana: If we go by my earlier recommendation of using NGRAPH_TF_DISABLE_DEADNESS_CHECK then this line won't be needed.

src/ngraph_mark_for_clustering.cc
shresthamalik: Ok. Would this be an environment variable NGRAPH_TF_DISABLE_DEADNESS_CHECK?

avijit-nervana: That's what I am leaning towards. By using this logic - if we can turn off the deadness check (for some experiments/tests) - then that's good. If not then I guess what you have is fine

shresthamalik:
TESTNOW

PR #260, opened 2018-10-15
Mingshan/Support SpaceToDepth and DepthToSpace in the bridge

mingshan-wang:
Implement SpaceToDepth and DepthToSpace ops (both array ops) in the bridge.


src/ngraph_builder.cc
```diff
@@ -2475,6 +2609,111 @@ static Status TranslateSoftmaxOp(
...
+                                       " is not divisible by block_size ",
+                                       block_size);
+      }
+      break;
+
+    default:
+      return errors::InvalidArgument(
+          "SpaceToDepth supported data format is NCHW, NHWC, or NCHW_VECT_C");
+  }
+
```
sayantan-nervana: similar comment here. the switch-case can probably be done away with

src/ngraph_builder.cc
```diff
@@ -1351,6 +1351,140 @@ static Status TranslateConv2DBackpropInputOp(
...
+      ng_transpose_shape.push_back(channel_dimension + num_spatial_dimensions);
+
+      output_shape.push_back(input_shape[0]);
+      for (int i = 0; i < num_spatial_dimensions; i++) {
+        output_shape.push_back(input_shape[i + 1] * block_size);
+      }
+      output_shape.push_back(input_shape[channel_dimension] / num_blocks);
+      break;
+    }  // end of case NHWC
+    // NCHW
```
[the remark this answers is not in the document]
mingshan-wang: Here I am actually already doing only format dependent code(populating the shape to be used later) inside the `switch` and actually computation outside the switch statement. Inside `switch`, the code looks similar, but actually the order of computation differs and matters. So I am keeping the minimal code inside switch statement now.

src/ngraph_builder.cc
mingshan-wang: Will make the proper changes related to switch statement. Thank you for the suggestion!

src/ngraph_builder.cc
```diff
@@ -1351,6 +1351,136 @@ static Status TranslateConv2DBackpropInputOp(
...
+  shared_ptr<ng::Node> ng_input;
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+
+  // Get the attributes
+  int64 block_size;
+  std::string tf_data_format;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "block_size", &block_size));
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "data_format", &tf_data_format));
+
+  // The input tensor's height and width must be divisible by block_size
```
sayantan-nervana: this comment could be moved to where the check is being performed (line 1390)

mingshan-wang: Sure. Thanks for catching this.

src/ngraph_builder.cc
```diff
@@ -2475,6 +2605,109 @@ static Status TranslateSoftmaxOp(
...
+          std::vector<size_t> strides = {1, 1, (size_t)block_size,
+                                         (size_t)block_size};
+          strided_slice_result.push_back(
+              make_shared<ng::op::Slice>(ng_input, begin, upper, strides));
+          counter_width = counter_width + 1;
+        }
+        counter_height = counter_height + 1;
+      }
+      break;
+  }
```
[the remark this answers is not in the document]
mingshan-wang: Sounds good! This way is cleaner, thanks for the suggestion.

src/ngraph_builder.cc
```diff
@@ -1351,6 +1351,126 @@ static Status TranslateConv2DBackpropInputOp(
...
+  TF_RETURN_IF_ERROR(GetInputNodes(ng_op_map, op, &ng_input));
+
+  // Get the attributes
+  int64 block_size;
+  std::string tf_data_format;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "block_size", &block_size));
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "data_format", &tf_data_format));
+
+  ng::Shape input_shape = ng_input->get_shape();
+  std::map<std::string, int> format_to_int_map = {
```
shresthamalik: We can use if-else loop and avoid this map. Not really sure of the advantage of using this map

shresthamalik: I see you are using it again. :)

mingshan-wang: Hi @shresthamalik, I chose map without any particular reason. But I feel like it is more readable than `if-else`. In terms of performance, the switch statement could be the same as `if-else`. Are you concerned about this one more data structure? I can change to `if-else` as well.

src/ngraph_builder.cc
```diff
@@ -1351,6 +1351,126 @@ static Status TranslateConv2DBackpropInputOp(
...
+
+      ng_output_shape.push_back(input_shape[0]);
+      ng_output_shape.push_back(input_shape[channel_dimension] / num_blocks);
+      for (int i = 0; i < num_spatial_dimensions; i++) {
+        ng_output_shape.push_back(input_shape[i + 2] * block_size);
+      }
+      break;
+    }  // end of case NCHW
+  }
+
```
shresthamalik: As discussed, we can have the logic for calculating reshape, transpose, output shapes for only one of the data format NCHW (or NHWC), which we can assume as the standard format. If input has the other format, we reshape the input to the std format , do our computations and reshape it back.  Please try this.

We will have a simpler implementation and bug fixes (later if any) would be simpler to introduce. We only have to worry computation in one format.

mingshan-wang: Hi @shresthamalik @sayantan-nervana, I discussed with @avijit-nervana and @aprocter today, we agreed on that the each op in the bridge should honor its' input tensor's data format, so it is ok or maybe more clearer to keep the logic separate in the translation function. In addition, adding transpose/reshape at the beginning and the end for another format will introduce two additional operations, it might be not good in terms of performance concern.

What do you think? Feel free to input your thoughts.

src/ngraph_builder.cc
```diff
@@ -1351,6 +1351,177 @@ static Status TranslateConv2DBackpropInputOp(
...
+      ng_output_shape.push_back(input_shape[0]);
+      for (int i = 0; i < num_spatial_dimensions; i++) {
+        ng_output_shape.push_back(input_shape[i + 1] * block_size);
+      }
+      ng_output_shape.push_back(input_shape[channel_dimension] / num_blocks);
+      break;
+    }  // end of case NHWC
+    // NCHW
+    case 1: {
+      cout << "shouldn't enter" << endl;
```
shresthamalik: Please remove the cout

src/ngraph_builder.cc
shresthamalik: Ok, then we can keep this implementation to save the additional Reshape ops.

src/ngraph_builder.cc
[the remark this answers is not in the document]
sayantan-nervana: could you please add some comments though
suggestions:
// This switch-case computes 3 things: ng_reshape_shape, ng_transpose_shape, ng_output_shape
and some more if needed, for example the data formats of the above 3 things.

src/ngraph_builder.cc
sayantan-nervana: Or a more descriptive graph-log why one shouldn't enter

test/test_array_ops.cpp
```diff
@@ -257,6 +338,119 @@ TEST(ArrayOps, SizeOpDefault) {
...
+TEST(ArrayOps, SpaceToDepthNCHW) {
+  std::map<std::vector<int64>, int> input_map;
+  input_map.insert(pair<std::vector<int64>, int>({1, 1, 2, 2}, 2));
+  input_map.insert(pair<std::vector<int64>, int>({1, 10, 5, 5}, 5));
+  input_map.insert(pair<std::vector<int64>, int>({1, 20, 3, 3}, 3));
+  input_map.insert(pair<std::vector<int64>, int>({2, 3, 6, 3}, 3));
+  input_map.insert(pair<std::vector<int64>, int>({10, 10, 10, 10}, 2));
+  input_map.insert(pair<std::vector<int64>, int>({2, 1, 15, 3}, 3));
+  input_map.insert(pair<std::vector<int64>, int>({30, 30, 30, 30}, 10));
+
```
sayantan-nervana: thumbs up for the thorough unit testing 👍

mingshan-wang:
Hi @shresthamalik @sayantan-nervana @avijit-nervana Like we discussed this morning, for NGraph ops that doesn't require a specific format, we will keep separate logic in the bridge level for different data format. So I still use my previous implementation, and add some comments in the DepthToSpace implementation.

test/test_utilities.cpp
```diff
@@ -70,6 +70,11 @@ void PrintTensor(const Tensor& T1) {
   LOG(INFO) << "print tensor values" << T1.DebugString();
 }
 
+// Only displays values in tensor without shape information etc.
+void PrintTensorAllValues(const Tensor& T1, int64 max_entries) {
+  LOG(INFO) << "all tensor values" << T1.SummarizeValue(max_entries) << endl;
+}
+
```
sayantan-nervana: This is an useful function.
ideally this function would accept a default val of `max_entries`, set to -1
when it sees `-1`, it would compute the total num of elements in the tensor and display it.

But we dont have to do this in this PR

PR #261, opened 2018-10-16
Sindhu/Python tf_test_runner to test tensorflow python tests

sindhu-nervana:
Here is the usage of this script. Please try out and provide feedback. 

usage: tf_unittest_runner.py [-h] --tensorflow_path TENSORFLOW_PATH
                             [--list_tests LIST_TESTS] [--run_test RUN_TEST]

required arguments:
  --tensorflow_path TENSORFLOW_PATH
                        Specify the path where Tensorflow is installed.
                        Eg:/localdisk/skantama/tf-ngraph/tensorflow

optional arguments:
  -h, --help            show this help message and exit
  --list_tests LIST_TESTS
                        Prints the list of test cases in this package.
                        Eg:math_ops_test
  --run_test RUN_TEST   Runs the testcase and returns the output.
                        Eg:math_ops_test.math_ops_test.DivNoNanTest.testBasic


sindhu-nervana:
We do not have a list of available python test packages in a file yet, so use these bazel commands to query them. 
bazel query 'kind(".*_test rule", //tensorflow/python/...)'
bazel query 'kind(".*_test rule", //tensorflow/python/...)' --output package
bazel query 'kind(".*_test rule", //tensorflow/python/...)' --output label

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+    all_dirs(arguments.tensorflow_path)
+
+    if(arguments.list_tests):
+        list_tests()
+
+    if(arguments.run_test):
+        run_test()
+
+from fnmatch import fnmatch
+def all_dirs(dirname):
```
sayantan-nervana: `all_dirs `->`add_all_dirs_to_path`

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+loader = unittest.TestLoader()
+parser = argparse.ArgumentParser()
+optional = parser._action_groups.pop() 
+required = parser.add_argument_group('required arguments')
+required.add_argument('--tensorflow_path', help="Specify the path where Tensorflow is installed. Eg:/localdisk/skantama/tf-ngraph/tensorflow \n", required=True)
+optional.add_argument('--list_tests', help="Prints the list of test cases in this package. Eg:math_ops_test \n")
+optional.add_argument('--run_test', help="Runs the testcase and returns the output. Eg:math_ops_test.math_ops_test.DivNoNanTest.testBasic")
+parser._action_groups.append(optional)
+arguments = parser.parse_args()
+
```
sayantan-nervana: The whole arg parsing can be moved to `main()`

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+    dirname = arguments.tensorflow_path
+    pattern = "*_test.py"
+    for path, subdirs, files in os.walk(dirname):
+        for name in files:
+            if fnmatch(name, pattern):
+                sys.path.append(path)
+
+def list_tests():
+    test_module = arguments.list_tests
+    module = __import__(test_module)
```
sayantan-nervana: arguments is a global variable. once arg parsing is moved to main it will not be visible
here, you can have `test_module` as argument to list_tests

similar comment for `run_test` and other functions

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+optional.add_argument('--run_test', help="Runs the testcase and returns the output. Eg:math_ops_test.math_ops_test.DivNoNanTest.testBasic")
+parser._action_groups.append(optional)
+arguments = parser.parse_args()
+
+def main():
+    all_dirs(arguments.tensorflow_path)
+
+    if(arguments.list_tests):
+        list_tests()
+
```
sayantan-nervana: `list_tests()` -> `list_tests(arguments.list_tests)`

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+    alltests = []
+    for test_class in test_modules:
+        alltests.append( ([i.id() for i in test_class._tests]))
+    print ('\n'.join((sorted(sum(alltests, [])))))
+
+
+def run_test():
+    test_name = arguments.run_test
+    tests = loader.loadTestsFromName(test_name)
+    test_result = unittest.TextTestRunner(verbosity=2).run(tests)
```
shresthamalik: More like questions for my understanding:

1. Line(53) This seems to run the test_name. Then depending on what the test_result was we append the status (1, 0, -1) to tests_run and errors or failures. But it seems we don't use either of these lists later. May be i am missing something? 
2. Is test_name expected to be name of one test ?

sindhu-nervana: @shresthamalik 
1. I am only storing the test_result status right now but not using them yet, I am trying to use unittest.testresult class to get some more error and failure log info with test_case name and print in a proper format. Still working on this. For now i am just using lists to track pass, fail, errors.
2. For now, yes. test_name takes only one test_name.

sayantan-nervana: Lets also have `def run_test(... verbosity=2)` as an argument.
In which case:
`unittest.TextTestRunner(verbosity=verbosity).run(tests)`

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,71 @@
...
+    for path, subdirs, files in os.walk(dirname):
+        for name in files:
+            if fnmatch(name, pattern):
+                sys.path.append(path)
+
+def list_tests():
+    test_module = arguments.list_tests
+    module = __import__(test_module)
+    test_modules = loader.loadTestsFromModule(module)
+    alltests = []
```
sayantan-nervana: loader is global here. could be initialized inside function

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,68 @@
+import unittest
+import sys
+import argparse
+import os
+
+#Import ngraph in builds without ngraph embedded
```
avijit-nervana: Why is this step necessary? Currently the Python tests won't work with nGraph loaded separately - so we should not `import ngraph`. Please consider removing this block.

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,68 @@
...
+try:
+    import ngraph
+except:
+    pass
+
+def main():
```
avijit-nervana: Add comment block (using Python style guide) that describes the command line parameters (just brief description) and an example bazel query to generate the list of test targets.

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,83 @@
...
+    parser = argparse.ArgumentParser()
+    optional = parser._action_groups.pop() 
+    required = parser.add_argument_group('required arguments')
+    required.add_argument('--tensorflow_path', help="Specify the path where Tensorflow is installed. Eg:/localdisk/skantama/tf-ngraph/tensorflow \n", required=True)
+    optional.add_argument('--list_tests', help="Prints the list of test cases in this package. Eg:math_ops_test \n")
```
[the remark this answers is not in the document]
sindhu-nervana: I did try making tensorflow_path as positional using required=Ture, but it still shows up in optional. So used add_argument_group instead.

test/python/tf_unittest_runner.py
```diff
@@ -0,0 +1,108 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+# ==============================================================================
+import unittest
+import sys
+import argparse
+import os
+"""
+tf_unittest_runner is primarily used to test tensorflow python unit tests uisng ngraph
```
kanvi-nervana: spelling error "using"
and may be you can say run tensorflow python unit tests instead of test tensorflow

test/python/tf_unittest_runner.py
```diff
@@ -77,6 +70,16 @@ def all_dirs_to_path(dirname):
 
 
 def list_tests(test_module):
+    """
```
avijit-nervana: Good documentation. Please split the lines such that the width is max 80 columns. Also - reduce the tabs (in commends) to 4 spaces.

PR #262, opened 2018-10-17
Shrestha/Fix Bug in Rewrite For Tracking. Manifested in RESNET when enabled with XLA

shresthamalik:
In rewrite for tracking we do some checks on the nGraphVariable ops and when we are tracking it, replace with another NGraphVariable op. We were missing the incoming edges to the replaced node.

Added some logs in tf_deadness_analysis.cc, so that when it fails we know which edge it is failing on.


shresthamalik:
TEST NOW

src/ngraph_rewrite_for_tracking.cc
```diff
@@ -83,6 +85,14 @@ Status RewriteForTracking(Graph* graph) {
...
         replacement->set_assigned_device_name(node->assigned_device_name());
 
+        // Add edge from the input nodes (to the variable node (NGraphVariable))
+        // to the new replacement node (also of type NGraphVariable)
+        for (auto edge : node->in_edges()) {
+          NGRAPH_VLOG(4) << "Replacing: " << edge->DebugString();
+          graph->AddEdge(edge->src(), edge->src_output(), replacement,
+                         edge->dst_input());
+        }
+
```
sayantan-nervana: Should it be AddEdge or UpdateEdge? (depending on whether we want to retain the original edge)
in case of addedge we are not really replacing the edge, so the print line could be:
`NGRAPH_VLOG(4) << "Cloning: " << edge->DebugString();` instead of `Replacing`

src/tf_deadness_analysis.cc
```diff
@@ -427,6 +452,15 @@ bool DeadnessAnalysisImpl::HasInputsWithMismatchingDeadness(const Node& node) {
...
+    if (it != predicate_map_.end()) {
+      NGRAPH_VLOG(5) << "Cannot find predicate for Edge ";
+      NGRAPH_VLOG(5) << "Src " << edge->src()->name() << "["
+                     << edge->src()->type_string() << "]"
+                     << " ,Src Idx " << edge->src_output() << " DST "
+                     << edge->dst()->name() << "[" << edge->dst()->type_string()
+                     << "]"
+                     << " ,Dst Idx " << edge->dst_input();
+    }
     CHECK(it != predicate_map_.end());
```
sayantan-nervana: There are 3 instances of `predicate_map_.find` in the tf_deadness code. The 3rd one is here:
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/tf_deadness_analysis.cc#L460

These are great/useful prints, and help a lot in debugging. But you have added them in 2 of the 3 places for instance. In the spirit of DRY, and also to cover the 3rd instance, what if we have a function in DeadnessAnalysisImpl

```
appropriate_type findInMapOrFail(const TensorId& tensor_id){
    auto it = predicate_map_.find(tensor_id);
    if (it != predicate_map_.end()) {the_debug_prints_you_added;}
    CHECK(it != predicate_map_.end());   //This i suppose exits the program?
    return it->second;
}
```

src/ngraph_rewrite_for_tracking.cc
sayantan-nervana: NVM, above comment is invalid

src/tf_deadness_analysis.cc
shresthamalik: I did not add it here , as the check prints the TensorID. There is no way of getting Edge from the TensorId

shresthamalik: The changes here were just to have some logs , as we were seeing many crashes. We have a Ticket to Update this code. May be i can PR there, if it is required? It is repeated once .

sayantan-nervana: Ah, makes sense. In which case, plz ignore my above comment

PR #265, opened 2018-10-17
Kanvi/skip assert

kanvi-nervana:


test/python/test_skip_assert.py
```diff
@@ -0,0 +1,51 @@
...
+
+        def run_test(sess):
+            return sess.run(
+                a2, feed_dict={
+                    x: test_input,
+                    y: test_input,
+                    z: test_input
+                })
+
+        self.with_ngraph(run_test)
```
sayantan-nervana: maybe we want to do this with and without ngraph?
```
assert (self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
```

src/ngraph_skip_assert.cc
```diff
@@ -0,0 +1,51 @@
...
+// Main entry point for skip assert.
+//
+Status SkipAssert(Graph* graph) {
+  for (auto node : graph->op_nodes()) {
+    if (node->type_string() == "Assert") {
+      NGRAPH_VLOG(4) << "Checking: " << node->name();
+      for (auto edge : node->out_edges()) {
+        if (edge->IsControlEdge()) {
+          NGRAPH_VLOG(4) << "Control edge: " << node->name();
+          graph->RemoveControlEdge(edge);
```
sayantan-nervana: We have already printed `node->name()` in an earlier line.
`edge->DebugString()` might be a more informative print. 

Also we could remove `NGRAPH_VLOG(4) << "Control edge removed. ";` since `RemoveControlEdge` has no chance of failure.

src/ngraph_skip_assert.cc
```diff
@@ -0,0 +1,51 @@
...
+          NGRAPH_VLOG(4) << "Control edge removed. ";
+        }
+      }
+    }
+  }
+  return Status::OK();
+}
```
sayantan-nervana: There can be no failures of TF in the lines executed above, so this function will always return `Status::OK()`

So we don't necessarily need `TF_RETURN_IF_ERROR` in this line `TF_RETURN_IF_ERROR(SkipAssert(options.graph->get()));` in `ngraph_rewrite_pass.cc`.

But in keeping with the other functions in rewrite_pass which return meaningful statuses, we could keep this function as it is now (returning Status instead of void)

src/ngraph_skip_assert.cc
shresthamalik: Yes, we can change the log in line 39, to "Removing control edge" << edge->DebugString()

test/python/test_skip_assert.py
shresthamalik: We can write a test that fails with tf and works with nGraph (as we got rid of control edges). Would that be ideal?

sayantan-nervana: I think as a test, it is a good idea. But in a real life network, its not ideal.

src/ngraph_rewrite_pass.cc
```diff
@@ -150,6 +151,16 @@ class NGraphVariableCapturePass : public NGraphRewritePass {
       DumpGraphs(options, idx, "captured", "Graph With Variables Captured");
     }
 
+    // Skip "Assert" if specifically asked by the user
+    if (std::getenv("NGRAPH_TF_SKIP_ASSERT") != nullptr) {
+      TF_RETURN_IF_ERROR(SkipAssert(options.graph->get()));
+      // If requested, dump unmarked graphs without asserts
+      if (DumpCapturedGraphs()) {
```
shresthamalik: Should we use a new check DumpSkipAssertGraphs() (We can check for DumpAllGraphs() and NGRAPH_TF_SKIP_ASSERT, May not want another flag) like the other ones?

src/ngraph_rewrite_pass.cc
```diff
@@ -150,6 +151,16 @@ class NGraphVariableCapturePass : public NGraphRewritePass {
       DumpGraphs(options, idx, "captured", "Graph With Variables Captured");
     }
 
+    // Skip "Assert" if specifically asked by the user
+    if (std::getenv("NGRAPH_TF_SKIP_ASSERT") != nullptr) {
```
avijit-nervana: Need to change this to `NGRAPH_TF_DISABLE_ASSERTS`. Also add a VLOG(0) indicating that the model is running with asserts disabled.

src/ngraph_rewrite_pass.cc
avijit-nervana: @shresthamalik good feedback. We need to refactor the code to take care of these. Please create a JIRA ticket and we will address these.

src/ngraph_skip_assert.cc
```diff
@@ -0,0 +1,51 @@
+/*******************************************************************************
```
avijit-nervana: My suggestion is to use the term `disable_asserts` - so need to rename this file and function names accordingly.

test/python/test_skip_assert.py
```diff
@@ -0,0 +1,51 @@
+# ==============================================================================
```
avijit-nervana: Same comment re. name change to `disable_asserts`

test/python/test_skip_assert.py
kanvi-nervana: What would be the purpose for such a test?

src/ngraph_rewrite_pass.cc
avijit-nervana: @kanvi-nervana No

test/python/test_disable_assert.py
```diff
@@ -0,0 +1,52 @@
...
+        def run_test(sess):
+            return sess.run(
+                a2, feed_dict={
+                    x: test_input,
+                    y: test_input,
+                    z: test_input
+                })
+
+        assert (
+            self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
```
shresthamalik: This test wont run with the env variable in the CI, right?

PR #267, opened 2018-10-19
Shrestha/fix draw

shresthamalik:
Using AddEdge Instead of UpdateEdge. tf was crashing while updating a control edge. However if we add the control edge and removed the node being replaced. It does the same thing.


src/ngraph_capture_variables.cc
```diff
@@ -102,8 +104,8 @@ Status CaptureVariables(Graph* graph) {
 
         for (auto edge : edges) {
           NGRAPH_VLOG(4) << "Replacing: " << edge->DebugString();
-          graph->UpdateEdge(replacement, edge->src_output(), edge->dst(),
-                            edge->dst_input());
+          graph->AddEdge(replacement, edge->src_output(), edge->dst(),
+                         edge->dst_input());
         }
```
sayantan-nervana: The comments on UpdateEdge:
// Updates the input to a node.  The existing edge to `dst` is removed and an
  // edge from `new_src` to `dst` is created. The NodeDef associated with `dst`
  // is also updated.


The comments on AddEdge:
  // Adds an edge that connects the xth output of `source` to the yth input of
  // `dest` and returns it. Does not update dest's NodeDef.

The comment for RemoveNode
 // Removes a node from this graph, including all edges from or to it.
  // *node should not be accessed after calling this function.
  // REQUIRES: node->IsOp()

UpdateEdge = removeedge + addedde + nodedef modification (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/graph/graph.cc#L525)
So UpdateEdge crashes perhaps when trying to modify the nodedef? We do see that when adding _mark_for_clustering + deadness.

While we know that `RemoveNode` at the end will clear the edge, do we want to do `AddEdge+RemoveEdge`? UpdateEdge does that + the nodedef modification. Given that TF's API's might change (though i doubt such basic functions as RemoveNode might change), maybe lets futureproof by deleting the edge and not relying upon some other call to delete it?

shresthamalik: Added specific calls to RemoveEdge()

PR #268, opened 2018-10-19
Add libsdl1.2-dev to Dockerfile for Python3, to run A3C model

crlishka:
Ubuntu package libsdl1.2-dev is needed to run the A3C inference model, which @kimjanik is automating.


crlishka:
TESTNOW

PR #270, opened 2018-10-20
kanvi/move_ngraph_disable_flag

kanvi-nervana:
Move the NGRAPH_TF_DISABLE flag to appropriate location so that it takes affect right from the beginning of the execution.

### DIFF
--- a/src/ngraph_rewrite_pass.cc
+++ b/src/ngraph_rewrite_pass.cc
@@ -128,6 +128,13 @@ mutex NGraphRewritePass::s_serial_counter_mutex;
 class NGraphVariableCapturePass : public NGraphRewritePass {
  public:
   Status Run(const GraphOptimizationPassOptions& options) override {
+    // If NGRAPH_TF_DISABLE is set we will not do anything; all subsequent
+    // passes become a no-op.
+    //
+    if (std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
+      return Status::OK();
+    }
+
     // If we don't get a main graph, log that fact and bail.
     if (options.graph == nullptr) {
       NGRAPH_VLOG(0) << "NGraphVariableCapturePass: options.graph == nullptr";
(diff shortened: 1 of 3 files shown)


src/ngraph_rewrite_pass.cc
```diff
@@ -128,6 +128,13 @@ mutex NGraphRewritePass::s_serial_counter_mutex;
 class NGraphVariableCapturePass : public NGraphRewritePass {
  public:
   Status Run(const GraphOptimizationPassOptions& options) override {
+    // If NGRAPH_TF_DISABLE is set we will not do anything; all subsequent
+    // passes become a no-op.
+    //
+    if (std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
+      return Status::OK();
+    }
+
```
shresthamalik: Please move this check after the Precaptured graph is dumped. That way, we can look at the state of the graph through tf execution also. This would be useful for debugging purpose.

We can then run TF with NGRAPH disabled and see the Graphs after TF
- Placement as (precaptured)
- RFE as (unmarked)

src/ngraph_rewrite_pass.cc
```diff
@@ -187,6 +194,12 @@ class NGraphVariableCapturePass : public NGraphRewritePass {
 class NGraphEncapsulationPass : public NGraphRewritePass {
  public:
   Status Run(const GraphOptimizationPassOptions& options) override {
+    // If NGRAPH_TF_DISABLE is set we will not do anything; all subsequent
+    // passes become a no-op.
+    //
+    if (std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
+      return Status::OK();
+    }
```
shresthamalik: Please move this after umarked_graph is dumped.

src/ngraph_rewrite_pass.cc
kanvi-nervana: You mean with NGRAPH disabled, correct?

shresthamalik: yes, corrected.

PR #275, opened 2018-10-22
Implement set_backend

earhart:


earhart:
FWIW -- I don't have access to the Jenkins build to see what failed; if it's anything to do with the change, let me know?  :-)

src/ngraph_backend_manager.cc
```diff
@@ -0,0 +1,62 @@
...
+#include <iostream>
+
+namespace tensorflow {
+namespace ngraph_bridge {
+
+BackendManager* BackendManager::Instance() {
+  static BackendManager instance;
+  return &instance;
+}
+
```
sayantan-nervana: It seems that `BackendManager` is supposed to be a singleton class.

In which case, can we make default/copy/assignment constructors/functions `private`

src/ngraph_backend_manager.h
```diff
@@ -0,0 +1,65 @@
...
+  // Gets the manager's current backend.
+  //
+  // If the backend has not previously been loaded, the implementation
+  // will attempt to load it, using the configuration specified in the
+  // NGRAPH_TF_BACKEND environment variable, or using "CPU" as a
+  // fallback if NGRAPH_TF_BACKEND is not set.
+  //
+  // The second member of the tuple will be true iff the manager is
+  // using the CPU backend.
+  std::tuple<std::shared_ptr<ngraph::runtime::Backend>, bool> GetBackend();
```
sayantan-nervana: `SetBackendConfig` has a nice clean interface but `GetBackend` is slightly overloaded, since it attempts to also load a backend if nothing is present. Perhaps a more descriptive name for what it is doing now is`SetBackendFromEnvironment`. In which case to make things consistent, can we have `SetBackendConfig` also return a tuple that `GetBackend` is currently returning

 ideally perhaps: a pure setter (which tries to load from environment if config is not specified) + a pure getter (which throws errors if nothing is set) + pure boolean `isCPUBackend(bool* b)`.

src/ngraph_backend_manager.cc
```diff
@@ -0,0 +1,62 @@
...
+    is_cpu = config_ == "CPU";
+  } else {
+    is_cpu = config_.substr(0, colon) == "CPU";
+  }
+
+  return std::make_tuple(backend_, is_cpu);
+}
+
+void BackendManager::LoadBackendLocked(std::string config) {
+  backend_ = ngraph::runtime::Backend::create(config);
```
sayantan-nervana: The function is named `LoadBackendLocked` but the mutex lock is in `SetBackendConfig` and `GetBackend`. I suppose the mutex is there to guard this line `backend_ = ngraph::runtime::Backend::create(config);`. In which case, can we have the `lock_guard` in `LoadBackendLocked`

src/ngraph_encapsulate_op.cc
```diff
@@ -310,6 +291,12 @@ class NGraphEncapsulateOp : public OpKernel {
         << "NGraphEncapsulateOp::Compute got freshness tracker for cluster "
         << m_ngraph_cluster;
 
+    // Get the backend.
+    std::shared_ptr<ng::runtime::Backend> backend;
+    bool is_cpu_backend;
+    std::tie(backend, is_cpu_backend) =
+        BackendManager::Instance()->GetBackend();
+
```
sayantan-nervana: the `BackendManager::Instance()->GetBackend()` should be in a try-catch

src/ngraph_backend_manager.cc
earhart: It's there to ensure that `config_` is set atomically with `backend_`. `GetBackend` checks `backend_` before calling `SetBackendLocked`; it needs to hold the mutex, ideally without dropping it between checking `backend_` and setting `backend_`.

(The naming convention comes from the NT kernel and low-level runtime -- "*Locked" means, to the reader, "Make sure the appropriate lock is held at all calls to this function."  :-)  I'm happy to rename it if another convention would make it more clear.)

src/ngraph_backend_manager.cc
earhart: It's usually a singleton class, but I've often found that it hurts testability to bake that knowledge into the class itself, beyond providing a Myers singleton accessor.  (I've added a comment to make it more clear what's going on)

src/ngraph_backend_manager.cc
```diff
@@ -0,0 +1,62 @@
...
+    LoadBackendLocked(config);
+  }
+
+  auto colon = config_.find(':');
+  bool is_cpu;
+  if (colon == std::string::npos) {
+    is_cpu = config_ == "CPU";
+  } else {
+    is_cpu = config_.substr(0, colon) == "CPU";
+  }
```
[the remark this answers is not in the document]
earhart: Done

src/ngraph_encapsulate_op.cc
earhart: Done

src/ngraph_backend_manager.h
earhart: I think the caller needs/wants a function somewhere whose functionality is, "Get me the right nGraph backend to use, and I don't care what you have to do in order to get it; I don't care whether you already have one, or if you've been given a config string, or if you're going to check an environment variable, or pull in something from dconf, or check the Registry, or whatever you might want to do on MacOS -- just get me something I can use, and throw an exception if you can't get it.  Oh, and I need to know whether it's the CPU backend."

So that's what `GetBackend()` is doing (and why it's not `backend() const`.)

I did think about splitting this up: we'd have a `BackendState` object, with a Meyers singleton instance accessor.  The state object would just be a struct with the mutex, config string, and current backend.  We could then have free functions that would manipulate the state object, basically exactly as they are now.

The downside is that it depends on all callers getting the manipulations correct.  It seems more obviously-correct to me to specialize the state object a bit for the current callers, keeping the internals private and exposing a couple of accessor APIs that are easy to validate.  But I agree, the resulting class is very specialized for its current role in nGraph-TF; it's not very OO, it's not very reusable, it's just a little bit of glue logic for this one particular situation.

Overall... I'm pretty comfortable either way.  My on-the-fly "which way should we go" landed on the specialized object implementation, but if you'd prefer a more OO approach, I'm happy to write it up that way, too.

(FWIW -- things calling `SetBackendConfig` really don't care about actually getting the current backend.)

src/ngraph_backend_manager.cc
```diff
@@ -0,0 +1,66 @@
...
+
+namespace tensorflow {
+namespace ngraph_bridge {
+
+BackendManager* BackendManager::Instance() {
+  static BackendManager instance;
+  return &instance;
+}
+
+void BackendManager::SetBackendConfig(const std::string& config) {
```
fengleitian: nit: this `SetBackendConfig` not only does Set.

src/ngraph_backend_manager.cc
sayantan-nervana: right. I realized after typing my comment that `std::getenv("NGRAPH_TF_BACKEND");` could cause races (thread1 reads backend = CPU, the env var changes to GPU, thread2 reads GPU. thread2 sets to GPU, thread 1 resets to CPU. Even though it should have been GPU it got set to CPU) after I wrote the above comment, but forgot to remove it from the review.

Interesting new info about the `*Locked` naming convention in NT. Since we do not have any such convention going, am happy with the name.

src/ngraph_backend_manager.h
sayantan-nervana: I do see your point of `Get me the right nGraph backend to use, and I don't care what you have to do in order to get it`. I am good either way. I'll let @avijit-nervana comment.

src/ngraph_backend_manager.h
```diff
@@ -0,0 +1,67 @@
...
+namespace tensorflow {
+namespace ngraph_bridge {
+
+// Manages the nGraph bridge's notion of the current execution backend.
+class BackendManager {
+ public:
+  // Returns the global BackendManager instance.
+  //
+  // (Note that in testing scenarios, there may be additional instances.)
+  static BackendManager* Instance();
```
avijit-nervana: This is a factory method - so need to delete all the other constructors or mark private with test class declared as friend if needed for testing purposes.

src/ngraph_backend_manager.h
```diff
@@ -0,0 +1,67 @@
...
+  // Sets the manager's current backend.
+  //
+  // Throws an exception if the supplied configuration string can't be
+  // resolved to an nGraph backend -- either if it doesn't match an
+  // available backend, or if the loaded backend is unable to process
+  // the backend-specific configuration component.
+  //
+  // If the configuration cannot be resolved, the manager's current
+  // backend will remain unchanged.
+  void SetBackendConfig(const std::string& config);
```
avijit-nervana: I don't see the need for a set config method. The purpose of the backend manager is to create a backend for a given configuration. So the GetBackend should use a configuration and return an appropriate backend. The issue with `SetConfig` and `GetObject`is that there is no way to guarantee the atomicity.

src/ngraph_backend_manager.h
avijit-nervana: In my opinion, it's better to keep functions simple and do what the name suggests. If the purpose of the function is to get a backend as requested by the caller - then that's what it needs to return. So - need to add the config parameter here and then we can have another method called GetDefaultBackend() or GetCpuBackend().

earhart: But then we'd need some place to store that config, and then _that_ becomes the global object... :-)

Okay -- I'm thinking it will be more clear what's going on if I break apart the data storage and the manipulation logic.  I'm in the middle of some other changes, but should have an update to this within the next day or two.

earhart:
Okay -- I've separated the backend state manipulation logic from the backend state data; thoughts?

PR #278, opened 2018-10-24
yeonsily/Extend the model test script

yeonsily:
- Get multiple output tensor layers
  - Change input numbers between 0-255
  - Load both pb and pbtxt type graph

### DIFF
--- a/diagnostics/model_test/verify_model.py
+++ b/diagnostics/model_test/verify_model.py
@@ -2,21 +2,44 @@ import tensorflow as tf
 import argparse
 import numpy as np
 import ngraph
+from google.protobuf import text_format
 import json
 import os
 
+def createFolder(directory):
+    try:
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+    except OSError:
+        print ('Error: Creating directory. ' +  directory)
+    os.chdir(output_folder)
+
+def set_os_env(select_device):
+    if select_device == 'CPU':
+        # run on TF only
+        ngraph.disable()
+        os.environ['NGRAPH_TF_BACKEND']='CPU'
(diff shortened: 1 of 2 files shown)


diagnostics/model_test/verify_model.py
```diff
@@ -123,54 +140,95 @@ def parse_json():
...
     input_tensor_dim_map = {}
     for (dim, name) in zip(input_dimension, input_tensor_name):
-        random_input = np.random.random_sample([bs] + dim)
+        #random_input = np.random.random_sample([bs] + dim)
+        random_input = np.random.randint(255, size=[bs] + dim).astype('float32')
         input_tensor_dim_map[name] = random_input
 
-    # Run the model on tensorflow
+<<<<<<< HEAD
+    # Run the model on device1
```
sayantan-nervana: merge artifact `<<<<<<< HEAD`

diagnostics/model_test/verify_model.py
```diff
@@ -123,54 +140,95 @@ def parse_json():
...
         input_tensor_name
     ), "input_tensor_name dimension should match input_dimension in json file"
 
     # Matches the input tensors name with its required dimensions
     input_tensor_dim_map = {}
     for (dim, name) in zip(input_dimension, input_tensor_name):
-        random_input = np.random.random_sample([bs] + dim)
+        #random_input = np.random.random_sample([bs] + dim)
+        random_input = np.random.randint(255, size=[bs] + dim).astype('float32')
         input_tensor_dim_map[name] = random_input
```
sayantan-nervana: need to specify this in json. we don't want to hard code the value range here

PR #280, opened 2018-10-24
Yeonsily/l2l test

yeonsily:
Modify the model test script
      - Change input numbers between 0-255
      - Load both pb and pbtxt type graph
      - Set back_end as parameter
      - Save test result and tensor array to .npy file

### DIFF
--- a/diagnostics/model_test/mnist_cnn.json
+++ b/diagnostics/model_test/mnist_cnn.json
@@ -1,8 +1,12 @@
 {
     "model_name": "mnist_cnn",
+    "device_name1": "CPU",
+    "device_name2": "NGRAPH_CPU",
+    "quantized_mode": "FLOAT",
     "frozen_graph_location": "/nfs/site/home/mingshan/frozen_graphs/frozen_mnist.pb",
     "input_tensor_name": ["import/Placeholder:0"],
     "output_tensor_name": ["import/Placeholder:0", "import/fc2/yconv:0"],
+    "log_file_name": "output.log",
     "l1_norm_threshold": 0.01,
     "l2_norm_threshold": 0.01,
     "inf_norm_threshold": 0.01,
--- a/diagnostics/model_test/verify_model.py
+++ b/diagnostics/model_test/verify_model.py
@@ -6,31 +6,34 @@ from google.protobuf import text_format
 import json
 import os
 
+
 def createFolder(directory):
     try:
         if not os.path.exists(directory):
             os.makedirs(directory)
     except OSError:
-        print ('Error: Creating directory. ' +  directory)
+        print('Error: Creating directory. ' + directory)
     os.chdir(output_folder)
+
 
 def set_os_env(select_device):
     if select_device == 'CPU':
         # run on TF only
         ngraph.disable()
-        os.environ['NGRAPH_TF_BACKEND']='CPU'
+        os.environ['NGRAPH_TF_BACKEND'] = 'CPU'
     else:
         if not ngraph.is_enabled():
             ngraph.enable()
 
         back_end = select_device.split("NGRAPH_")
-        os.environ['NGRAPH_TF_BACKEND']=back_end[1]
+        os.environ['NGRAPH_TF_BACKEND'] = back_end[1]
 
         if quantized_mode == 'QUANTIZED':
-            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH']='1'
+            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = '1'
         else:
             # float mode
-            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH']='0'
+            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = '0'
+
 
 def calculate_output(param_dict, select_device, input_example):
     """Calculate the output of the imported frozen graph given the input.
@@ -162,11 +165,12 @@ if __name__ == '__main__':
 
     # Get log file name to save output
     log_file = parameters["log_file_name"]
-    output_folder = device1+"-"+device2+"-"+quantized_mode
+    output_folder = device1 + "-" + device2 + "-" + quantized_mode
     createFolder(output_folder)
     file = open(log_file, "w")
     file.write("Model name: {}\n".format(parameters["model_name"]))
-    file.write("L1/L2/Inf norm configuration: {}, {}, {}\n".format(l1_norm_threshold, l2_norm_threshold, inf_norm_threshold))
+    file.write("L1/L2/Inf norm configuration: {}, {}, {}\n".format(
+        l1_norm_threshold, l2_norm_threshold, inf_norm_threshold))
 
     # Generate random input based on input_dimension
     np.random.seed(100)
@@ -197,11 +201,11 @@ if __name__ == '__main__':
             out_tensor_names_cpu, result_ngraph_arrs, result_tf_graph_arrs):
         file.write(">>>Start {}\n".format(tname))
 
-        new_out_layer = tname.replace("/","_")
+        new_out_layer = tname.replace("/", "_")
         nparray_tf = np.array(result_tf_graph)
         nparray_ngraph = np.array(result_ngraph)
-        np.save(device1+"-"+new_out_layer+".npy", nparray_tf)
-        np.save(device2+"-"+new_out_layer+".npy", nparray_ngraph)
+        np.save(device1 + "-" + new_out_layer + ".npy", nparray_tf)
+        np.save(device2 + "-" + new_out_layer + ".npy", nparray_ngraph)
 
         l1_norm = calculate_norm(result_ngraph, result_tf_graph, 1)
         l2_norm = calculate_norm(result_ngraph, result_tf_graph, 2)
@@ -210,8 +214,9 @@ if __name__ == '__main__':
         if l1_norm > l1_norm_threshold:
             print("The L1 norm %f is greater than the threshold %f for %s" %
                   (l1_norm, l1_norm_threshold, tname))
-            file.write("L1 norm test - Fail: The L1 norm %f is greater than the threshold %f\n" %
-                  (l1_norm, l1_norm_threshold))
+            file.write(
+                "L1 norm test - Fail: The L1 norm %f is greater than the threshold %f\n"
+                % (l1_norm, l1_norm_threshold))
         else:
             print("L1 norm test passed for ", tname)
             file.write("L1 norm test - Pass\n")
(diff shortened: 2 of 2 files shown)


diagnostics/model_test/verify_model.py
```diff
@@ -2,10 +2,39 @@
...
 
 
+def createFolder(directory):
+    try:
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+    except OSError:
+        print('Error: Creating directory. ' + directory)
+    os.chdir(output_folder)
+
```
sayantan-nervana: did u mean `directory` instead of `output_folder`

yeonsily: Yes, you're right. Actually let me move that line outside of createFolder().
It will be like this.
   # Get log file name to save output
    log_file = parameters["log_file_name"]
    output_folder = device1+"-"+device2+"-"+quantized_mode
    createFolder(output_folder)
    **os.chdir(output_folder)**

diagnostics/model_test/mnist_cnn.json
```diff
@@ -1,8 +1,13 @@
 {
     "model_name": "mnist_cnn",
+    "device_name1": "CPU",
+    "device_name2": "NGRAPH_CPU",
```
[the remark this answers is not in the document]
yeonsily: device_name1 could not always tensorflow_backend. There's a case it also has ngrpah_backend. For example, comparing ngraph_CPU vs ngraph_NNPI.

diagnostics/model_test/mnist_cnn.json
```diff
@@ -1,8 +1,13 @@
...
+    "quantized_mode": "FLOAT",
     "frozen_graph_location": "/nfs/site/home/mingshan/frozen_graphs/frozen_mnist.pb",
     "input_tensor_name": ["import/Placeholder:0"],
     "output_tensor_name": ["import/Placeholder:0", "import/fc2/yconv:0"],
+    "log_file_name": "output.log",
+    "random_val": 255,
```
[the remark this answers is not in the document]
yeonsily: Ok. Will update it.

diagnostics/model_test/mnist_cnn.json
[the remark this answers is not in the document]
mingshan-wang: I see. Ok, that makes sense. Essentially, this script is trying to compare one testing backend result with a reference(correct) backend result.  

Then how about `reference_backend`, meaning the one that we want to compare the result with, and then `testing_backend`? Or other names you thinks fit.

diagnostics/model_test/verify_model.py
```diff
@@ -133,44 +182,68 @@ def parse_json():
...
         if inf_norm > inf_norm_threshold:
             print("The inf norm %f is greater than the threshold %f for %s" %
                   (inf_norm, inf_norm_threshold, tname))
+            file.write(
+                "Inf norm test - Fail: The inf norm %f is greater than the threshold %f\n"
+                % (inf_norm, inf_norm_threshold))
         else:
             print("inf norm test passed for ", tname)
+            file.write("Inf norm test - Pass\n")
+
```
sayantan-nervana: about file.write: It seems it is only writing whatever is being printed to a file. Unless one wants to dump a lot more extra information, we do not need the file.write, as we can just pipe the output to a log file.

my suggestion is that if u need some small info to be printed, u can add a new print line. if u want large chunks to be output, only then we can use the log file. For numbers dump we are doing the .npy output in any case, so perhaps the log file is redundant

diagnostics/model_test/verify_model.py
```diff
@@ -2,10 +2,38 @@
...
+    except OSError:
+        print('Error: Creating directory. ' + directory)
+
+
+def set_os_env(select_device):
+    if select_device == 'CPU':
+        # run on TF only
+        ngraph.disable()
+        os.environ['NGRAPH_TF_BACKEND'] = 'CPU'
+    else:
```
sayantan-nervana: I think we don't need `os.environ['NGRAPH_TF_BACKEND'] = 'CPU'` when ngraph is disabled

diagnostics/model_test/verify_model.py
```diff
@@ -2,10 +2,38 @@
...
+        if not ngraph.is_enabled():
+            ngraph.enable()
+
```
sayantan-nervana:
```suggestion
assert select_device[:7]=="NGRAPH_", "Expecting device name to start with NGRAPH_"
```

diagnostics/model_test/verify_model.py
```diff
@@ -2,10 +2,38 @@
...
+        back_end = select_device.split("NGRAPH_")
+        os.environ['NGRAPH_TF_BACKEND'] = back_end[1]
+
+        if quantized_mode == 'QUANTIZED':
+            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = '1'
+        else:
+            # float mode
+            os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = '0'
+
```
sayantan-nervana:
```suggestion
os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = '1' if quantized_mode == 'QUANTIZED' else '0'
```

Also why have quantized_mode be a string. if its a number then we can simply do:
`os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = str(quantized_mode)`

diagnostics/model_test/verify_model.py
yeonsily: I agree that the output file is redundant. The idea was if we run many layers then the log screen will scroll up but if the log is saved to a file then it's easy to check whenever we need later. I think it's still ok to have but if this is against code design policy, I'll remove them.

diagnostics/model_test/verify_model.py
yeonsily: Will remove it.

diagnostics/model_test/verify_model.py
yeonsily: You're right. Let me update json file also to get quantized mode as a number.

diagnostics/model_test/verify_model.py
sayantan-nervana: We can always pipe the output to a file
```
python verify_model.py --json_file minst.json > log.txt
```

mingshan-wang:
Hi @yeonsily, since this change added a lot of functionalities to the model_test part, maybe considering update the README under the model_test/ folder. This way the user will not be confused when they use the testing script. What do you think?

yeonsily:
@mingshan-wang Ok, let me update README too.

diagnostics/model_test/README.md
```diff
@@ -1,21 +1,24 @@
-# Compare model output between Tensorflow and NGraph
+# Compare model output between two different backends
 
-### This model_test tool will run the model inference seperately on TF and NGraph, and the desired output from TF and NGraph should match given the same inputs. It can be used as a debugging tool, and also a verification that NGraph produces the same output as Tensorflow. 
+### This model_test tool will run the model inference seperately on two specified backends(e.g.Tensorflow and nGraph) in json file and the desired outputs from each backend should match given the same inputs. It can be used as a debugging tool for layer by layer comparison, and also a verification that nGraph produces the same output as Tensorflow. 
 
 # Required files to use the tool:
-* A json file: Provide model specific parameters. Look at the example ```mnist_cnn.json```. You can start with the ```template.json``` and modify it to match your model
+* A json file: Provide model specific parameters. Look at the example ```mnist_cnn.json```.
 * A tensorflow frozen graph: A model frozen graph(.pb file) with trained weights and model architecture
```
mingshan-wang: With your change, we can also have a model .pbtxt file without the weights right? So user can supply either a tensorflow frozen graph, or a tensorflow model .pbtxt file.

diagnostics/model_test/README.md
```diff
@@ -1,21 +1,24 @@
...
+multiple input tensors and output tensors. Put the input tensor names as a list in the ```input_tensor_name``` field of the json file, and the output tensor name as a list in the ```output_tensor_name``` field of the json file. If no outputs are specified in the ```output_tensor_name```, then it will compare all output tensors
 * You will need the input dimensions for all the input tensors provided. Put the dimensions information as a list in the ```input_dimension``` field of the json file, and the corresponding order of ```input_tensor_name``` list should match the ```input_dimension``` list. Therfore, the length of ```input_tensor_name``` list should match the length of ```input_dimension``` list
 * Specify the the location of the frozen graph in the ```frozen_graph_location``` of the json file
 * Specify the ```batch_size``` field in the json file to the desired batch size for inference
 * Specify the tolerance between the TF and NGraph outputs at ```l1_norm_threshold```, ```l2_norm_threshold``` and ```inf_norm_threshold``` in the json file 
+* Specify the ```random_val_range``` used for the test
```
mingshan-wang: Maybe change `used for the test` to
`used to generate the input within 0 to random_val_range`

sayantan-nervana: @mingshan-wang 
Yes, for now it is ok to specify only the max as an input, but generally we can support min and max as inputs, which we can do in a separate PR. We can provide defaults of 0 and 1 for min and max if nothing is specified

diagnostics/model_test/verify_model.py
```diff
@@ -2,14 +2,39 @@
...
+    else:
+        if not ngraph.is_enabled():
+            ngraph.enable()
+
+        assert select_device[:
+                             7] == "NGRAPH_", "Expecting device name to start with NGRAPH_"
+        back_end = select_device.split("NGRAPH_")
+        os.environ['NGRAPH_TF_BACKEND'] = back_end[1]
+        # quantized mode:1 / float mode:0
+        os.environ['NNPI_CONVERT_FLOAT_TO_QUANT_GRAPH'] = str(quantized_mode)
```
jiminha: I'm wondering if this should be removed, but we setup something specifically, Not sure if this is right place to have NNPI specific changes go in.

sayantan-nervana: Good point

yeonsily: Ok. make sense. Let me remove all the code related quantization.

diagnostics/model_test/README.md
```diff
@@ -1,21 +1,23 @@
-# Compare model output between Tensorflow and NGraph
+# Compare model output between two different backends
 
-### This model_test tool will run the model inference seperately on TF and NGraph, and the desired output from TF and NGraph should match given the same inputs. It can be used as a debugging tool, and also a verification that NGraph produces the same output as Tensorflow. 
+### This model_test tool will run the model inference seperately on two specified backends(e.g.Tensorflow and nGraph) in json file and the desired outputs from each backend should match given the same inputs. It can be used as a debugging tool for layer by layer comparison, and also a verification that nGraph produces the same output as Tensorflow.
```
shresthamalik: May be this will be more informative?
```
any two specified backends (e.g. Tensorflow (without nGraph), nGraph-CPU , nGraph-GPU, etc.)
```

yeonsily: Will update it for next patch.

PR #282, opened 2018-10-26
Sindhu/tf unittest regexsupport

sindhu-nervana:


sindhu-nervana:
@sayantan-nervana Moved the file for review. Please check now

test/python/tf_unittest_runner.py
```diff
@@ -39,89 +42,147 @@ def main():
...
     if (arguments.list_tests):
-        list_tests(arguments.list_tests)
+        try:
+            module_list = regex_walk(arguments.tensorflow_path,
+                                     arguments.list_tests)
+            list_tests(module_list, arguments.list_tests)
+        except:
+            print(
+                "Enter a valid argument to LIST tests.\nList of accepted formats:"
+            )
```
sayantan-nervana: `LIST tests`-->`--list_tests`

sayantan-nervana: So there is a try-except here because?
which of the function calls `regex_walk` or `list_tests` is expected to fail?

test/python/tf_unittest_runner.py
```diff
@@ -39,89 +42,147 @@ def main():
...
+        try:
+            module_list = regex_walk(arguments.tensorflow_path,
+                                     arguments.run_test)
+            test_list = list_tests(module_list, arguments.run_test)
+            run_test(test_list)
+        except:
+            print(
+                "Enter a valid argument to RUN tests.\nList of accepted formats:"
+            )
+            print('\n'.join(accepted_formats))
```
sayantan-nervana: `RUN tests.` --> `--run_tests`

sayantan-nervana: There is not much difference between the 2 modes, except `run_test(test_list)`

how about this refactor:

```
def helper(tf_path, test_regex):
    try:
            module_list = regex_walk(arguments.tensorflow_path,
                                     arguments.run_test)
            test_list = list_tests(module_list, arguments.run_test)
   except:
        pass #stuff here

if list:
    test_list = helper(arguments.tensorflow_path, arguments.run_test)
    print_tests(test_list)
if run:
    test_list = helper(arguments.tensorflow_path, arguments. list_tests)
   run_test(test_list)
```

of course, in this refactor, `list_tests` would not do the printing of functions

test/python/tf_unittest_runner.py
```diff
@@ -39,89 +42,145 @@ def main():
...
+        "math_ops_test.DivNoNanTest.testBasic", "math_ops_test.DivNoNanTest.*",
+        "math_ops_test.D*", "math_ops_test.*", "math_*_test", "math_*_*_test",
+        "math*_test"
+    ]
+    try:
+        module_list = regex_walk(tf_path, test_regex)
+    except:
+        module_list = []
+        print(
+            "Invalid module name.Use bazel query to get list of tensorflow python test modules"
```
sayantan-nervana: `name.Use` -> `name. Use`

`Use bazel query` ... maybe we can print the exact command, if its easy enough to print (i dont know the command, so not sure if its a small or a large thing)

test/python/tf_unittest_runner.py
```diff
@@ -39,89 +42,145 @@ def main():
...
-
+        test_list = helper(arguments.tensorflow_path, arguments.list_tests)
+        print('\n'.join(test_list))
     if (arguments.run_test):
-        run_test(arguments.run_test)
+        test_list = helper(arguments.tensorflow_path, arguments.run_test)
+        run_test(test_list)
+
+
+def helper(tf_path, test_regex):
```
sayantan-nervana: maybe a more descriptive name than `helper`. `get_atomic_test_list`?

sindhu-nervana: @sayantan-nervana I actually changed the name to get_test_list but did not go in previous commit. Updated now, please review.

test/python/tf_unittest_runner.py
```diff
@@ -39,89 +42,147 @@ def main():
...
-        sys.exit()
-    elif test_result.errors:
-        errors.append([test_name, test_result.errors])
-        tests_run.append([testsRun, 0])
-    elif test_result.failures:
-        failures.append([test_name, test_result.failures])
-        tests_run.append([testsRun, -1])
+    for test in test_list:
+        tests = loader.loadTestsFromName(test)
+        test_result = unittest.TextTestRunner(verbosity=verbosity).run(tests)
```
sayantan-nervana: we do not do anything with the results?
We should print the test name and the result at least right?

sindhu-nervana: I am working on another PR to print the results and write the output to an XML file. So you will have test name, status and log.

PR #284, opened 2018-10-26
Overhaul docker system to support many image types, add image for gcc48 on ubuntu1604

crlishka:
This PR does the following:

1. Overhauls the docker image system to handle many Dockerfiles, split by purpose (builds vs inference testing vs Klocwork scans), python version, gcc version, and whatever else we think of in the future.
2. Adds a Dockerfile for building with gcc 4.8 on Ubuntu 16.04, using the same build script as Ubuntu 16.04 with native gcc.
3. Maintains compatibility with the old Dockerfiles that are still in use in some Jenkins automation.
4. Adds initial **incomplete** scripts for automating Klocwork scans, folding in some work that was done in an earlier (and older).  This is being used to do manual Klocwork scans in a docker image.  The automation work will be completed later.
5. Adds the **basis** for providing commands for developers to run a docker container with /localdisk, although developer-friendly commands _have not yet been written_.
6. Removes the obsolete "MKLDNN" build.

The new Docker image system and gcc 4.8 builds are functional now and necessary for the upcoming Beta Release.  This PR is being merged now so that these features are ready before Beta Release.

More work will be done later on developer-friendly commands to run docker containers, and to migrate off the old Dockerfiles.


crlishka:
TESTNOW

PR #285, opened 2018-10-27
Shrestha/Cluster Nodes According to Matching Deadness Predicate

shresthamalik:


shresthamalik:
TESTNOW

src/ngraph_assign_clusters.cc
```diff
@@ -195,20 +240,95 @@ Status AssignClusters(Graph* graph) {
...
+      if (src_predicate != TRUE_PRED_STRING &&
+          dst_predicate != TRUE_PRED_STRING && src_predicate != dst_predicate) {
+        continue;
+      }
+
+      // Case src X , dst True // invalid scenario
+      if (src_predicate != TRUE_PRED_STRING &&
+          dst_predicate == TRUE_PRED_STRING) {
+        return errors::Internal("Attempting to cluster control-flow node ",
+                                dst->name(), "[", dst->type_string(), "]");
```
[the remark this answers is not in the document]
shresthamalik: Yes, src is a non-true predicate and dst is true_predicate, which is possible only when dst is a control-flow op.

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,56 @@ namespace {
...
+    deadness_analyzer->GetNodePredicate(*node, pred_string);
+    nodes_predicate_map[node] = pred_string;
+
     cluster_map[node]->index = new_index;
     cluster_map[node]->nodes.insert(node);
+    cluster_map[node]->predicate_string = pred_string;
+    std::set<const Edge*> temp(node->out_edges().begin(),
+                               node->out_edges().end());
+
+    cluster_map[node]->outgoing_edges = temp;
```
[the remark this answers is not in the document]
shresthamalik: I had tried few other ways to get rid of temp (but not this). This works Thanks! 👍

src/ngraph_assign_clusters.cc
```diff
@@ -195,20 +240,95 @@ Status AssignClusters(Graph* graph) {
...
+        // Merge dst cluster into src cluster
+        std::string cluster_pred =
+            (src_predicate != TRUE_PRED_STRING) ? src_predicate : dst_predicate;
+        cluster_map[src]->predicate_string = cluster_pred;
+
+        auto cluster_dst = cluster_map[dst];
+        // Update outgoing edges of the merged cluster
+        for (auto cluster_dst_out_edge : cluster_dst->outgoing_edges) {
+          cluster_map[src]->outgoing_edges.insert(cluster_dst_out_edge);
+        }
```
sayantan-nervana: this I believe is an union right?
I think it can be done without looping by passing in a range: `cluster_map[src]->outgoing_edges.insert(cluster_map[dst]->outgoing_edges.begin(), cluster_map[dst]->outgoing_edges.end())`

(but i m not sure, only the compiler (and unit tests) can tell)

test/graph_rewrites/deadness_test.cc
```diff
@@ -134,6 +136,305 @@ TEST(DeadnessCheck, livedead1TF) {
...
+  ASSERT_EQ(N3_Sub_cluster, N4_Mul_cluster);
+  // B, N5 and N6 are in same cluster
+  ASSERT_EQ(N5_Add_cluster, N6_Mul_cluster);
+  ASSERT_EQ(N5_Add_cluster, B_cluster);
+
+  // N1 and N3 are in different cluster
+  ASSERT_NE(N1_Add_cluster, N3_Sub_cluster);
+  // N1 and N5 are in different cluster
+  ASSERT_NE(N1_Add_cluster, N5_Add_cluster);
+  // N3 and N5 are in differenct cluster
```
sayantan-nervana: `differenct`-> `different`

src/ngraph_assign_clusters.cc
```diff
@@ -245,37 +446,37 @@ Status AssignClusters(Graph* graph) {
...
-        }
-
-        if (!NodeIsMarkedForClustering(node)) {
-          return errors::Internal("Node ", node->DebugString(),
-                                  " was not marked for clustering but was "
-                                  "placed in an nGraph cluster.");
-        }
+    for (auto node : cluster->nodes) {
+      if (NGRAPH_VLOG_IS_ON(5)) {
+        NGRAPH_VLOG(5) << ">> cluster " << cluster_idx << ": " << node->id()
```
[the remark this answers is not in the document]
shresthamalik: No, it is by design. The logs stand out that way.

src/ngraph_mark_for_clustering.cc
```diff
@@ -62,18 +61,6 @@ static Status NGraphPlacementRequested(Node* node, bool& placement_ok) {
   return Status::OK();
 }
 
-#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
-// Checks if the node's inputs have mismatching deadness
-static Status DeadnessOk(Node* node,
-                         std::unique_ptr<DeadnessAnalysis>* deadness_analyzer,
-                         bool& deadness_ok) {
-  deadness_ok =
-      !(node->IsMerge() ||
```
kanvi-nervana: I know this is removed, but just curious as to why we were only checking for IsMerge() and not IsSwitch etc.?

shresthamalik: This is a requirement of HasInputsWithMismatchingDeadness() api. It is designed to take only non-merge ops.  It fails if the node is a merge op.

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+
+  std::string cluster_pred = GetMergedClusterPred(src_predicate, dst_predicate);
+
+  cluster_map[src]->predicate_string = cluster_pred;
+  // Update outgoing edges of the merged cluster
+  cluster_map[src]->outgoing_edges.insert(
+      cluster_map[dst]->outgoing_edges.begin(),
+      cluster_map[dst]->outgoing_edges.end());
+  cluster_map[src]->outgoing_edges.erase(edge);
+#endif
```
sayantan-nervana: maybe you have already done this: is it possible to run a small test/network to make sure that if NGRAPH_TF_DISABLE_DEADNESS_CHECK is set, then all the macros work as they should and we do not get compile errors or runtime errors.

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+  NGRAPH_VLOG(5) << "Contracting: " << src->name() << "[" << src->type_string()
+                 << " , " << edge->src_output() << "]@" << src_index << " -> "
+                 << dst->name() << "[" << dst->type_string() << " , "
+                 << edge->dst_input() << "]@" << dst_index;
+
+#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
+  string src_predicate = cluster_map[src]->predicate_string;
+  string dst_predicate = cluster_map[dst]->predicate_string;
+  NGRAPH_VLOG(5) << "Src pred: " << src_predicate
+                 << " ,Dst pred: " << dst_predicate;
```
sayantan-nervana: `" ,Dst"` --> `", Dst"`

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+
+  std::string cluster_pred = GetMergedClusterPred(src_predicate, dst_predicate);
```
sayantan-nervana: you could also do the `DeadnessAnalysis::IsTruePredString(src_predicate) || DeadnessAnalysis::IsTruePredString(dst_predicate)` assert before calling the function here. In which case this function would return TF status and not void

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
+  std::string predicate_string;
+  std::set<const Edge*> outgoing_edges;
+#endif
 };
+
+#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
+// Returns the predicate of the merged cluster
+// If Src Predicate is TRUE then merged cluster gets the dst predicate
+string GetMergedClusterPred(string& src_predicate, string& dst_predicate) {
```
sayantan-nervana: could be inlined perhaps

src/tf_deadness_analysis.h
```diff
@@ -71,9 +87,40 @@ class DeadnessAnalysis {
...
+  // Predicate of all its input predicates) and we can attach a predicate string
+  // to the data-flow node (predicate of its output edge). Control flow ops are
+  // assigned a placeholder predicate string (CONTROL_FLOW_PRED_STRING) .
+  virtual Status GetNodePredicate(const Node& node, string& pred_string) = 0;
+
+  static bool IsControlFlowPredString(string& predicate) {
+    return CONTROL_FLOW_PRED_STRING == predicate;
+  }
+
+  static bool IsTruePredString(string& predicate) {
```
sayantan-nervana: could be `const`

src/ngraph_assign_clusters.cc
sayantan-nervana: both reference args could be `const` (if `IsTruePredString` is also const)

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+        node_pred_string,
+        "should not be clustered in cluster with pred_String ",
+        cluster_pred_string);
+  }
+  return Status::OK();
+}
+#endif
+
+// Merges src and dst clusters of the edge
+void MergeClusters(Edge* edge,
```
sayantan-nervana: maybe add a warning. Only call this when you are sure that X/Y/Z checks have passed (in the sense that this function does not perform any checks)

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+#if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
+    // get predicate string for the node
+    string pred_string;
+    TF_RETURN_IF_ERROR(deadness_analyzer->GetNodePredicate(*node, pred_string));
+    nodes_predicate_map[node] = pred_string;
+    cluster_map[node]->predicate_string = pred_string;
+
+    cluster_map[node]->outgoing_edges = std::set<const Edge*>(
+        node->out_edges().begin(), node->out_edges().end());
+    NGRAPH_VLOG(5) << node->name() << "[" << node->type_string() << "]"
```
sayantan-nervana: minor: well `out_edges` does return an `EdgeSet` which is a set, so we might use that instead of copying to a STL set. But no need to rewrite all the types and functions for that now.

src/ngraph_assign_clusters.cc
```diff
@@ -164,7 +352,8 @@ Status AssignClusters(Graph* graph) {
           gc_success &= gc.InsertEdge(shadow_node_index,
                                       cluster_map[static_edge->dst()]->index);
           if (!gc_success)
-            errors::Internal("Unable to create shadow edges in GraphCycles");
+            return errors::Internal(
+                "Unable to create shadow edges in GraphCycles");
```
sayantan-nervana: Whoa! no error returned before?! :-D

src/tf_deadness_analysis.cc
```diff
@@ -484,6 +467,33 @@ bool DeadnessAnalysisImpl::HasInputsWithMismatchingDeadness(const Node& node) {
...
+    DeadnessAnalysis::GetControlFlowPredString(pred_string);
+    return Status::OK();
+  }
+
+  Predicate* pred = nullptr;
+  for (const Edge* edge : node.out_edges()) {
+    auto it = predicate_map_.find(InputEdgeToTensorId(edge));
+    CHECK(it != predicate_map_.end()) << edge->DebugString();
+
+    // This node is not control flow but has
```
sayantan-nervana: but has?

src/tf_deadness_analysis.h
```diff
@@ -71,9 +87,40 @@ class DeadnessAnalysis {
...
+
+  // For Data Flow ops, updates predicate_string
+  // Deadness is typically introduced by control flow ops. So, all the outgoing
+  // edges from the data flow op have the same deadness predicate ('And'
+  // Predicate of all its input predicates) and we can attach a predicate string
+  // to the data-flow node (predicate of its output edge). Control flow ops are
+  // assigned a placeholder predicate string (CONTROL_FLOW_PRED_STRING) .
+  virtual Status GetNodePredicate(const Node& node, string& pred_string) = 0;
+
+  static bool IsControlFlowPredString(string& predicate) {
```
sayantan-nervana: could be `const`

src/tf_deadness_analysis.h
```diff
@@ -71,9 +87,40 @@ class DeadnessAnalysis {
...
+    return CONTROL_FLOW_PRED_STRING == predicate;
+  }
+
+  static bool IsTruePredString(string& predicate) {
+    return TRUE_PRED_STRING == predicate;
+  }
+
+  static void GetControlFlowPredString(string& predicate) {
+    predicate = CONTROL_FLOW_PRED_STRING;
```
sayantan-nervana: could be `const`

sayantan-nervana: minor: return not needed. the earlier functions do't have it, so for consistency.

src/tf_deadness_analysis.h
```diff
@@ -71,9 +87,40 @@ class DeadnessAnalysis {
...
+    return;
+  }
+
+  static void GetTruePredString(string& predicate) {
+    predicate = TRUE_PRED_STRING;
+    return;
```
sayantan-nervana: could be `const`
minor: return not needed

sayantan-nervana: All the above small functions could be inlined perhaps

test/graph_rewrites/assign_clusters.cc
```diff
@@ -127,6 +136,15 @@ TEST(assign_clusters, cone) {
...
                 .Finalize(&g, &node3));
 
+  // Add edges from SRC to node1 and node2
+  // Add edge from node3 to SINK
+  // The graph is disconnected without these edges
+  Node* source = g.source_node();
+  Node* sink = g.sink_node();
+  g.AddEdge(source, Graph::kControlSlot, node1, Graph::kControlSlot);
+  g.AddEdge(source, Graph::kControlSlot, node2, Graph::kControlSlot);
+  g.AddEdge(node3, Graph::kControlSlot, sink, Graph::kControlSlot);
```
sayantan-nervana: ah, so this is what you meant when u said 'current tests do not work'

test/graph_rewrites/deadness_test.cc
```diff
@@ -48,92 +81,349 @@ as far as the executor is concerned and said node won't get scheduled if any of
 its inputs are dead.
 
 *******************************************************************************/
+//
```
sayantan-nervana: The comment above says `a single XLA cluster`

Are we copying the XLA comment because we want to tell which XLA test we are doing here? or is this comment meant as a description of our test?

test/graph_rewrites/deadness_test.cc
```diff
@@ -48,92 +81,349 @@ as far as the executor is concerned and said node won't get scheduled if any of
...
+  // N1 and N3 are in different cluster
+  ASSERT_NE(N1_Add_cluster, N3_Sub_cluster);
+  // N1 and N5 are in different cluster
+  ASSERT_NE(N1_Add_cluster, N5_Add_cluster);
+  // N3 and N5 are in different cluster
+  ASSERT_NE(N3_Sub_cluster, N5_Add_cluster);
+  // A, N1, N3 and N5 are a different cluster
+  ASSERT_NE(A_cluster, N1_Add_cluster);
+  ASSERT_NE(A_cluster, N3_Sub_cluster);
+  ASSERT_NE(A_cluster, N5_Add_cluster);
```
sayantan-nervana: TBH, glossed over the tests. Hopefully everything is connected right :-D

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+        edge->DebugString());
+  }
+
+  // Case src X , dst Y , X!=Y // cannot be contracted
+  if (!DeadnessAnalysis::IsTruePredString(src_predicate) &&
+      !DeadnessAnalysis::IsTruePredString(dst_predicate) &&
+      src_predicate != dst_predicate) {
+    is_deadness_ok = false;
+    return Status::OK();
+  }
```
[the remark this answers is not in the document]
sayantan-nervana: I am thinking we are handling `true` in an all too special way (well ok, `true` is special, being the zero element of and, which is the property we are exploiting here)

Consider this graph:
```
P1--->P3=P1&P2<-----P2
            |
            v
```

When we consider the P1->P3 edge for merge, then we do not merge it in the current implementation. But they could be merged, since the resulting cluster of P1+P3 has a single outgoing edge, which before and after the merge still holds the same predicate, which is P3

So the rule we have is:
Rule 1: `Consider a edge connecting true and P3. If all destinations of True are P3, then merge`

Instead we could have:
Apply rule 1
Then apply Rule 2: `Consider a edge connecting P1 and P3. it is a single output edge that is being contracted, then contract it`

There is a possibility of a Rule 3 in the horizon when P1 has multiple outputs and we merge P-->P3, when all outputs of P1 are leading to P3s. (because P3=P1&P2, after merge, P1's edges get deadness P1&P3 which equals P3)

After typing it out i realize maybe we are assuming `true` because it makes the computation easier, and we might come back again in a separate PR for a more general case?

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+
+  // Case src X , dst True // invalid scenario
+  // If src has Non-True Predicate and dst has True Predicate, it implies that
+  // the dst node is control flow
+  if (!DeadnessAnalysis::IsTruePredString(src_predicate) &&
+      DeadnessAnalysis::IsTruePredString(dst_predicate)) {
+    return errors::Internal("Attempting to cluster control-flow node ",
+                            dst->name(), "[", dst->type_string(), "]");
+  }
```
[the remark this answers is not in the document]
shresthamalik: As it returns an error i was not setting it.

src/ngraph_assign_clusters.cc
shresthamalik: This function is called inside MergeClusters() which should be called only when ready to merge. So avoided any checks here.

src/tf_deadness_analysis.cc
shresthamalik: Completed this comment :)

src/tf_deadness_analysis.h
sayantan-nervana: I mean `const string& predicate`

all the `could be const` comments mean the references could be const

src/ngraph_assign_clusters.cc
shresthamalik: Yes I have done this test, before the PR

src/ngraph_assign_clusters.cc
shresthamalik: It would still do a copy to a new EdgeSet that we use to define a cluster. May be i am missing the point here.

src/ngraph_assign_clusters.cc
shresthamalik: Yes :D

test/graph_rewrites/assign_clusters.cc
shresthamalik: Yes :)

test/graph_rewrites/deadness_test.cc
shresthamalik: I was quoting the commit message. So kept it as it is. Meant as a description of the test and cite as a reference.

test/graph_rewrites/deadness_test.cc
shresthamalik: Yes, these tests can make your eyes bleed :P .. My glasses broke trying to make sure they are correct. :D

src/tf_deadness_analysis.h
sayantan-nervana: NVM, silly me

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
 
+// Some sanity checks for Node's cluster assignment wrt Deadness
+Status CheckNodeClusterAssignmentWRTDeadness(
+    Node* node, std::map<Node*, string>& nodes_predicate_map,
+    string& cluster_pred_string) {
+  if (nodes_predicate_map.find(node) == nodes_predicate_map.end()) {
+    return errors::Internal("Node ", node->name(), " [", node->type_string(),
+                            "]", " not found in predicate map");
+  }
+  std::string node_pred_string = nodes_predicate_map[node];
```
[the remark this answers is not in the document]
shresthamalik: Did not really care about this since map search is supposed to be O(1) (amortized). But on more thought this should be more efficient . Thanks. 👍

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+string GetMergedClusterPred(string& src_predicate, string& dst_predicate) {
+  return DeadnessAnalysis::IsTruePredString(src_predicate) ? dst_predicate
+                                                           : src_predicate;
+}
+
+// Checks whether it's ok to contract the edge as far as deadness is concerned
+// Source and Dst Predicates of the edge should match
+Status CanContractEdgeDeadnessCheck(
+    Edge* edge, std::map<Node*, std::shared_ptr<Cluster>>& cluster_map,
+    bool& is_deadness_ok) {
```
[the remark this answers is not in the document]
shresthamalik: Tried this. The compiler complains when I am fetching something from the cluster. It thinks I am trying to modify the Cluster arguments. 
For e.g at line number this:
```
string src_predicate = cluster_map[src]->predicate_string;
string dst_predicate = cluster_map[dst]->predicate_string;
```

I am not able to find a simple workaround on this.

src/ngraph_assign_clusters.cc
```diff
@@ -83,23 +91,203 @@ namespace {
...
+  }
+
+  if (!DeadnessAnalysis::IsTruePredString(node_pred_string) &&
+      node_pred_string != cluster_pred_string) {
+    return errors::Internal(
+        "Node ", node->name(), " [", node->type_string(), "]", " Predicate : ",
+        node_pred_string,
+        "should not be clustered in cluster with pred_String ",
+        cluster_pred_string);
+  }
```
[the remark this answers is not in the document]
shresthamalik: In the current implementation, the bottom node can be True. For .e.g this graph merges into 1 cluster with pred P1. This is also one of the unit test case.
```
T --->P1
|
|
v
T----> P1 
|
|
v
P1
```
So, the outgoing edges can be either True or the Non-True Cluster Predicate P1. I will modify the check like this.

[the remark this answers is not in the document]
shresthamalik: I see, in your graph the bottom node is outside the cluster. So the test you suggested takes care of all scenarios. This is a valid Assert when the cluster pred of Node is Non-True (P1), which is evident in the notation, but not in the pseudocode. 
Reiterating for our benefit to avoid any misinterpretation or missing corner cases.
```
//consider a node named node
if (pred_node is True){
    P1 = predicate_of_cluster_node_is_in
    **if(P1 is not True){**
    for n in output_connection_nodes{
            if n is in same cluster as node{
                  //doesnt matter, continue.
            }
            else {
                node n must be in a cluster that has predicate == P1
            }
    }
}
}
```
Some more graphs 
Graph 1
```
T --> T
|   \
|    T 
v
P1
```
In this case there will be cluster of T nodes and an outgoing edge to P1.


Graph 2

T - - >* T
|   \
|    T 
v
P1

In this case there will be a cluster of True Nodes, 1 outgoing edge to True , and another to P1.

src/ngraph_assign_clusters.cc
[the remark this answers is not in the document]
sayantan-nervana: No problem. `[]` has non-constness. For convincing the compiler of your readonly intentions, `map_var.at(key)` works. But minor thing.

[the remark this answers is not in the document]
shresthamalik: Thanks 👍

PR #286, opened 2018-10-29
Shrestha/Add clustering information to NGRAPH_TF_LOG_PLACEMENT.

shresthamalik:
Log
	1. Total Number of Nodes
	2. Number of nodes marked for clustering
	3. Number of nodes assigned a cluster
	4. Number of nGraph Clusters
	5. Size of each nGraph Cluster

### DIFF
--- a/src/ngraph_deassign_clusters.cc
+++ b/src/ngraph_deassign_clusters.cc
@@ -77,17 +77,17 @@ static void MaybeLogPlacement(const Graph* graph) {
   }
 
   int perc_marked_for_clustering_of_total =
-      number_of_nodes > 0
-          ? (int)((float)nodes_marked_for_clustering) * 100.0 / number_of_nodes
-          : 0;
+      number_of_nodes > 0 ? (int)(((float)nodes_marked_for_clustering) * 100.0 /
+                                  number_of_nodes)
+                          : 0;
   int perc_assigned_clusters_of_total =
       number_of_nodes > 0
-          ? (int)((float)nodes_assigned_cluster) * 100.0 / number_of_nodes
+          ? (int)(((float)nodes_assigned_cluster) * 100.0 / number_of_nodes)
           : 0;
   int perc_assigned_clusters_of_marked =
       nodes_marked_for_clustering > 0
-          ? (int)((float)nodes_assigned_cluster) * 100.0 /
-                nodes_marked_for_clustering
+          ? (int)(((float)nodes_assigned_cluster) * 100.0 /
+                  nodes_marked_for_clustering)
           : 0;
 
   std::cout << "Number of nodes in the graph: " << number_of_nodes << std::endl;
--- a/src/ngraph_deassign_clusters.cc
+++ b/src/ngraph_deassign_clusters.cc
@@ -77,13 +77,9 @@ static void MaybeLogPlacement(const Graph* graph) {
   }
 
   int perc_marked_for_clustering_of_total =
-      number_of_nodes > 0 ? (int)(((float)nodes_marked_for_clustering) * 100.0 /
-                                  number_of_nodes)
-                          : 0;
+      (int)(((float)nodes_marked_for_clustering) * 100.0 / number_of_nodes);
   int perc_assigned_clusters_of_total =
-      number_of_nodes > 0
-          ? (int)(((float)nodes_assigned_cluster) * 100.0 / number_of_nodes)
-          : 0;
+      (int)(((float)nodes_assigned_cluster) * 100.0 / number_of_nodes);
   int perc_assigned_clusters_of_marked =
       nodes_marked_for_clustering > 0
           ? (int)(((float)nodes_assigned_cluster) * 100.0 /
--- a/src/ngraph_deassign_clusters.cc
+++ b/src/ngraph_deassign_clusters.cc
@@ -58,7 +58,7 @@ static void MaybeLogPlacement(const Graph* graph) {
 
   std::map<int, std::set<const Node*>> final_cluster_map;
   int number_of_nodes = 0, nodes_marked_for_clustering = 0,
-      nodes_assigned_cluster = 0;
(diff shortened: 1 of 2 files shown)


src/ngraph_deassign_clusters.cc
```diff
@@ -57,19 +57,62 @@ static void MaybeLogPlacement(const Graph* graph) {
...
+    } else {
+      nodes_assigned_cluster++;
     }
     final_cluster_map[cluster_idx].insert(node);
   }
 
+  int perc_marked_for_clustering_of_total =
+      number_of_nodes > 0 ? (int)(((float)nodes_marked_for_clustering) * 100.0 /
+                                  number_of_nodes)
+                          : 0;
```
sayantan-nervana: is `number_of_nodes` ever expected to be zero?
maybe we should fail there due to zero-division so that we would be aware of such a case

sayantan-nervana: might even:
`int number_of_nodes = graph->nodes().size() //assuming size exists`
`if (number_of_nodes<=0) throw error`

then proceed to sure-to-pass division

but this is a minor thing how we calculate `number_of_nodes`. all I am saying is my assumption is `number_of_nodes` is a natural number. so safeguarding division is not required.

sayantan-nervana: (int)(((float)nodes_marked_for_clustering) * 100.0 / number_of_nodes)

the (int) cast at the very top: is it required?

might as well make `perc_marked_for_clustering_of_total` float.

the (float) cast on `nodes_marked_for_clustering` is not required, but does not hurt our cause.

similar comments for all perf variables, they are essentially floats. `int`-ing them is a rounding off to nearest integer

sayantan-nervana: Wait, now that I think of it, maybe u r converting them to int, so that `cout` does not show a gazillion numbers after the decimal. In which case I am ok with treating percentage as an integer. (...maybe... setprecision (but it is persistent, so might have to unset the precision of cout, if u set it) or some-such. so int is fine by me, as it is more hassle-free)

shresthamalik: Yes, I am converting to int to round off to nearest integer. It is just to get an estimate of how many nodes got placed on nGraph, so we do not really care about the decimal point.

Regarding the number_of_nodes, we should not have number_of_nodes<=0 . I did the check for consistency.

src/ngraph_deassign_clusters.cc
```diff
@@ -57,19 +57,58 @@ static void MaybeLogPlacement(const Graph* graph) {
...
+      (int)(((float)nodes_marked_for_clustering) * 100.0 / number_of_nodes);
+  int perc_assigned_clusters_of_total =
+      (int)(((float)nodes_assigned_cluster) * 100.0 / number_of_nodes);
+  int perc_assigned_clusters_of_marked =
+      nodes_marked_for_clustering > 0
+          ? (int)(((float)nodes_assigned_cluster) * 100.0 /
+                  nodes_marked_for_clustering)
+          : 0;
+
+  std::cout << "Number of nodes in the graph: " << number_of_nodes << std::endl;
```
kanvi-nervana: Do we want to print these logs all the time?

src/ngraph_deassign_clusters.cc
```diff
@@ -57,19 +57,58 @@ static void MaybeLogPlacement(const Graph* graph) {
   if (!config::IsLoggingPlacement()) return;
 
   std::map<int, std::set<const Node*>> final_cluster_map;
-
+  int number_of_nodes = 0, nodes_marked_for_clustering = 0,
+      nodes_assigned_cluster = 0;
```
kanvi-nervana: May be change this to nodes_assigned_to_cluster?

src/ngraph_deassign_clusters.cc
```diff
@@ -57,19 +57,58 @@ static void MaybeLogPlacement(const Graph* graph) {
...
     if (!GetNodeCluster(node, &cluster_idx).ok()) {
       cluster_idx = -1;
+    } else {
+      nodes_assigned_cluster++;
     }
     final_cluster_map[cluster_idx].insert(node);
   }
 
+  int perc_marked_for_clustering_of_total =
+      (int)(((float)nodes_marked_for_clustering) * 100.0 / number_of_nodes);
```
kanvi-nervana: Do you need explicit type casting to float and then to int?
Wouldn't (int)/(int) * (int) give an (int) ?

src/ngraph_deassign_clusters.cc
sayantan-nervana: `MaybeLogPlacement` is independent of log levels. hence the `cout` instead of VLOG.

however, we could have 
```
config::IsLoggingPlacement() || log_level >=1
```
in which case `MaybeLogPlacement` is activated through both NGRAPH_TF_LOG_PLACEMENT or VLOG levels. This may or may not be in the scope of this PR

shresthamalik: @kanvi-nervana  If we use NGRAPH_VLOG instead of cout then we will need both NGRAPH_TF_LOG_PLACEMENT and NGRAPH_TF_VLOG_LEVEL set to see the prints. The couts are consistent with the current implementation of NGRAPH_TF_LOG_PLACEMENT/MaybeLogPlacement(). 

@sayantan-nervana Good Suggestion. Though beyond this PR. I am not sure how cluttered we want our NGRAPH_TF_VLOG logs to be. This function does create a lot of logs.  Will wait for @avijit-nervana s comments.

avijit-nervana: We want these logs to go to stdout. If we use vlog - then there are lots of file/timestamp information that get prefixed making it difficult to parse/understand. Also - the nGraph op placement information may be intermixed with other log items - which is not desired either.

PR #287, opened 2018-10-30
R0.7 changes merge to master

avijit-nervana:


README.md
```diff
@@ -154,6 +157,10 @@ The installation prerequisites are the same as described in the TensorFlow
 This final step automatically downloads the necessary version of `ngraph` and 
 the dependencies.
 
+8. Test the installation by running the following command:
+
+        python -c "import tensorflow as tf; print('TensorFlow version: r',tf.__version__);import ngraph; print(ngraph.__version__)"
+
```
sayantan-nervana: This is a good 1 line check!

src/version.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#define VERSION_STR_HELPER(x) #x
+#define VERSION_STR(x) VERSION_STR_HELPER(x)
+
+// e.g. "0.7.0" or "0.7.0-rc0".
+#define NG_TF_VERSION_STRING                                    \
+  (VERSION_STR(NG_TF_MAJOR_VERSION) "." VERSION_STR(            \
+      NG_TF_MINOR_VERSION) "." VERSION_STR(NG_TF_PATCH_VERSION) \
+       NG_TF_VERSION_SUFFIX)
+
```
sayantan-nervana: Unified versioning file!

test/build_tf_cc.sh
```diff
@@ -57,7 +57,7 @@ pushd "${TF_ROOT}"
 ./configure
 
 # Build the TensorFlow C++ Library 
-bazel build --config=opt ${COPT} --config=mkl //tensorflow:libtensorflow_cc.so
+bazel build --config=opt ${COPT} //tensorflow:libtensorflow_cc.so
```
sayantan-nervana: We are removing mkl build?

avijit-nervana: Yes - that's because when we use mkl - some TF ops go to MKL while others go to Eigen - resulting in failures due to precision issues. We shouldn't be testing with MKL (unless we are comparing performance) - so I removed it.

src/version.cc
avijit-nervana: Yes :-)

PR #289, opened 2018-10-31
Sindhu/tf unittest inputfromfile

sindhu-nervana:


test/python/tests_to_run.txt
```diff
@@ -0,0 +1,8 @@
+math_ops_test.DivNoNanTest.testBasic
+#nn_ops
+math_ops_test.X*
+softmax_centered_test.S*
+softmax_op_test.Softmax#
```
sayantan-nervana: is this by design? that `#` anywhere is to be interpreted as comment (instead of just at the beginning)

test/python/tf_unittest_runner.py
```diff
@@ -53,7 +58,15 @@ def main():
         print('\n'.join(test_list))
     if (arguments.run_test):
         test_list = get_test_list(arguments.tensorflow_path, arguments.run_test)
+        print('\n'.join(test_list))
         run_test(test_list)
+    if (arguments.run_tests_from_file):
+        list_of_tests = run_tests_from_file(arguments.run_tests_from_file)
```
sayantan-nervana: the name of the function should be `read_tests_from_file` probably

test/python/tf_unittest_runner.py
```diff
@@ -185,5 +198,18 @@ def run_test(test_list, verbosity=2):
...
+    f = open('tests_to_run.txt')
+    line = f.readline()
+    list_of_tests = []
+    while line:
+        line = line.rstrip('\n')
+        if '#' not in line:
+            list_of_tests.append(line)
+        line = f.readline()
+    f.close()
+    return list_of_tests
```
sayantan-nervana: could be simplified to:
```
with open('tests_to_run.txt') as file_handle:
    return [line.rstrip('\n') for line in file_handle.readlines() if '#' not in line]
```

test/python/tf_unittest_runner.py
```diff
@@ -44,6 +44,11 @@ def main():
         help=
         "Runs the testcase and returns the output. Eg:math_ops_test.DivNoNanTest.testBasic"
     )
+    optional.add_argument(
+        '--run_tests_from_file',
+        help=
+        "Runs the tests specified in a file.Enter the file name. Eg:--run_tests_from_file=tests_to_run.txt"
```
sayantan-nervana: `Eg:--run_tests_from_file` -> `Eg: --run_tests_from_file`

test/python/tests_to_run.txt
sindhu-nervana: Yes, this is by design.  Anything after "#" needs to be ignored. Actually, I need to make few changes here for line 5 to work.

test/python/tf_unittest_runner.py
sindhu-nervana: Hmm we are reading from file and running them instead of just using --run_test where you could input the test name. So run_tests_from_file makes more sense right?

test/python/tests_to_run.txt
sayantan-nervana: anything after `#` is a comment. meaning `abc#def` means `abc`?
There are 3 options I suppose:
1. starts with # is a comment: `x[0]=='#'`
2. contains # is a comment. ignore full line: `'#' in x`
3. ignore after #, but take things before it.: `x.split('#')[0]`
option 3 is the most powerful one I guess

sindhu-nervana: Yes, that is right. I just implemented option #3

test/python/tf_unittest_runner.py
```diff
@@ -185,5 +197,14 @@ def run_test(test_list, verbosity=2):
         test_result = unittest.TextTestRunner(verbosity=verbosity).run(tests)
 
 
+def read_tests_from_file(filename, verbosity=2):
+    with open('tests_to_run.txt') as list_of_tests:
```
mingshan-wang: Should here be `with open(filename)`?

sayantan-nervana: Also, `verbosity` is not required as an input for this function

sindhu-nervana: Thanks for pointing out @mingshan-wang. Made this change and removed verbosity.

PR #297, opened 2018-11-06

yeonsily:
- Beautify the test result
- Return exit code

### DIFF
--- a/diagnostics/model_test/verify_model.py
+++ b/diagnostics/model_test/verify_model.py
@@ -213,6 +213,7 @@ if __name__ == '__main__':
     assert all(
         [i == j for i, j in zip(out_tensor_names_cpu, out_tensor_names_ngraph)])
     passed = True
+    th_dict = {"L1": l1_norm_threshold, "L2": l2_norm_threshold, "inf": inf_norm_threshold}
     for tname, result_ngraph, result_tf_graph in zip(
             out_tensor_names_cpu, result_ngraph_arrs, result_tf_graph_arrs):
         new_out_layer = tname.replace("/", "_")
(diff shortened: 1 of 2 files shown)


avijit-nervana:
@yeonsily The CI failed due to format check:
```
check-code-format.sh STATUS: All 52  C/C++ files pass the code-format check.
1 of 41 source files failed the code-format check:
    diagnostics/model_test/verify_model.py
```
Please run `maint/apply-code-format.sh` and resolve the conflict. Thanks.

diagnostics/model_test/verify_model.py
```diff
@@ -223,20 +225,29 @@ def parse_json():
         l2_norm = calculate_norm(result_ngraph, result_tf_graph, 2)
         inf_norm = calculate_norm(result_ngraph, result_tf_graph, np.inf)
```
sayantan-nervana: Yes, its a good idea to group by tensorname and then display the norms.

The printing of the 3 norms have a lot of repeat code, might I suggest:
```
th_dict = {"L1": l1_norm_threshold, "L2": l2_norm_threshold, "Linf": inf_norm_threshold}

#start the loop and compute norms
norm_dict = {"L1": l1_norm, "L2": l2_norm, "Linf": inf_norm}
print("\n["+tname+"]")
for norm_name in norm_dict:
    if norm_dict[norm_name] > th_dict[norm_name]:
        print("The ",norm_name," norm %f is greater than ",norm_name," threshold %f" % (norm_dict[norm_name], th_dict[norm_name]))
        passed = False
    else:
        #the other prints here
```

diagnostics/model_test/verify_model.py
```diff
@@ -223,20 +225,29 @@ def parse_json():
...
-                  (inf_norm, inf_norm_threshold, tname))
+            print("The inf norm %f is greater than inf threshold %f" %
+                  (inf_norm, inf_norm_threshold))
+            passed = False
         else:
-            print("inf norm test passed for ", tname)
+            print("Inf norm test passed - inf norm: %f, inf treshold: %f" %
+                  (inf_norm, inf_norm_threshold))
+    if not passed:
+        sys.exit(1)
```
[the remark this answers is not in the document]
nervana-davidco1: The exit code serves as a clear failure indicator for automated scripts. Its easier to check exit code than check for some string that indicated a failure.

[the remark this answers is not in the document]
sayantan-nervana: Cool, got it 👍

diagnostics/model_test/verify_model.py
yeonsily: Actually there's a request to show the norms even for passed case. So I put it intentionally. Because we want to see how much gap we have compare to the threshold and if needed to adjust the value.

mingshan-wang: I see. Sounds good, then the current format is good.

PR #301, opened 2018-11-08

yeonsily:
- Beautify the test result
- Return exit code


yeonsily:
Please hold a review.  I keep getting CI check failure and trying to figure out why. I did run apply-code-format.sh though.

PR #302, opened 2018-11-08

yeonsily:
- Beautify the test result
- Return exit code

### DIFF
--- a/diagnostics/model_test/verify_model.py
+++ b/diagnostics/model_test/verify_model.py
@@ -213,7 +213,11 @@ if __name__ == '__main__':
     assert all(
         [i == j for i, j in zip(out_tensor_names_cpu, out_tensor_names_ngraph)])
     passed = True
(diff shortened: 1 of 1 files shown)


yeonsily:
I have no idea why CI check is failing. Can someone check the CI build log and let me know if this is a real issue?

sayantan-nervana:
Yes, CI looks good. Will investigate why it fails

yeonsily:
Was using python3 to run format check script.

diagnostics/model_test/verify_model.py
```diff
@@ -211,6 +212,12 @@ def parse_json():
 
     assert all(
         [i == j for i, j in zip(out_tensor_names_cpu, out_tensor_names_ngraph)])
+    passed = True
+    th_dict = {
```
shresthamalik: Please rename th_dict to threshold_dict.

PR #310, opened 2018-11-15
Use output of forward pooling in ngraph::op::MaxPoolBackprop

csullivan:
This will require targeting a yet-to-be merged branch of ngraph before the bridge CI will be able to compile.


sayantan-nervana:
TESTNOW

sayantan-nervana:
Merging master which contains changes that fixes quantization api change related build breaks

sayantan-nervana:
Failing now due to old ngraph. should pass when we move to 0.10.0.rc0

PR #311, opened 2018-11-15
Sindhu/Fix GNMT crash for DE-EN use case

sindhu-nervana:


src/ngraph_builder.cc
```diff
@@ -3264,35 +3264,54 @@ static Status TranslateSplitVOp(
...
+  int shape_length = 0;
+  int idx;
+  for (int i = 0; i < lengths.size(); ++i) {
+    if (lengths[i] != -1) {
+      length += lengths[i];
+    } else {
+      idx = i;
+    }
+  }
+
```
sayantan-nervana: init `idx` to something like `-1`, else klocwork would probably complain

src/ngraph_builder.cc
```diff
@@ -3264,35 +3264,54 @@ static Status TranslateSplitVOp(
...
-
-  for (int i = 0; i < rank; ++i) {
-    lower.push_back(0);
-    upper.push_back(shape[i]);
+  for (int i = 0; i < shape.size(); ++i) {
+    shape_length += shape[i];
+  }
+  if (shape_length > length) {
+    lengths[idx] = shape_length - length;
   }
```
sayantan-nervana: the `shape_length` needs only be computed only when `idx < 0`
so something like:
```
if (idx < 0){
//compute shape_length
lengths[idx] = shape_length - length; //if (shape_length > length) not required
}
```

src/ngraph_builder.cc
```diff
@@ -3264,35 +3264,54 @@ static Status TranslateSplitVOp(
...
-    lower[split_dim] = cursor;
-    cursor += lengths[i];
-    upper[split_dim] = cursor;
-    SaveNgOp(ng_op_map, op->name(),
-             make_shared<ng::op::Slice>(ng_input, lower, upper));
+  if (lengths.size() != 1) {
+    for (int i = 0; i < lengths.size(); ++i) {
+      cout << "Printing lengths vector:" << lengths[i] << endl;
+      lower[split_dim] = cursor;
+      upper[split_dim] = cursor;
```
```
lower[split_dim] = cursor;
upper[split_dim] = cursor;
```

Both are being set to exact same value, hence we see `0`-dimension

please note of the original code, which does the cursor updating in between setting lower and upper

src/ngraph_builder.cc
sayantan-nervana: shape[split_dim] - length

test/test_array_ops.cpp
```diff
@@ -667,6 +667,40 @@ TEST(ArrayOps, SpaceToDepthNCHW) {
   }
 }  // end of op SpaceToDepthNCHW
 
+// Test SplitV op
+TEST(ArrayOps, SplitV) {
+  std::vector<std::vector<int64>> input_shapes;
+  input_shapes.push_back({1, 2, 6, 1});
```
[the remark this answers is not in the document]
sindhu-nervana: Added more tests @sayantan-nervana

sindhu-nervana:
@jianyinglang  Could you please review this change when you get time.

src/ngraph_builder.cc
```diff
@@ -3327,35 +3327,51 @@ static Status TranslateSplitVOp(
...
-  }
+  std::vector<size_t> lower(rank, 0);
+  std::vector<size_t> upper(shape);
 
   std::vector<int> split_dim_vec;
   TF_RETURN_IF_ERROR(
       GetStaticInputVector(op, 2, static_input_map, &split_dim_vec));
-  int split_dim = split_dim_vec[0];
+
+  int split_dim = split_dim_vec[0] + (split_dim_vec[0] < 0 ? (int64)rank : 0);
```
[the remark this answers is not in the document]
sindhu-nervana: Yes, it could be. The dimension along which to split. Must be in the range [-rank(value), rank(value)).

jianyinglang:
LGTM

PR #314, opened 2018-11-16
Shrestha/op executor tensor creation

shresthamalik:
The Op Executor was not copying the inputs to the backend memory space. Added that. Other than that have added some comments for freshness_tracker.


shresthamalik:
Needs testing on some backends like GPU etc. before merge.

test/opexecuter.cpp
```diff
@@ -394,7 +401,17 @@ void OpExecuter::ExecuteOnNGraph(vector<Tensor>& ngraph_outputs,
 
   // Execute the nGraph
   NGRAPH_VLOG(5) << " Executing on nGraph ";
-  backend->call(ng_function, ng_op_tensors, ng_ip_tensors);
+  BackendManager::LockBackend(ng_backend_type);
+  try {
+    backend->call(ng_function, ng_op_tensors, ng_ip_tensors);
+  } catch (const std::exception& exp) {
```
sayantan-nervana: `std::exception` catches a broad class of errors. Should we have 1 case for ngraph errors that catch `ngraph_error`? https://github.com/NervanaSystems/ngraph/blob/b0b8719d7fd2dd5fce0ddc982fb57e876ddc0c00/src/ngraph/except.hpp#L25

Well, in keeping with the `call` function in encapsulate op, I suppose we have 2 catches one for `std::exception` and one for everything else.

The only reason it is working out for now is `ngraph_error` is actually a `std::runtime_error`. So it is getting caught there.

But I wonder, `std::exception` if we should have "catch ngraph errors, then catch std::exception, then catch everything else". A short term solution would be to catch only `std::runtime_error` and call then ngraph errors, and let everything fall to the default case.

avijit-nervana: Since this is a unit test class - I would say this level of exception handling is fine for now. We need to review the EncapSulate op though to see if we need more fine grained error handling.

sayantan-nervana: Agree this is beyond the scope of this PR...  in case we need finer error handling. Then we can change both encapsulate op and opexecutor's try-catches.

avijit-nervana:
TESTNOW

PR #317, opened 2018-11-19
Langjian/klocwork bugs

jianyinglang:
Try some simple fix.


src/ngraph_builder.cc
```diff
@@ -2432,7 +2432,8 @@ Status QuantizeAndDequantizeV2Helper(
                                     ? max_quantized / max_range
                                     : std::numeric_limits<T>::max();
   T scale, inverse_scale;
-  if (scale_from_min_side < scale_from_max_side) {
+  if (scale_from_min_side < scale_from_max_side && min_quantized != 0) {
+    // min_quantized != 0 is not really necessary but klocwork complains
```
sayantan-nervana: `const int64 min_quantized = signed_input ? -(1ULL << (num_bits - 1)) : 0;`
`min_quantized` is `0` only when `signed_input` is `false`

In which case, `min_quantized=0` and `max_quantized = ((1ULL << num_bits) - 1)`
Then `scale_from_min_side = std::numeric_limits<T>::max();`

Now, `scale_from_max_side = (max_quantized * max_range > 0) ? max_quantized / max_range : std::numeric_limits<T>::max();` ... which is non-zero unless `max_quantized * max_range <= 0 and max_quantized=0`

`if scale_from_min_side < scale_from_max_side`
but `scale_from_min_side=std::numeric_limits<T>::max()`

So the condition reduces to: `if std::numeric_limits<T>::max() < scale_from_max_side`

If `scale_from_max_side = std::numeric_limits<T>::max()` too, then we go to the else part, where we divide by `max_quantized`. But `max_quantized = ((1ULL << num_bits) - 1)`, so safe

if `scale_from_max_side = max_quantized / max_range` then the condition becomes:
`if std::numeric_limits<T>::max() < max_quantized / max_range`
or
`if std::numeric_limits<T>::max() < ((1ULL << num_bits) - 1) / max_range`
Note `signed_input=false` in our initial assumption, so all numbers are `>=0`

For that to be true, we must have: `((1ULL << num_bits) - 1)/std::numeric_limits<T>::max()  > max_range`

Lets sat `T=float`, then `std::numeric_limits<T>::max() = 3.40282e+38`. with `num_bits=8`, we get:
`((1<<8) -1 / 3.40282e+38) > max_range`
`(255 / 3.40282e+38) > max_range`
`7.4937846e-37 > max_range`
Now min possible value for float is 1.17549e-38, so it is possible to get `0 < max_range < 7.4937846e-37`.

However, this code is from TF's reference kernel, so not sure if we want to change behaviour in that corner case.
https://github.com/tensorflow/tensorflow/blob/9590c4c32dd4346ea5c35673336f5912c6072bf2/tensorflow/core/kernels/quantize_and_dequantize_op.h#L72

PR #324, opened 2018-11-21
Fix new klocwork error reports on 11/20/2018

jianyinglang:
Try to fix the critical error found in klocwork on 11/20/2018.

### DIFF
--- a/src/tf_deadness_analysis.cc
+++ b/src/tf_deadness_analysis.cc
@@ -490,7 +490,9 @@ Status DeadnessAnalysisImpl::GetNodePredicate(const Node& node,
   }
 
   // All outputs have the same predicate
-  pred_string = pred->ToString();
+  if (pred != nullptr) {
+    pred_string = pred->ToString();
+  }
   return Status::OK();
 }
 
--- a/src/ngraph_deassign_clusters.cc
+++ b/src/ngraph_deassign_clusters.cc
@@ -75,6 +75,7 @@ static void MaybeLogPlacement(const Graph* graph) {
     }
     final_cluster_map[cluster_idx].insert(node);
   }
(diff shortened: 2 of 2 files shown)


src/tf_deadness_analysis.cc
```diff
@@ -490,7 +490,9 @@ Status DeadnessAnalysisImpl::GetNodePredicate(const Node& node,
   }
 
   // All outputs have the same predicate
-  pred_string = pred->ToString();
+  if (pred != nullptr) {
+    pred_string = pred->ToString();
+  }
```
sayantan-nervana: Minor:

Here is another possibility:
Advantages of this 
1) Ensures that `pred_string` is assigned unconditionally.
2) Removes the `pred != nullptr` check
3) removes the assignment `pred = it->second;`, which keeps happening each time in the loop, but is not needed.

```
auto out_edgeset = node.out_edges();
// Initialize with 0th element
// It is guaranteed that there will be at least 1 out edge
Predicate* pred = predicate_map_.find(InputEdgeToTensorId(*(out_edgeset.begin())));
// Iterate from the 1th element to end
for (const Edge* edge : out_edgeset.begin() + 1)) {
     auto it = predicate_map_.find(InputEdgeToTensorId(edge));
     CHECK(it != predicate_map_.end()) << edge->DebugString();
 
     // This node is not control flow but has different output predicates
// Now we do not need the check "pred != nullptr "
     if (*pred != *it->second) {
       return errors::Internal(node.name(), "[", node.type_string(), "]",
                               " is a non control flow op. But its outputs have "
                               "different predicates");
     }
    // This line is not needed
    // pred = it->second;
   }


pred_string = pred->ToString();
```

PR #328, opened 2018-11-27
Modify Mac builds to use TensorFlow 1.12.0 for builds

crlishka:
Mac builds were still using "tensorflow==1.12.0rc2" for builds.  This PR updates them to use "tensorflow==1.12.0".

Successful testing: 
https://aipg-rancher.intel.com/jenkins/algo/job/ngraph-tf-mac-daily-build/86/


avijit-nervana:
BTW - all builds should use TF1.12.0 (and not the rc2)

crlishka:
TESTNOW

crlishka:
> BTW - all builds should use TF1.12.0 (and not the rc2)

Agreed.  This is one that I forgot to update (probably right before I went on vacation).

Ideally, in the future, we can just use "pip install tensorflow", without any version number.  But for now we will lock the version.

crlishka:
Previous CI job ended up on a node, hung.  Checked with Nick, job should not have hung, he rebooted the node.  I am resubmitting the CI job.

crlishka:
TESTNOW

PR #330, opened 2018-11-28
Get libplaidml from the plaidml package

earhart:
With this change, the generated wheel doesn't have to include libplaidml itself in order for PlaidML to work; PlaidML can be installed separately, as long as the wheel contains libplaidml_backend (which is ngraph code).


earhart:
Removing the "enhancement" label since, now that libplaidml.dylib is installed in a more useful location, adjusting the rpath of the backend actually breaks things.

earhart:
🎉 Thanks, Avijit!  :-)

PR #333, opened 2018-11-29
Mingshan/Added floor_div unit tests for negative integer and float inputs

mingshan-wang:
Added a floor div C++ unit tests to track the wrong implementation of floor_div op in NGraph for integer inputs.


test/test_math_ops.cpp
```diff
@@ -671,6 +671,57 @@ TEST(MathOps, FloorModBroadcasting) {
...
+  vector<float> A_vect{-1.f}; 
+  vector<float> B_vect{3.f}; 
+
+  AssignInputValues(A, A_vect);
+  AssignInputValues(B, B_vect);
+
+  vector<int> static_input_indexes = {};
+  auto R = ops::FloorDiv(root, A, B);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
```
sayantan-nervana: If inputs are float, out of floordiv is float, but if they are int, output is int? That is interesting, because usually functions like say `std::floor` always output float/double for both int and float inputs.

Do we also want to add a mixed test? A is int, B is float?

sayantan-nervana:
formatting failure

test/test_math_ops.cpp
mingshan-wang: I think for all Integer input case, the expected output is also Integer. So current NGraph calculation was not right for this case. 

Yes, agree, I will add a mixed test case as well.

avijit-nervana:
@mingshan-wang Please resolve the conflicts

sayantan-nervana:
TESTNOW

mingshan-wang:
TESTNOW

PR #337, opened 2018-11-29
Kanvi/fix split op

kanvi-nervana:
- Add functionality to handle negative axis
 - Add unit tests for the op as well


test/test_array_ops.cpp
```diff
@@ -906,6 +906,70 @@ TEST(ArrayOps, DISABLED_StridedSlice) {
...
+// Test SplitNegativeAxis op
+TEST(ArrayOps, SplitNegativeAxis) {
+  std::vector<std::vector<int64>> input_shapes;
+  input_shapes.push_back({1, 2, 8, 1});
+  // num_split : The number of ways to split. Must evenly divide
+  // value.shape[split_dim]
+  int64_t num_splits = 4;
+
+  vector<int> static_input_indexes = {0};
+  vector<DataType> output_datatypes = {DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT};
```
sayantan-nervana: minor:
`vector<DataType> output_datatypes(DT_FLOAT, num_splits)` is safer. Then you don't have to manually count the number of outputs

PR #340, opened 2018-11-30
Mingshan/Adding floor_mod test with negative int and float inputs

mingshan-wang:
Added FloorMod negative values input C++ unit test. Disabled integer input for now since it is failing, until NGraph fixes FloorDiv, that should also fixes the FloorMod issue. 

Float negative inputs are working.


test/test_math_ops.cpp
```diff
@@ -671,6 +671,57 @@ TEST(MathOps, FloorModBroadcasting) {
...
+  auto R = ops::FloorMod(root, A, B);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "FloorMod", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op FloorModNegInt
```
sayantan-nervana: Very minor:
// end of test op FloorModNegFloat

mingshan-wang: Thanks for catching this!

PR #342, opened 2018-11-30
Fixed klockworks errors

avijit-nervana:
This PR fixes the Klockworks errors. Here's the link to the appropriate view: https://klocwork-ir2.devtools.intel.com:8080/review/insight-review.html#issuelist_goto:project=ngraph_tf,searchquery=state%253A%252BNew,sortcolumn=id,sortdirection=ASC,view_id=8

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -3517,8 +3517,10 @@ static Status TranslateStridedSliceOp(
     // The first 2 cases breaks down this range
     if (idx >= 0 && idx <= (dim - 1)) {
       return idx;
-    } else if (idx < 0 && idx + static_cast<int>(dim) >= 0) {  // careful not to do idx >= -dim
-                                             // (since dim is unsigned)
+    } else if (idx < 0 &&
+               idx + static_cast<int>(dim) >=
+                   0) {       // careful not to do idx >= -dim
+                              // (since dim is unsigned)
       return idx + (int)dim;  // Type casting to int to enable unambiguous auto
                               // type inference of return type
     } else if (idx > dim - 1) {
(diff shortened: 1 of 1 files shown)


src/ngraph_builder.cc
```diff
@@ -3523,12 +3523,15 @@ static Status TranslateStridedSliceOp(
                               // type inference of return type
     } else if (idx > dim - 1) {
       return (int)dim;
-    }  // The next case handles the clamping (differently for inclusive and
-       // exclusive cases)
-    else if (idx + dim <
-             0) {  // careful not to do idx < -dim (since dim is unsigned)
+    } else if (static_cast<int>(idx + dim) < 0) {
```
sayantan-nervana: there is a `idx + dim >= 0` in line 3520 as well, which adds an int and a size_t. Do you want to cast that as well.
I think clockwork complains for line 3520 as well: https://klocwork-ir2.devtools.intel.com:8080/review/insight-review.html#issuedetails_goto:offset=0,problemid=9450,project=ngraph_tf,searchquery=state%253A%252BNew,sortcolumn=id,sortdirection=ASC,start=0,view_id=8

Also, I wonder should we cast dim to int (like line 3522) or cast the sum to int? casting the sum to int works, but in that case we are still adding 2 different types. casting dim to int first would add same types.

src/ngraph_builder.cc
```diff
@@ -3523,12 +3523,15 @@ static Status TranslateStridedSliceOp(
...
+      // The next case handles the clamping (differently for inclusive and
+      // exclusive cases)
+
+      // careful not to do idx < -dim (since dim is unsigned)
       return 0 - (inclusive ? 0 : 1);
     }
+    // Default case
+    return 0;
```
sayantan-nervana: looks good: This line should never be reached, so any value returned here is fine.

src/ngraph_builder.cc
avijit-nervana: Yes - I think so.

PR #344, opened 2018-11-30
Fix SplitV op

kanvi-nervana:
- Add invalid argument errors if size_splits != length

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -3332,30 +3332,48 @@ static Status TranslateSplitVOp(
   std::vector<int> lengths;
   TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &lengths));
 
-  int length = 0;
-  int idx = -1;
-  for (int i = 0; i < lengths.size(); ++i) {
-    if (lengths[i] != -1) {
-      length += lengths[i];
-    } else {
-      idx = i;
-    }
-  }
-
   ng::Shape shape = ng_input->get_shape();
   int rank = shape.size();
   std::vector<size_t> lower(rank, 0);
   std::vector<size_t> upper(shape);
-
   std::vector<int> split_dim_vec;
   TF_RETURN_IF_ERROR(
       GetStaticInputVector(op, 2, static_input_map, &split_dim_vec));
 
   int split_dim = split_dim_vec[0] + (split_dim_vec[0] < 0 ? (int64)rank : 0);
 
+  bool has_one_neg = false;
+  
+  // length: Length of size_splits
+  int length = 0;
+  int idx = -1;
+
+  // Find out the total length of the splits
+  for (int i = 0; i < lengths.size(); ++i) {
+    if (lengths[i] != -1) {
+      length += lengths[i];
+    } else {
+      if(has_one_neg) {
+        return errors::InvalidArgument(
+          "size_splits can only have one -1");
+      } else {
+        idx = i;
+      }
+    }
+  }
+
   // Size splits must sum to the dimension of value along split_dim
-  if (idx > 0) {
+  if(idx == -1) { // there is no size_splits with -1
+    if(length != shape[split_dim]) {
+        return errors::InvalidArgument(
+          "The length of size_splits must sum to the dimension of value along split_dim");
+    }
+  } else { // there is a size_splits with -1
     lengths[idx] = shape[split_dim] - length;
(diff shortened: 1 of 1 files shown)


src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,45 @@ static Status TranslateSplitVOp(
...
+  // Find out the total length of the splits
+  for (int i = 0; i < lengths.size(); ++i) {
+    if (lengths[i] != -1) {
+      length += lengths[i];
+    } else {
+      if(has_one_neg) {
+        return errors::InvalidArgument(
+          "size_splits can only have one -1");
+      } else {
+        idx = i;
```
shresthamalik: Need to set has_one_neg true, in else. May be has_only_one_neg would be more explicable?

src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,45 @@ static Status TranslateSplitVOp(
...
+      }
+    }
+  }
+
+  if(idx > 0) {
     lengths[idx] = shape[split_dim] - length;
   }
 
+  if(length != shape[split_dim] || length != shape[split_dim] - lengths[idx]) {
```
shresthamalik: Please add two separate conditions. ```(length!=shape[split_dim)``` is a valid error condition when ```idx==-1```

src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,45 @@ static Status TranslateSplitVOp(
...
-  // Size splits must sum to the dimension of value along split_dim
-  if (idx > 0) {
+  bool has_one_neg = false;
+  
+  // length: Length of size_splits
+  int length = 0;
+  int idx = -1;
+
+  // Find out the total length of the splits
+  for (int i = 0; i < lengths.size(); ++i) {
```
sayantan-nervana: The comment before this for loop says: `Find out the total length of the splits`, but from what I understand it is locating the one negative value if present.

Also, I think it was checked last time, but can we recheck what happens if we pass if a neg number other than -1? Are we relying on TF complaining about it during construction?

kanvi-nervana: TF complains if we specify any other neg number other than -1

src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,46 @@ static Status TranslateSplitVOp(
...
-  // Size splits must sum to the dimension of value along split_dim
-  if (idx > 0) {
+  bool has_one_neg = false;
+  
+  // length: Length of size_splits
+  int length = 0;
+  int idx = -1;
+
+  // Find out the total length of the splits and locate it's index
+  for (int i = 0; i < lengths.size(); ++i) {
```
sayantan-nervana: Minor:

`locate it's index` -> `locate it's negative (-1) index, if any`

src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,46 @@ static Status TranslateSplitVOp(
...
+        has_one_neg = true;
+      }
+    }
+  }
+
+  if(idx > 0) {
     lengths[idx] = shape[split_dim] - length;
   }
 
+  if((idx == -1 && length != shape[split_dim]) || (idx > 0 && lengths[idx] <= 0)) {
```
sayantan-nervana: Minor readability comment:
instead of `idx == -1` we can use `!has_one_neg`, which might be more readable
Similarly for `idx > 0` we can use `has_one_neg`
So: `(!has_one_neg && length != shape[split_dim]) || (has_one_neg && lengths[idx] <= 0)`

Possible corner case:
I am wondering about the `<= 0` part.. Can a split have `0` length? We could have a test. If `0` size splits are acceptable, then we need to make that `< 0` and not `<= 0`

src/ngraph_builder.cc
```diff
@@ -3332,32 +3332,46 @@ static Status TranslateSplitVOp(
...
+          return errors::InvalidArgument(
+          "The length of size_splits must sum to the dimension of value along split_dim");
```
sayantan-nervana: `split_dim` is an internal local variable. We should use `axis`, which is the TF name for this variable.
`dimension of value` -> did you mean `value of dimension`?

Also you could print out the value of `axis`

Ideally, the 2 error cases should get different error messages. Maybe we can split them into 2 `if`s, instead of `or`ing the error condition `((idx == -1 && length != shape[split_dim]) || (idx > 0 && lengths[idx] <= 0))`

PR #345, opened 2018-11-30
Support multiple random value range for multiple input tensors

yeonsily:


diagnostics/model_test/verify_model.py
```diff
@@ -187,21 +187,26 @@ def parse_json():
...
 
     # Matches the input tensors name with its required dimensions
     input_tensor_dim_map = {}
+    i = 0
     for (dim, name) in zip(input_dimension, input_tensor_name):
         random_input = np.random.randint(
-            rand_val_range, size=[bs] + dim).astype('float32')
+            rand_val_range[i], size=[bs] + dim).astype('float32')
         input_tensor_dim_map[name] = random_input
+        i = i + 1
```
sayantan-nervana: Style issue about the extra variable `i`:

Instead of zipping input_dimension and input_tensor_name, and then tracking an `i` variable separately, we can either `for i in range(len(input_dimension)):`
or: zip all three input_dimension, input_tensor_name and rand_val_range

```
for (dim, name, val_range) in zip(input_dimension, input_tensor_name, rand_val_range):
    input_tensor_dim_map[name] = np.random.randint(val_range, size=[bs] + dim).astype('float32')
```

yeonsily: Thanks for the comment. I chose to zip all three parameters.

diagnostics/model_test/README.md
```diff
@@ -14,7 +14,7 @@ multiple input tensors and output tensors. Put the input tensor names as a list
 * Specify the the location of the graph file in the ```graph_location``` of the json file
 * Specify the ```batch_size``` field in the json file to the desired batch size for inference
 * Specify the tolerance between the TF and NGraph outputs at ```l1_norm_threshold```, ```l2_norm_threshold``` and ```inf_norm_threshold``` in the json file 
-* Specify the ```random_val_range``` used to generate the input within 0 to random_val_range
+* Specify the ```random_val_range``` used to generate the input within 0 to random_val_range. You will need to specify them for all the input tensors provided
```
mingshan-wang: Can we change the name of this parameter to `random_val_max` because it is specifying the max value only, not like a range (0~255)?

yeonsily: Make sense. Will include it for next change.

PR #347, opened 2018-11-30
Sindhu/Fix TF Python Slicetest segfault

sindhu-nervana:


src/ngraph_builder.cc
```diff
@@ -3658,6 +3657,11 @@ static Status TranslateStridedSliceOp(
   auto dim_vec = ng_input->get_shape();
   auto in_rank = dim_vec.size();
 
+  if (in_rank == 0) {
+    return errors::InvalidArgument("Index out of range using input dim ",
+                                   "; input has only ", in_rank, " dims");
+  }
```
sayantan-nervana: New tests:
x of dim [4,0,5].
`x[1:2, 2:2, 1:2]`. (slicing an empty dim by empty slice)
`x[1:2, 1:2, 1:2]`. (slicing an empty dim by non empty slice)

`x = 0` (scalar)
`x[:]` (supposed to fail.. this PR fixes the output error for this case)
`x[1:1]` (empty slice. Does this fail, if so what failure)

sindhu-nervana: @sayantan-nervana Added these tests, please review.

test/python/test_slice.py
```diff
@@ -150,3 +151,45 @@ def run_test(sess):
...
+
+        #(slicing an empty dim by empty slice)
+        slice_ts.append(x[1:1])
+
+        def run_test(sess):
+            return sess.run(slice_ts, feed_dict={x: inp})
+
+        with pytest.raises(Exception) as excinfo:
+            slice_vals = self.with_ngraph(run_test)
+            assert str(excinfo.value) == str(excinfo)
```
shresthamalik: I was trying something similar and found that assert inside with pytest.raises doesnt work. Would want to verify with you.
Right now the assert condition doesnt really check anything
We need to change it to:
```
assert "<error print>" in excinfo.value.message
```

sindhu-nervana: Changed this, please review.

PR #348, opened 2018-12-01
Mingshan/Adding support for FusedBatchNormV2 on the bridge

mingshan-wang:
Added: Use the same translation function as FusedBatchNorm, since the only different between FusedBatchNorm and FusedBatchNormV2 is the supported input data type.

Also add a TODO comment in the type_constraint_map to support DT_HALF and DT_BFLOAT16 for FusedBatchNormV2 when NGraph supports those data type.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -3982,6 +3982,7 @@ const static std::map<
         {"FloorDiv", TranslateFloorDivOp},
         {"FloorMod", TranslateFloorModOp},
         {"FusedBatchNorm", TranslateFusedBatchNormOp},
+        {"FusedBatchNormV2", TranslateFusedBatchNormOp},
         {"FusedBatchNormGrad", TranslateFusedBatchNormGradOp},
         {"Greater", TranslateBinaryOp<ngraph::op::Greater>},
         {"GreaterEqual", TranslateBinaryOp<ngraph::op::GreaterEq>},
--- a/src/ngraph_mark_for_clustering.cc
+++ b/src/ngraph_mark_for_clustering.cc
@@ -239,6 +239,8 @@ Status MarkForClustering(Graph* graph) {
       confirmation_function_map["FloorMod"] = SimpleConfirmationFunction();
       confirmation_function_map["FusedBatchNorm"] =
           SimpleConfirmationFunction();
+      confirmation_function_map["FusedBatchNormV2"] =
+          SimpleConfirmationFunction();
       confirmation_function_map["FusedBatchNormGrad"] = [](Node* n,
                                                            bool* result) {
         TF_RETURN_IF_ERROR(GetNodeAttr(n->attrs(), "is_training", result));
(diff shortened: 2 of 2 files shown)


mingshan-wang:
Sounds good, I will add a unit test for FusedBatchNormV2 in our C++ unit tests.

test/test_nn_ops.cpp
```diff
@@ -588,7 +588,7 @@ TEST(NNOps, Conv2DBackpropInputNCHWSame) {
   opexecuter_tf.ExecuteOnTF(tf_outputs);
 
   // Compare NGraph and TF Outputs
-  Compare(tf_outputs, ngraph_outputs);
+  Compare(tf_outputs, ngraph_outputs, 1e-05, 1e-05);
```
shresthamalik: Are they not passing with the default values ? 1e-05, 1e-08 ?

mingshan-wang: Yes, I ran the whole tests for ~50 runs, for some run, those nn ops fail, so I changed them to a looser tolerance.

One thing to notice is that even with 1e-05 and 1e-05, they are more strict than our previous Compare method I believe because according to our formula, if rtol = 0, the atol is 1e-05, so abs(desired - actual) <= 0.00001, which previously is 0.001.

src/ngraph_builder.cc
```diff
@@ -1744,7 +1761,7 @@ static Status TranslateFusedBatchNormOp(
     SaveNgOp(ng_op_map, op->name(), ng_batch_norm);
   }
 
-  SaveNgOp(ng_op_map, op->name(), ng_batch_norm);
+  // SaveNgOp(ng_op_map, op->name(), ng_batch_norm);
```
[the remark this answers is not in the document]
mingshan-wang: Yes, realized it. Once yours is merged, I will merge with master again and resolve this conflict.

test/test_nn_ops.cpp
```diff
@@ -912,6 +912,92 @@ TEST(NNOps, Conv2DBackpropInputNHWCWithDilation) {
...
+  // used for inference only, must be empty for training
+  Tensor mean(DT_FLOAT, TensorShape({3}));
+  // 1D tensor for population variance
+  // used for inference only, must be empty for training
+  Tensor variance(DT_FLOAT, TensorShape({3}));
+
+  AssignInputValuesRandom<float>(x, -30.0f, 50.0f);
+  AssignInputValuesRandom<float>(scale, 0.2f, 3.f);
+  AssignInputValuesRandom<float>(offset, 1.1f, 1.5f);
+  AssignInputValuesRandom<float>(mean, 0.5f, 3.5f);
```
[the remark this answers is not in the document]
mingshan-wang: Sounds good! Thanks for the suggestion, will input negative values to test.

test/test_utilities.h
```diff
@@ -85,26 +85,31 @@ void AssignInputValuesRandom(Tensor& A, T min, T max) {
...
+bool Compare(float arg0, float arg1, float rtol, float atol);
 
 // Compares two Tensors
+// Right now only tensors contain float values will modify the tolerance
+// parameters
 template <typename T>
-void Compare(const Tensor& T1, const Tensor& T2) {
+void Compare(const Tensor& T1, const Tensor& T2,
+             float rtol = static_cast<float>(1e-05),
+             float atol = static_cast<float>(1e-08)) {
```
shresthamalik: Should we define global macros for RTOL and ATOL values and use them in all the functions? @sayantan-nervana any suggestions around this. We have the rtol and atol as default arguments for many functions

PR #349, opened 2018-12-01
Doc update

avijit-nervana:


README.md
```diff
@@ -181,7 +111,7 @@ above and use the following option for building with nGraph:
 
         cd tensorflow
         git checkout v1.12.0
-        bazel build --config=opt --config=mkl --config=ngraph //tensorflow/tools/pip_package:build_pip_package 
+        bazel build --config=opt --config=ngraph //tensorflow/tools/pip_package:build_pip_package
```
sayantan-nervana: Around line 192: `ngraph module` -> `nGraph module`

shresthamalik:
Do we want to add the git checkout instruction in examples/README.md as discussed in the morning?
git checkout 4c7b09ad87bbfc4b1f89650bcee40b3fc5e7dfed

README.md
```diff
@@ -168,9 +101,6 @@ with nGraph backends.
 
 Please add the following line to enable nGraph: `import ngraph_bridge`
 
-Note: The actual filename for the pip package may be different as it's version 
-dependent. Please check the `build/python/dist` directory for the actual pip wheel.
-
```
sayantan-nervana: Line 176:
`nGraph is being added to the TensorFlow source tree the using pull requests from 
 time to time`

extra `the`

How about:
`nGraph is updated in the TensorFlow source tree using pull requests from time to time`

README.md
sayantan-nervana: `With this configuration, your TensorFlow model scripts will work without any changes.`

how about:
`With this configuration, your TensorFlow model scripts will work without any changes, ie, you do not need to add import ngraph_bridge, like option 1 and 2.`

sayantan-nervana: This line:
```
cd test
./gtest_ngtf
```

They need code formatting. Currently they show up as plain text

sayantan-nervana: The mint example model: `some MNIST data on your local machine`
This model is not directly runnable. The `python keras_sample.py` works out-of-the-box. Do we want to add that?

It tests both inference and training.

sayantan-nervana: `idential`->`identical`

sayantan-nervana: You mentioned for Mac we do not need this part: `Export the appropriate paths `
But if they do no harm, we can keep them for now I suppose?

README.md
```diff
@@ -82,84 +78,21 @@ The installation prerequisites are the same as described in the TensorFlow
...
-        pip install -U ./tensorflow-1.*whl
-   
-6. Now clone the `ngraph-tf` repo one level above -- in the 
-  *parent* directory of the `tensorflow` repo cloned in step 1:
-
-        cd ..
         git clone https://github.com/NervanaSystems/ngraph-tf.git
         cd ngraph-tf
         git checkout v0.8.0
```
shresthamalik: There is no tag for v0.8.0
See this error when i checkout 
```
git checkout v0.8.0
```
error: pathspec 'v0.8.0' did not match any file(s) known to git.

README.md
```diff
@@ -181,7 +111,7 @@ above and use the following option for building with nGraph:
 
         cd tensorflow
         git checkout v1.12.0
```
shresthamalik: Do we need to talk about the ./configure options
Say N to  Cuda etc.
And -march=broadwell ?

shresthamalik: May be move option 2 -step 3 to option 3

sayantan-nervana: We might also need to add `https://github.com/tensorflow/tensorflow.git` before the `cd tensor flow`

sayantan-nervana: oh, its not `v` any more, its `r`:
`v1.12.0`->`r1.12`

avijit-nervana: It's embedded in the Python build step. Don't want to talk about it here.

PR #350, opened 2018-12-01
Fix transpose op

kanvi-nervana:
Add error checking for transpose op
Add unit test for transpose op

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -3870,6 +3870,20 @@ static Status TranslateTransposeOp(
   TF_RETURN_IF_ERROR(
       GetStaticInputVector(op, 1, static_input_map, &permutation));
 
+  auto ng_input_rank = ng_input->get_shape().size();
+  std::unordered_set<int> axes;
+  for (auto p : permutation)
+  {
+      if (p < ng_input_rank && !axes.count(p))
+      {
+          axes.insert(p);
+      }
+      else
+      {
+          return errors::InvalidArgument("2 is missing from {0, 1, 1}.");
+      }
+  }
+  
   ng::AxisVector ng_axis_order;
   ng_axis_order.reserve(permutation.size());
 
--- a/test/test_array_ops.cpp
+++ b/test/test_array_ops.cpp
@@ -1106,6 +1106,30 @@ TEST(ArrayOps, Tile) {
   }
 }  // end of test op Tile
 
+// Test op: Transpose
+TEST(ArrayOps, Transpose) {
+  Scope root = Scope::NewRootScope();
+  int dim1 = 2;
+  int dim2 = 3;
+  int dim3 = 2;
+
+  Tensor A(DT_FLOAT, TensorShape({dim1, dim2, dim3}));
+  Tensor perm(DT_INT32, TensorShape({3}));
+  AssignInputValues(A, 7.5f);
+  AssignInputValues(perm, vector<int>{2,1,0});
+
+  vector<int> static_input_indexes = {1};
+  auto R = ops::Transpose(root, A, perm);
+
+  vector<DataType> output_datatypes = {DT_FLOAT};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Transpose", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op Transpose
+
 // Unpacks the given dimension of a rank R tensor into a (R-1) tensor
 TEST(ArrayOps, Unpack) {
   std::vector<std::vector<int64>> input_sizes;


src/ngraph_builder.cc
```diff
@@ -3870,6 +3870,34 @@ static Status TranslateTransposeOp(
...
+      std::ostringstream oss;
+      // Convert all but the last element to avoid a trailing ","
+      std::copy(permutation.begin(), permutation.end() - 1,
+                std::ostream_iterator<int>(oss, ", "));
+
+      // Now add the last element with no delimiter
+      oss << permutation.back();
+
+      return errors::InvalidArgument(missing_axes, " is missing from {",
+                                     oss.str(), "}.");
```
sayantan-nervana: Can we transpose an empty tensor, like a const?
The question being, does TF catch that in construction time or in runtime. If the error is runtime, we should be throwing the error.

src/ngraph_builder.cc
```diff
@@ -3870,6 +3870,34 @@ static Status TranslateTransposeOp(
...
+        }
+      }
+
+      std::ostringstream oss;
+      // Convert all but the last element to avoid a trailing ","
+      std::copy(permutation.begin(), permutation.end() - 1,
+                std::ostream_iterator<int>(oss, ", "));
+
+      // Now add the last element with no delimiter
+      oss << permutation.back();
```
sayantan-nervana: There is an `ngraph::join`: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/util.hpp#L49
`ng::join(permutation)` should work

src/ngraph_builder.cc
```diff
@@ -3870,6 +3870,34 @@ static Status TranslateTransposeOp(
...
+    } else {
+      auto missing_axes = 0;
+      for (int i = 0; i < ng_input_rank; i++) {
+        if (std::find(permutation.begin(), permutation.end(), i) ==
+            permutation.end()) {
+          missing_axes = i;
+          break;
+        }
+      }
+
```
sayantan-nervana: The current implementation involves an `axes.count(p)` and an `std::find` to search for `i` and `insert`.
I think it can be converted to a counting sort like implementation and getting rid of `axes`. That would avoid the finds we are doing.
```
rank...
vector<bool> count(rank, false);
for (auto p : permutation) {
   if (0 <= p <= rank-1) {
        count[p] = true;
   }
   else {
      //Is this even possible. Maybe TF would raise a error if we pass it stuff outside this range?. 
      //If not raise appropriate error
  }
}
for (i -> 0 to rank){
  if (!count[i]){
    push_back to missing  
 }
}
if (missing.size() > 0){
  raise appropriate error.
}
```

The worst case of this is `2*O(n)`. Though the avg case of the `unordered_set` impl is `O(n)`, the worst case is quadratic because of the finds/counts in the loop. But of course `permutation` is a small array, so ...

src/ngraph_builder.cc
shresthamalik: @sayantan-nervana Regarding this, I tried few cases, sharing my observations. 
```
a. transpose a const (permutation is an empty vector)
b. permutation vector contains numbers outside 0 to rank-1 (negative indexing?)
```

a. transpose a const (permutation is an empty vector)
   - with const input and perm vector [2] or [0] or [1] : we get a construction time error
   - with const input and empty perm vector [] : the test passes added a C++ unit test
b. permutation vector contains numbers outside 0 to rank-1 (negative indexing?)
     - if the size of the permutation vector is other than the rank of input : we get a construction error
- if the perm vector is the same size and either a negative value or number>input_rank is one of the perm vector : eg rank =3, perm = [-3,0,1] or [0,1,12] or [-111, 0,1]
 our check detects it as one of the dimension missing from the perm vector . So this is handled
- for some values (while testing negative indexes) i randomly see seg fault in tf itself (much before the execution reaches ngraph),  which we cant do anything about.

shresthamalik:
TESTNOW

test/test_array_ops.cpp
```diff
@@ -1106,6 +1106,47 @@ TEST(ArrayOps, Tile) {
...
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Transpose", static_input_indexes,
+                        output_datatypes, sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of test op Transpose
+
+// Test op: Transpose With Constant input and empty permuation vector
+TEST(ArrayOps, TransposeConstant) {
```
mingshan-wang: Just curious, for this test case, TF and NGraph runs without throwing any errors right? So TF allow passing constant value to Transpose op with no axis order?

sayantan-nervana: So it seems.

shresthamalik: Yes, it allows.

PR #352, opened 2018-12-03
Support backend configuration strings

earhart:


avijit-nervana:
It's actually not a bug. We need to discuss the backend specifying protocol across all backends (as some have a number representing the cardinality and some has other ways). Once we review that,  will come up with the proper implementation.

earhart:
Okay, sounds good -- in the meantime, what's the right way to pass a configuration string to my backend?  The PlaidML backend uses more than just the number of the device -- e.g. right now I'm visualizing my network by setting `NGRAPH_TF_BACKEND=PLAIDML:graphviz=demo.svg`

earhart:
FWIW -- the PlaidML backend does support the notion of using the config string to select a hardware device, but once https://github.com/NervanaSystems/ngraph-tf/pull/330 is landed, users will be able to use `plaidml-setup` instead.  :-)

fboemer:
> It's actually not a bug. We need to discuss the backend specifying protocol across all backends (as some have a number representing the cardinality and some has other ways). Once we review that, will come up with the proper implementation.

HE-transformer would also benefit from this change.
Currently, we use `HE:SEAL:CKKS` and `HE:SEAL:BFV` backend names, which are both exported in `libhe_backend.so`.

PR #346, opened 2018-11-30
Fixed the march flags et. al

avijit-nervana:

PR #354, opened 2018-12-03
Fboemer/fix license location

fboemer:
* License location is incorrect when using ngraph-tf as external project. Was incorrectly fixed in #346
* Typos and whitespace


avijit-nervana:
We fixed this last week in another PR. @fboemer please look at the code in master and see if this PR is still needed.

fboemer:
Yes; the fix in #346  was incorrect (changed the source instead of the destination)

fboemer:
Closed in favor of https://github.com/NervanaSystems/ngraph-tf/pull/359

PR #357, opened 2018-12-04
Langjian/batch norm unittest

jianyinglang:
Fix the difference between tensorflow and ngraph caused by Bessel correction.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -1727,21 +1727,26 @@ static Status TranslateFusedBatchNormOp(
     // This is for Bessel's correction in ng_variance:
     size_t ng_input_size = 1.0;
     size_t ng_scale_size = 1.0;
-    for (size_t i=0; i<ng_input->get_shape().size(); i++) {
-      ng_input_size *= ng_input->get_shape()[i]; 
+    for (size_t i = 0; i < ng_input->get_shape().size(); i++) {
+      ng_input_size *= ng_input->get_shape()[i];
     }
-    for (size_t i=0; i<ng_scale->get_shape().size(); i++) {
-      ng_scale_size *= ng_scale->get_shape()[i]; 
+    for (size_t i = 0; i < ng_scale->get_shape().size(); i++) {
+      ng_scale_size *= ng_scale->get_shape()[i];
     }
-    size_t sample_size = ng_input_size/ng_scale_size;
-    size_t sample_size_minus_one = sample_size > 1.0 ? (sample_size - 1.0) : 1.0;
-    std::vector<size_t> sample_values(ng::shape_size(ng_variance->get_shape()), sample_size);
-    auto sample = std::make_shared<ng::op::Constant>(ng_variance->get_element_type(), ng_variance->get_shape(),
+    size_t sample_size = ng_input_size / ng_scale_size;
+    size_t sample_size_minus_one =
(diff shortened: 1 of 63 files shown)


sayantan-nervana:
@mingshan-wang : could you please test this fix with your PR (where we are moving from relative error to abs error). BN tests should now pass with abs error I think.

src/ngraph_builder.cc
```diff
@@ -1724,12 +1724,29 @@ static Status TranslateFusedBatchNormOp(
     ng_y = make_shared<ng::op::GetOutputElement>(ng_batch_norm, 0);
     ng_mean = make_shared<ng::op::GetOutputElement>(ng_batch_norm, 1);
     ng_variance = make_shared<ng::op::GetOutputElement>(ng_batch_norm, 2);
+    // This is for Bessel's correction in ng_variance:
```
sayantan-nervana: Bessel correction is just `n/(n-1)` multiplied by variance
In batchnorm we compute variance over teh batchsize. So `n` is most probably batchsize.

Could you please provide the link to the code that you are refering to when you make this correction, so that we have a better understanding.

jianyinglang: The `n` is input data size divided by the channel size and not equal to n. For large input data size, this correction is negligible. 
Some of example implementation in TensorFlow can be found here: https://github.com/tensorflow/tensorflow/blob/a8e78e2e617b6ca10f4878fe99fdf43ddedfa7c6/tensorflow/core/kernels/mkl_fused_batch_norm_op.cc#L193

src/ngraph_builder.cc
```diff
@@ -1724,12 +1724,29 @@ static Status TranslateFusedBatchNormOp(
...
+    float ng_input_size = ng::shape_size(ng_input->get_shape());
+    float ng_scale_size = ng::shape_size(ng_scale->get_shape());
+    float sample_size = ng_input_size / ng_scale_size;
```
sayantan-nervana: `ng_input_size` and `ng_scale_size` are integer types. `sample_size` is also guaranteed to be integer type (since `ng_input_size%ng_scale_size == 0`
`sample_size_minus_one` can also be int.

I am think this: instead of introducing a vector division node in ngraph here: `auto Bess_scale = sample / minus_one;`
what is we do this:
```
float bess_scale = (float)(sample_size) / (float)(sample_size - 1);
std::vector<float> bess_scale_vector(ng::shape_size(ng_variance->get_shape()), sample_size);
auto ng_bess_scale = std::make_shared<ng::op::Constant>(ng_variance->get_element_type(), ng_variance->get_shape(), bess_scale_vector);
auto variance = ng_variance * ng_bess_scale;
```
(I suspect ngraph will have some kind of constant folding that would do the division at compile time, but am not sure. In that case the extra division node wouldn't matter much as it would be constant folded)


Other minor suggestions:
1. `ng_variance->get_shape()` and `ng_scale->get_shape()` are both `num_channels`. We can have a single variable and then use that to make it more readable.
2. We are using a `vector<float>` to initialize a `ng_variance->get_element_type()` type `ng::op::Constant`. Usually the type is expected to be `float` but could there be other types (at least other real types, if not integral types). Would they initialize well with `vector<float>` or would it complain?

jianyinglang: Thanks! changed

PR #358, opened 2018-12-04
Sindhu/Fix SplitV segfault 

sindhu-nervana:
Fix segfault during SplitV Op translation seen while running "split_op_test.SplitOpTest.testNonexistentDimTensor"

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -3327,6 +3327,7 @@ static Status TranslateSplitVOp(
     const Node* op, const std::vector<const Tensor*>& static_input_map,
     Builder::OpMap& ng_op_map) {
   shared_ptr<ng::Node> ng_input, ng_length, ng_split_dim;
+
   TF_RETURN_IF_ERROR(GetInputNode(ng_op_map, op, 0, &ng_input));
 
   std::vector<int> lengths;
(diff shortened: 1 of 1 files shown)


src/ngraph_builder.cc
```diff
@@ -3347,7 +3348,18 @@ static Status TranslateSplitVOp(
   std::vector<size_t> lower(rank, 0);
   std::vector<size_t> upper(shape);
 
-  std::vector<int> split_dim_vec;
+  std::vector<int64> split_dim_vec;
+
+  TF_RETURN_IF_ERROR(CheckAxisDimInRange(split_dim_vec, rank));
+
+  // there should be atleast one element specified as axis and not more than one
```
mingshan-wang: very minor: `atlesat` -> `at least`

src/ngraph_builder.cc
```diff
@@ -3347,7 +3348,18 @@ static Status TranslateSplitVOp(
...
+  // as axis is 0-D
+  if (split_dim_vec.size() != 1) {
+    return errors::InvalidArgument(
+        "split_dim_tensor must have "
+        "exactly one element.");
+  }
+
```
shresthamalik: Minor, the axis size check should come before the value check.

sindhu-nervana:
TESTNOW

PR #359, opened 2018-12-05
Fboemer/external project fix

fboemer:
HE-transfomer needs these changes to use the latest ngraph-tf.
* Fixes license install location (as in #354)
* Passes `CMAKE_CXX_COMPILER` flag to external projects. This ensures ng-tf can be built as an external project using a single compiler
* Passes CXX11 ABI flags through to gtest. This is needed when a `GCC >= 5` compiler is used as an external project, but `CXX11 ABI = 0` (for instance, when using `pip install tensorflow` on `ubuntu 16.04`, TF is installed with `CXX11 ABI = 0`)


fboemer:
TESTNOW

aprocter:
I'm not really competent to review extensive build system changes, so I'm going to have to recuse myself.

CMakeLists.txt
```diff
@@ -131,7 +131,9 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     message( STATUS "nGraph-TensorFlow using CXX11 ABI:  ${NGRAPH_TF_CXX11_ABI}" )
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=${NGRAPH_TF_CXX11_ABI}")
+if(DEFINED ${NGRAPH_TF_CX11_ABI})
```
avijit-nervana: This shouldn't be a parameter. This should be auto detected like it's done now.

fboemer: The auto-detection (`set(NGRAPH_TF_CXX11_ABI "${TensorFlow_CXX_ABI}")) is only enabled for GNU compilers. For Clang, this will pass an empty flag to ```-D_GLIBCXX_USE_CXX11_ABI```

fboemer: As discussed, I'll move this to the if('GNU" above

PR #361, opened 2018-12-05
Shrestha/fix tf python kernel tests conv ops test.conv2 d test.test op edge cases

shresthamalik:


shresthamalik:
TESTNOW

src/ngraph_builder.cc
```diff
@@ -1092,6 +1092,22 @@ static Status TranslateConv2DOp(
...
+          op->type_string());
+    }
+  } else {
+    if (tf_strides[0] != 1 || tf_strides[1] != 1) {
+      return errors::InvalidArgument(
+          "Strides in batch and depth dimensions is not supported: ",
+          op->type_string());
+    }
+  }
+
```
sayantan-nervana: Suggesting this (cause the error returned is same in both cases):
```
if (tf_strides[0] != 1 || tf_strides[is_nhwc ? 3 : 1] != 1) {
  return errors::InvalidArgument(
          "Strides in batch and depth dimensions is not supported: ",
          op->type_string());
}
```

PR #368, opened 2018-12-10
Shrestha/fix tf master compilation error and upgrade nGraph version

shresthamalik:
Tested with ng-tf (with tf-master)
Pending test with tf-ng (with tf-master)


shresthamalik:
Tested both ng-tf and tf-ng builds

PR #374, opened 2018-12-13
Sindhu/Translate Rank Op

sindhu-nervana:


test/test_array_ops.cpp
```diff
@@ -512,6 +512,29 @@ TEST(ArrayOps, QuantizeAndDequantizeV2x8xtruexfalse) {
   opexecuter.RunTest();
 }  // end of test op QuantizeAndDequantizeV2x8xtruexfalse
 
+// Test op: Rank Op l
```
mingshan-wang: Minor detail: there is a extra `l` in the end.

test/test_array_ops.cpp
```diff
@@ -512,6 +512,29 @@ TEST(ArrayOps, QuantizeAndDequantizeV2x8xtruexfalse) {
...
+
+  vector<DataType> output_datatypes = {DT_INT32};
+
+  std::vector<Output> sess_run_fetchoutputs = {R};
+  OpExecuter opexecuter(root, "Rank", static_input_indexes, output_datatypes,
+                        sess_run_fetchoutputs);
+
+  opexecuter.RunTest();
+}  // end of RankOp
+
```
sayantan-nervana: can you also please enable/add the TF python tests relating to rank in the .txt file if we have any

PR #378, opened 2018-12-19
Mingshan/rewind DepthToSpace translation to have three reshapes instead of two

mingshan-wang:
Due to the ReshapeSinking and Reshape Elimination performance issue, on the bridge side, we are separating transpose and reshape op instead of combine them into one.


mingshan-wang:
TESTNOW

PR #386, opened 2018-12-29
Enable automated docker builds and testing with build_ngtf.py and test_ngtf.py

crlishka:
Add system to build and test ngraph-tf using the new build_ngtf.py and test_ngtf.py scripts.  Modify test_ngtf.py to produce xUnit/jUnit XML results files for gtest, pytest, and model runs.

Note that I am enabling the new automation system which uses build_ngtf.py _alongside_ the existing system for the following reasons:
- I want to avoid having to do a coordinated merge between two repos (ngraph-tf and internal cje-algo) for builds and CI which is running in production;
- The new build_ngtf.py does not yet support builds in a Python 2 environment.

Once I have the internal cje-algo automation repo updated to use these new build_ngtf.py, and it supports Python 2 builds with this new system, then I will remove the old build system (in a later PR).

Note also that the tensorflow tests in test_ngtf.py are not producing jUnit XML output yet because the **tf_unittest_runner.py** would need to be modified.  I suggest that a Jira issue be opened for this work.

See build #1601 in internal ngraph-tf-unittest job for an example of successful testing in automation, including posting of XML unit-test results to the job page.

I also did a manual run of build_ngtf.py and test_ngtf.py directly on titanxp26 (no docker).  I did not have any problems (other than having to remove my Bazel .cache), and the junit*.xml and xunit*.xml files were properly created.


test_ngtf.py
```diff
@@ -123,6 +126,8 @@ def run_resnet50(build_dir):
 
     call(['git', 'checkout', '4c7b09ad87bbfc4b1f89650bcee40b3fc5e7dfed'])
 
+    junit_script = os.path.abspath('%s/test/ci/junit-wrap.sh' % root_pwd)
```
avijit-nervana: Nice to see that you were able add the junit support easily!

PR #390, opened 2019-01-04
Mingshan/log change

mingshan-wang:
Change the NGRAPH_TF_LOG_PLACEMENT log with more op related summary.


src/ngraph_deassign_clusters.cc
```diff
@@ -78,36 +82,65 @@ static void MaybeLogPlacement(const Graph* graph) {
...
+           return a.second > b.second;
+         });
+
+    for (auto node : v) {
+      if (++deassigned_histogram_counter < deassigned_histogram_size) {
+        std::cout << " " << node.first << " -> " << node.second << ",";
+      } else {
+        std::cout << " " << node.first << " -> " << node.second << std::endl;
+      }
+    }
```
sayantan-nervana: Minor:

I think the `if-else` is to avoid the trailing comma in the end.

`ng::join` can be used here: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/util.hpp#L49

```
vector<string> string_vect(v.size());
transform(v.begin(), v.end(), string_vect.begin(), [](std::pair<string, int>& p){return p.first + " -> " + p.second});
std::cout << ng::join(string_vect)
```

or, some code duplication can be avoided by:
```
for (auto node : v) {
    bool endelem = node != v.back();
    std::cout << " " << node.first << " -> " << node.second << (endelem ? "" : ",");
}
```

src/ngraph_deassign_clusters.cc
```diff
@@ -78,36 +82,65 @@ static void MaybeLogPlacement(const Graph* graph) {
...
     }
   }
 
+  // log the ops gets deassigned
+  int deassigned_histogram_counter = 0;
+  int deassigned_histogram_size = deassigned_histogram.size();
+  std::cout << "NGTF_SUMMARY: Op_deassigned: ";
+  if (deassigned_histogram_size == 0) {
+    std::cout << "None" << std::endl;
+  } else {
```
sayantan-nervana: Could we move the `else` section here (sorting and printing) to a function in `ngraph_utils` with a signature like:
`void print_node_histogram(std::unordered_map<string, int>, bool sorted=true)`

Because I think this will be reused in a couple of other places

src/ngraph_deassign_clusters.cc
```diff
@@ -170,6 +215,8 @@ Status DeassignClusters(Graph* graph) {
         node->ClearAttr("_ngraph_cluster");
         // TODO(amprocte): move attr name to a constant
         node->ClearAttr("_ngraph_marked_for_clustering");
+
+        deassigned_histogram[node->type_string()]++;
```
sayantan-nervana: In our discussion we identified 4 tiers of ops:
1) Rejected in all forms
2) Rejected under certain static conditions:
     2a. confirmationfunction passes
     2b. datatypes match
3) Rejected under some dynamic conditions (resulting from autoclustering) (too small cluster)
4) Accepted

We had decided to print out histograms of (1), (2) and (3).

Below is a list of actionable items from these histograms:
*Histogram of (1)*: This tells us that an op appears a lot in the graph, but we don't support it, so we probably should think of supporting it
*Histogram of (2)*: For an op is "partially" supported, this histogram will remind us that the op is rejected because of the cases we do not cover. For example we only do "scaled" mode in `QuantizeV2` op. If `QuantieV2` starts showing up in this histogram, we will be made aware that we are seeing graphs where our bridge is failing to accept these and we can then decide to cover those gaps
*Histogram of (3)*: A histogram of ops that were declustered. Note that category (1) and (2) are static categories. Category (3) is dynamic, in the sense results can't be predicted till we run the auto-cluster algorithm. Then it is not really the "fault" of the node that it landed up in a cluster of 1 node only and subsequently got declustered. That is knowing that "const" got declustered 5 times and "add" got declustered 10 times is not telling us much (or is it?). Lets discuss tomorrow with @avijit-nervana what this histogram represents and how it can be turned into an actionable item.


Of course the other correction you made in the PR is still valid (relating to `num_nodes_marked_before_deassign`) which now gets rid of the `100%` bug

src/ngraph_deassign_clusters.cc
mingshan-wang: Yes, indeed to get rid of the comma in the end. 

Thanks for the suggestion, I will make the change based on the suggested method.

src/ngraph_deassign_clusters.cc
mingshan-wang: Yes, that's my plan too. I will move to the` ngraph_utils`.

src/ngraph_deassign_clusters.cc
mingshan-wang: Thanks for the detailed explanation!

Regarding the third histogram with the declustered node, it may help us:
1. Make sure important nodes are not declustered. If they are, probably we want to look at the graph to analyse it more. 
2. The information may help us when we do model analysis on the bridge. Then we know which part of the graph split to look at, and maybe cross checking with your edge histogram to make sure there is no other case that prevents the node being clustered with others.

Those are my thoughts. We can discuss more today!

mingshan-wang:
Yes, I will post a snippet here after I have the histogram 1 and 2.

mingshan-wang:
With my change, the log looks like this right now:

```
NGTF_SUMMARY: Op_not_supported:  Enter -> 14, NGraphVariable -> 10, Switch -> 5, Merge -> 5, NextIteration -> 5, _Arg -> 4, Select -> 3, _Retval -> 3, Exit -> 3, TensorArrayV3 -> 2, Range -> 2, TensorArrayReadV3 -> 1, Assert -> 1, TensorArraySizeV3 -> 1, TensorArrayGatherV3 -> 1, TensorArrayWriteV3 -> 1, LoopCond -> 1, TensorArrayScatterV3 -> 1
NGTF_SUMMARY: Op_failed_type_constraint:  Const -> 2
NGTF_SUMMARY: Number of nodes in the graph: 126
NGTF_SUMMARY: Number of nodes marked for clustering: 59 (46% of total nodes)
NGTF_SUMMARY: Number of nodes assigned a cluster: 48 (38% of total nodes)        (81% of nodes marked for clustering)
NGTF_SUMMARY: Number of ngraph clusters :5
NGTF_SUMMARY: Size of nGraph Cluster[14]:       10
NGTF_SUMMARY: Size of nGraph Cluster[15]:       6
NGTF_SUMMARY: Size of nGraph Cluster[16]:       4
NGTF_SUMMARY: Size of nGraph Cluster[21]:       23
NGTF_SUMMARY: Size of nGraph Cluster[29]:       5
NGTF_SUMMARY: Op_deassigned:  Const -> 7, Less -> 2, GreaterEqual -> 1, LogicalAnd -> 1

OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/Select_1 (Select)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/NextIteration_3 (NextIteration)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/Select_2 (Select)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/Select (Select)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/NextIteration_4 (NextIteration)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/TensorArrayWrite/TensorArrayWriteV3 (TensorArrayWriteV3)
OP_placement:   Host    ac_lstm_net_-1/myLSTM/ac_lstm_net_-1/while/NextIteration_2 (NextIteration)
OP_placement:   Host    _arg_ac_lstm_net_-1/Placeholder_0_0 (_Arg)
OP_placement:   Host    _arg_ac_lstm_net_-1/myconv2d/Placeholder_0_1 (_Arg)
OP_placement:   Host    _arg_ac_lstm_net_-1/myconv2d/Placeholder_1_0_2 (_Arg)
OP_placement:   Host    _arg_ac_lstm_net_-1/myconv2d/Placeholder_2_0_3 (_Arg)
...
```

Feel free to give any suggestions.

src/ngraph_mark_for_clustering.cc
```diff
@@ -528,6 +530,7 @@ Status MarkForClustering(Graph* graph) {
       if (!confirmation_constraint_ok) {
         NGRAPH_VLOG(5) << "Node does not meet confirmation constraints: "
                        << node->name();
+        no_support_histogram[node->type_string()]++;
```
sayantan-nervana: `no_support_histogram` smushes together "op absolutely not supported" and "supported under certain conditions"

would it be possible to separate them in `ConfirmationOk` into 2 separate histograms.

That is an extra headache, u may have to split up `ConfirmationOk`

I suggest using `NGraphPlacementRequested` for one check (which is dummy right now) and `ConfirmationOk` for the other

Something like this in pseudo/python code
`NGraphPlacementRequested  = lambda op : op.name in dictionary`
`ConfirmationOk  = lambda op : dictionary[op.name](op)`

mingshan-wang: As we discussed, since `NGraphPlacementRequested` doesn't take confirmation_function_map as an input, instead use a global map to keep track the node that not supported.

PR #396, opened 2019-01-09
Build ngraph-tf not only on Centos

samolisov:
When ngraph-tf is being built on a 64-bits OS another than Centos,
the directory with libraries ('build/artifacts/lib64') could not be
found since CMakeLists.txt and *.cmake files make a reference to 'lib64'
if and only if the host OS is Centos:

if(OS_VERSION STREQUAL "\"centos\"")

But if the host OS is not Centos (SUSE for example), the condition is
always false and the 'lib' directory instead of 'lib64' will be
looked for. To fix the situation the condition is changed to something
similar to:

if(EXISTS ${NGRAPH_INSTALL_DIR}/lib64)

In order to create the right one of the 'build/install/lib64' or
'build/install/lib' directories, the 'GNUInstallDirs' CMake module is
used on the Linux platform (the idea is stolen from NGraph). The
'NGTF_INSTALL_LIB_DIR' option is set to
'${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}':

set(NGTF_INSTALL_LIB_DIR ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})

Signed-off-by: Pavel Samolysov <samolisov@gmail.com>

### DIFF
--- a/python/CreatePipWhl.cmake
+++ b/python/CreatePipWhl.cmake
@@ -36,10 +36,10 @@ if (PYTHON)
     file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python/ngraph_bridge)
 
     # Get the list of libraries we need for the Python pip package
-    # If we are building on CentOS then it's lib64 - else lib
+    # If we are building on a 64-bits OS then it's lib64 - else lib
     set(LIB_SUFFIX lib)
     if(NOT APPLE)
-        if(OS_VERSION STREQUAL "centos")
+        if(EXISTS ${NGTF_INSTALL_DIR}/lib64)
             set(LIB_SUFFIX lib64)
         endif()
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ cmake_minimum_required(VERSION 3.1)
 
 project (ngraph_tensorflow_bridge CXX)
 
+if (UNIX AND NOT APPLE)
+    set(LINUX TRUE)
+endif()
+
 # set directory where the custom finders live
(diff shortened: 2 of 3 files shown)


python/CreatePipWhl.cmake
```diff
@@ -36,10 +36,10 @@ if (PYTHON)
     file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/python/ngraph_bridge)
 
     # Get the list of libraries we need for the Python pip package
-    # If we are building on CentOS then it's lib64 - else lib
+    # If we are building on a 64-bits OS then it's lib64 - else lib
```
avijit-nervana: Is this statement accurate? Ubuntu is also 64bit OS - no?

samolisov: @avijit-nervana I've tested on Open SUSE

avijit-nervana: @samolisov Sure. This sounds like a good rewording.

samolisov: @avijit-nervana Well, since the master has been uploaded to my branch and there are some merging conflicts during rebasing, I've made a new commit to the change.

samolisov:
@avijit-nervana I've updates the comment. Could you please merge the PR if there is no concerns more? I'm working on OpenSUSE and making manual changing after each `git pull` makes me unhappy :-(

samolisov:
@avijit-nervana What can we do to merge the PR? Do I have to sign any contribution agreement or something like this?

SleepProgger:
Same problem on Manjaro.
Creating a link from lib64 to lib works, but i would really appreciate not to have to do this.

PR #397, opened 2019-01-10
Sarkars/edge contraction logging

sayantan-nervana:
it prints:
```
src_encapsulate_id->dst_encapsulate_id: reason
src_encapsulate_id->dst_encapsulate_id: reason
src_encapsulate_id->dst_encapsulate_id: reason
src_encapsulate_id->dst_encapsulate_id: reason

predicates of stuff not merging due to deadness

NGTF_SUMMARY: Ratio of uncontracted edges: x/y = z
NGTF_SUMMARY: UNSUPPORTED: 8, DEADNESS: 1, BACKEND: 0, STATICINPUT: 0, PATHEXISTS: 3
```

Sample output when `NGRAPH_TF_LOG_PLACEMENT=1 ./test/gtest_ngtf --gtest_filter="*DeadnessCheck.livedead1"` is run:

```
=============Edge contraction logs=============
Encapsulates i->j non contraction reason: (Cannot be UNSUPPORTED because unsupported ops will not be assigned an encapsulate)
0->1: DEADNESS
Source predicate: #true Destination predicate: predS:0 Neighbours predicates: #control_flow

NGTF_SUMMARY: Ratio of uncontracted edges: 12/19 = 0.631579
NGTF_SUMMARY: UNSUPPORTED: 8, DEADNESS: 1, BACKEND: 0, STATICINPUT: 0, PATHEXISTS: 3
```

### DIFF
--- a/src/ngraph_assign_clusters.cc
+++ b/src/ngraph_assign_clusters.cc
@@ -466,8 +466,16 @@ Status AssignClusters(Graph* graph) {
       Node* src = edge->src();
       Node* dst = edge->dst();
 
+      if (src->name() == "ConstantFolding/tower_0/v/gradients/tower_0/v/cg/MobilenetV2/Conv_1/Conv2D_grad/ShapeN-matshapes-1"){
+        cout << "\nHEREHEREHERE1 " << dst->name() << "\n";
+      }
+
       if (!src->IsOp() || !dst->IsOp()) {
         continue;
+      }
+
+      if (src->name() == "ConstantFolding/tower_0/v/gradients/tower_0/v/cg/MobilenetV2/Conv_1/Conv2D_grad/ShapeN-matshapes-1"){
+        cout << "\nHEREHEREHERE2 " << dst->name() << "\n";
       }
 
       int src_index = cluster_map[src]->index;
@@ -483,6 +491,10 @@ Status AssignClusters(Graph* graph) {
               .push_back(EdgeNonContractionReasons::UNSUPPORTED);
         }
         continue;
+      }
+
+      if (src->name() == "ConstantFolding/tower_0/v/gradients/tower_0/v/cg/MobilenetV2/Conv_1/Conv2D_grad/ShapeN-matshapes-1"){
+        cout << "\nHEREHEREHERE2.5 " << dst->name() << "\n";
       }
 
 #if !defined(NGRAPH_TF_DISABLE_DEADNESS_CHECK)
@@ -518,6 +530,10 @@ Status AssignClusters(Graph* graph) {
       }
 #endif
 
+if (src->name() == "ConstantFolding/tower_0/v/gradients/tower_0/v/cg/MobilenetV2/Conv_1/Conv2D_grad/ShapeN-matshapes-1"){
+        cout << "\nHEREHEREHERE3 " << dst->name() << "\n";
+      }
+
       // check if the edge can be constracted with respect to backend
       bool is_backend_ok = false;
       TF_RETURN_IF_ERROR(
@@ -535,6 +551,13 @@ Status AssignClusters(Graph* graph) {
         continue;
       }
 
+      if (src->name() == "ConstantFolding/tower_0/v/gradients/tower_0/v/cg/MobilenetV2/Conv_1/Conv2D_grad/ShapeN-matshapes-1"){
+        cout << "\nHEREHEREHERE4 " << dst->name() << "\n";
+        cout << src_index << " " << dst_index << "\n"; //23, 23
+        cout << gc.HasEdge(src_index, dst_index) << "\n";   // 0 (False)
+
+      }
+
       // Check if contracting the edge will lead to cycles
       // if not, MergeClusters
       if (gc.HasEdge(src_index, dst_index) &&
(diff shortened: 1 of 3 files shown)


sayantan-nervana:
Logs look slightly different now. When printing deadness predicates, encapsulate numbers are also printed.

With this command: `NGRAPH_TF_LOG_PLACEMENT=1 ./test/gtest_ngtf --gtest_filter="*DeadnessCheck*G5*"` the logs are:

```
=============Edge contraction logs=============
Encapsulates i->j: non contraction reason (Cannot be UNSUPPORTED because unsupported ops will not be assigned an encapsulate)
2->3: DEADNESS
0->2: DEADNESS
0->3: DEADNESS
Source[0] predicate: #true Destination[2] predicate: (~PredY:0 & PredX:0) Neighbours predicates: (~PredY:0 & PredZ:0 & PredX:0)
Source[2] predicate: (~PredY:0 & PredX:0) Destination[3] predicate: (~PredY:0 & PredZ:0 & PredX:0) Neighbours predicates: 
Source[0] predicate: #true Destination[3] predicate: (~PredY:0 & PredZ:0 & PredX:0) Neighbours predicates: (~PredY:0 & PredX:0)

NGTF_SUMMARY: Ratio of uncontracted edges: 18/29 = 0.62069
NGTF_SUMMARY: UNSUPPORTED: 10, DEADNESS: 3, BACKEND: 0, STATICINPUT: 0, PATHEXISTS: 5
```

src/ngraph_assign_clusters.cc
```diff
@@ -493,14 +543,46 @@ Status AssignClusters(Graph* graph) {
...
+        }
       }
     }
+
+    if (!changed && config::IsLoggingPlacement() &&
+        !collect_non_contracting_edge_info) {
+      changed = true;
+      collect_non_contracting_edge_info = true;
+    }
+    if (collect_non_contracting_edge_info) {
```
mingshan-wang: Is this `if` statement doing anything here?  `collect_non_contracting_edge_info`  seems only = true when the above if looped is entered at the end of the clustering. I guess you could add an assert statement here for `changed` variable.

sayantan-nervana: Ah yes, had a comment there, forgot to add it later :-D

src/ngraph_assign_clusters.cc
```diff
@@ -493,14 +550,51 @@ Status AssignClusters(Graph* graph) {
...
+
+    if (!changed && config::IsLoggingPlacement()) {
+      // This will be entered only once if logging is enabled
+      // When entered, it will force the do-while to run one last time,
+      // collecting information
+      if (!collect_non_contracting_edge_info) {
+        changed = true;
+        collect_non_contracting_edge_info = true;
+      } else {
+        if (changed) {
```
shresthamalik: the outer if loop checks for !changed so is the else -> if(changed) code block required?

sayantan-nervana: This if does not do anything. It just asserts, "you should NOT be in this state"

ie, if collect_non_contracting_edge_info is true, then changed should be false (else raise error)

src/ngraph_assign_clusters.cc
```diff
@@ -559,12 +653,88 @@ Status AssignClusters(Graph* graph) {
...
+      auto find_in_map = [&cluster_to_encapsulate, &cluster_id_vector](int x) {
+        auto itr = cluster_to_encapsulate.find(stoi(cluster_id_vector[x]));
+        return itr == cluster_to_encapsulate.end() ? -1 : itr->second;
+      };
+      int src_encapsulate = find_in_map(0);
+      int dst_encapsulate = find_in_map(1);
+      bool both_src_dst_are_encapsulates =
+          src_encapsulate >= 0 && dst_encapsulate >= 0;
+      bool src_dst_are_distinct = src_encapsulate != dst_encapsulate;
+      if (both_src_dst_are_encapsulates && src_dst_are_distinct) {
```
shresthamalik: If both are encapsulated they will always be distinct? Do we need the src_dst_are_distinct check?

src/ngraph_assign_clusters.cc
```diff
@@ -559,12 +653,88 @@ Status AssignClusters(Graph* graph) {
...
+        << "_ngraph_cluster i->j: non contraction reason histogram (Cannot be "
+           "UNSUPPORTED or NOTANOP because unsupported ops will not be "
+           "assigned an "
+           "encapsulate)\n";
+    for (auto it : cluster_separation_reason) {
+      num_non_contracted += it.second.size();
+      vector<string> reasons_for_this_edge;
+      for (auto& inner_itr : it.second) {
+        reason_count[inner_itr]++;
+        reasons_for_this_edge.push_back(reason_string[inner_itr]);
```
shresthamalik: Regarding the data structure named, reasons_for_this_edge.
From what i understand, there can be only one (known) reason for an edge to not contract, as we exit from the loop as soon as we get a reason. There can be more than one reasons for two clusters to not merge (one reason for each edge between the clusters). If this is correct, then please consider renaming it.

src/ngraph_assign_clusters.cc
shresthamalik: changed will be always false, as the outer if loop checks if(!changed)

sayantan-nervana: ah yes. I see

src/ngraph_encapsulate_clusters.cc
```diff
@@ -256,6 +267,37 @@ Status EncapsulateClusters(Graph* graph) {
...
+    std::cout << "NGTF_SUMMARY: Types of edges:: args: " << count_arg
+              << ", retvals: " << count_retval
+              << ", both arg and retval: " << count_both_arg_retval
+              << ", free: " << count_free
+              << ", encapsulated: " << count_encapsulated
+              << ", total: " << count_tot
+              << ", computed total: " << computed_edge_number << endl;
+    if (!(computed_edge_number == count_tot &&
+          count_tot == graph->num_edges())) {
+      return errors::Internal("Computed number of edges ", computed_edge_number,
```
shresthamalik: Please separate out the check as the error message does not clarify when count_tot!=graph->num_edges

sayantan-nervana: This assert is trying to check a==b==c. So printing out all 3 numbers when the check fails

src/ngraph_assign_clusters.cc
sayantan-nervana: yes it should be `reasons_for_this_separation` as clusters can have multiple reasons to separate, but edges will have only 1 reason (well, they may have 2, but there is an ordering)

src/ngraph_assign_clusters.cc
sayantan-nervana: Yes, we always expect them to be distinct. But if they are ever not, then we will miss counting one edge which will later get caught in the "count total edges vs num_edges()" assert, so we will know.

sayantan-nervana: The endpoints of an edge could lie in the same cluster. Without this I see prints that say "2,2 did not merge because of PATHEXIXTS" etc

PR #399, opened 2019-01-15
Copy ngraph plaidml auto-detect logic

earhart:


earhart:
Done.  N.B. This instruction won't work until the next PlaidML release (any day now).

CMakeLists.txt
```diff
@@ -144,14 +144,16 @@ option(NGRAPH_ARTIFACTS_DIR "Where would nGraph be installed after build" FALSE)
...
+
+find_package(PlaidML CONFIG)
+option(NGRAPH_PLAIDML_ENABLE "Build PlaidML backend" ${PLAIDML_FOUND})
 
 # Validate the options
 if (NGRAPH_PLAIDML_ENABLE)
-    if (NOT DEFINED PlaidML_DIR)
-        message(FATAL_ERROR "Provide the PlaidML SDK directory: -DPlaidML_DIR=<directory>")
+    if (NOT PLAIDML_FOUND)
+        message(FATAL_ERROR "Unable to find PlaidML; use \"pip install plaidml\" to install it.")
```
avijit-nervana: Does this mean we won't be able to build without PlaidML? If so then this needs to be made optional like other options (e.g., `NGRAPH_DISTRIBUTED_ENABLE`).

Then you need to add a command line option for `build_ngtf.py` to enable this based on command line option (see the how the `distributed_build ` option is handled by this script for example)

earhart: Nope :-). The default here for `NGRAPH_PLAIDML_ENABLE` is the value of `PLAIDML_FOUND`.  cmake searches for the PlaidML installation, and if it doesn't find it, `PLAIDML_FOUND` will be false, and the build will proceed normally.  But if the user explicitly sets `NGRAPH_PLAIDML_ENABLE`, it'll override the option default; if `PLAIDML_FOUND` is false in that case, it's a fatal error -- so if the user explicitly asks for it, and it's not there, the build fails.

(This is basically an exact copy of the code that's checked into ngraph.)

PR #413, opened 2019-01-28
Shrestha/add control dependencies in ngraph cluster

shresthamalik:
Add control dependency to the translated nGraph nodes. first_ng_op_map stores the entry point nGraph nodes and ng_op_map stores the exit point nGraph nodes for each Tf translated op. After the TF graph has been translated to nGraph, we add the control dependencies to the nGraph nodes using the stored entry and exit points.

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -459,12 +459,20 @@ static Status TranslateAddNOp(
     TF_RETURN_IF_ERROR(
         GetInputNode(ng_op_map, op, inp_idx, &ng_arg_vec[inp_idx]));
 
-  // TODO (malikshr): find out the first node here
+  // TODO (malikshr): Is empty list a valid input?
+  // TODO (malikshr) : No op is created
+  if (ng_arg_vec.size() == 1) {
+    SaveNgOp(ng_op_map, op->name(), ng_arg_vec[0]);
+    return Status::OK();
+  }
+
+  auto ng_first_add = ng_arg_vec[0] + ng_arg_vec[1];
+  SaveNgOp(first_ng_op_map, op->name(), ng_first_add);
   SaveNgOp(ng_op_map, op->name(),
-           std::accumulate(std::next(ng_arg_vec.begin()), ng_arg_vec.end(),
-                           ng_arg_vec.at(0)));  // accumulation: start with
-                                                // first element. default op is
-                                                // addition
+           std::accumulate(ng_arg_vec.begin() + 2, ng_arg_vec.end(),
+                           ng_first_add));  // accumulation: start
+                                            // with first element.
+                                            // default op is addition
   return Status::OK();
 }
 
@@ -777,19 +785,27 @@ static Status TranslateBatchMatMulOp(
 
   auto ng_lhs_axes = out_axes;
   auto ng_rhs_axes = out_axes;
+  bool got_first_node = false;
+
   if (tf_adj_x) {
     ng_lhs_axes.push_back(n_dims - 1);
     ng_lhs_axes.push_back(n_dims - 2);
     ng_lhs = ng::builder::numpy_transpose(ng_lhs, ng_lhs_axes);
+    got_first_node = true;
+    SaveNgOp(first_ng_op_map, op->name(), ng_lhs);
   }
   if (tf_adj_y) {
     ng_rhs_axes.insert(ng_rhs_axes.begin(), n_dims - 2);
     ng_rhs_axes.insert(ng_rhs_axes.begin(), n_dims - 1);
     ng_rhs = ng::builder::numpy_transpose(ng_rhs, ng_rhs_axes);
+    got_first_node = true;
+    SaveNgOp(first_ng_op_map, op->name(), ng_rhs);
   } else {
     ng_rhs_axes.insert(ng_rhs_axes.begin(), n_dims - 1);
     ng_rhs_axes.insert(ng_rhs_axes.begin(), n_dims - 2);
     ng_rhs = ng::builder::numpy_transpose(ng_rhs, ng_rhs_axes);
+    got_first_node = true;
+    SaveNgOp(first_ng_op_map, op->name(), ng_rhs);
   }
 
   ng_lhs_shape = ng_lhs->get_shape();
@@ -802,15 +818,18 @@ static Status TranslateBatchMatMulOp(
   }
   if (n_dims == 2) {
     auto dot_output = make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs);
-    SaveNgOp(first_ng_op_map, op->name(), dot_output);
+    if (!got_first_node) {
+      SaveNgOp(first_ng_op_map, op->name(), dot_output);
+    }
     SaveNgOp(ng_op_map, op->name(), dot_output);
 
   } else {
     auto output_shape = ng_lhs_shape;
     output_shape[n_dims - 1] = ng_rhs_shape[1];
     auto dot_output = make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs);
-    SaveNgOp(first_ng_op_map, op->name(), dot_output);
-
+    if (!got_first_node) {
+      SaveNgOp(first_ng_op_map, op->name(), dot_output);
+    }
     size_t compound_size = 1;
     for (int i = 0; i < out_axes.size(); i++) {
       compound_size *= output_shape[i];
@@ -1106,9 +1125,8 @@ static Status TranslateConv2DOp(
   std::shared_ptr<ng::Node> ng_conv = make_shared<ng::op::Convolution>(
       ng_input, ng_filter, ng_strides, ng_dilations, ng_padding_below,
       ng_padding_above);
-  if (!is_nhwc) {
-    SaveNgOp(first_ng_op_map, op->name(), ng_conv);
-  }
+
+  SaveNgOp(first_ng_op_map, op->name(), ng_conv);
 
   BatchToTensorflow(is_nhwc, ng_conv);
   SaveNgOp(ng_op_map, op->name(), ng_conv);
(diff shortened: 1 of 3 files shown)


src/ngraph_builder.cc
```diff
@@ -1516,6 +1580,7 @@ static Status TranslateDepthToSpaceOp(
   std::iota(ng_axis_order.begin(), ng_axis_order.end(), 0);
   auto reshaped =
       make_shared<ng::op::Reshape>(ng_input, ng_axis_order, ng_reshape_shape);
+  SaveNgOp(first_ng_op_map, op->name(), reshaped);
 
   auto transposed =
```
mingshan-wang: Question: `ng::builder::numpy_transpose` seems also doing something and at the end create a ngraph reshape op, in this case, should we also consider it as an input? 
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/builder/numpy_transpose.cpp#L77

src/ngraph_builder.cc
```diff
@@ -1081,14 +1126,16 @@ static Status TranslateConv2DOp(
       ng_input, ng_filter, ng_strides, ng_dilations, ng_padding_below,
       ng_padding_above);
 
+  SaveNgOp(first_ng_op_map, op->name(), ng_conv);
```
kanvi-nervana: Did we not save the entry nodes already here: "SaveNgOp(first_ng_op_map, op->name(), ng_filter);" ?

src/ngraph_builder.cc
```diff
@@ -345,13 +345,14 @@ Builder::TF_NGRAPH_CONST_MAP() {
 //  }
 static Status TranslateUnaryOp(
     const Node* op, const std::vector<const Tensor*>& static_input_map,
-    Builder::OpMap& ng_op_map,
+    Builder::OpMap& ng_op_map, Builder::OpMap& first_ng_op_map,
```
kanvi-nervana: Can we rename this to something like entry_nodes_ng_op_map and similary ng_op_map to exit_nodes_ng_op_map to make it more readable?

src/ngraph_builder.cc
shresthamalik: no, the reshape by numpy_transpose is part of the translate function so not an input.

mingshan-wang: I see. Thanks for the confirmation!

shresthamalik: If the reshape is the entry node, then must be saved

src/ngraph_builder.cc
shresthamalik: will check

src/ngraph_builder.cc
```diff
@@ -1161,6 +1211,9 @@ static Status TranslateConv2DBackpropFilterOp(
                       static_cast<unsigned int>(tf_filter_sizes[0]),
                       static_cast<unsigned int>(tf_filter_sizes[1])};
   BatchToNGraph(is_nhwc, ng_output_delta);
+  if (is_nhwc) {
```
kanvi-nervana: Q: We have already taken care of the case for "is_nhwc" here: https://github.com/NervanaSystems/ngraph-tf/pull/413/files?utf8=%E2%9C%93&diff=unified&w=1#diff-148e2a1788305495ad1a483e17b53eaaR1203
Why are we doing it again?

src/ngraph_builder.cc
```diff
@@ -1293,6 +1350,8 @@ static Status TranslateConv2DBackpropInputOp(
           ng_padding_below, ng_padding_above,
           ng::Strides(ng_batch_shape.size() - 2, 1));
 
+  SaveNgOp(first_ng_op_map, op->name(), ng_data);
```
kanvi-nervana: Q: Did we not save the entry nodes here: https://github.com/NervanaSystems/ngraph-tf/pull/413/files?utf8=%E2%9C%93&diff=unified&w=1#diff-148e2a1788305495ad1a483e17b53eaaR1336

src/ngraph_builder.cc
```diff
@@ -1368,7 +1430,9 @@ static Status TranslateConv3DOp(
   std::shared_ptr<ng::Node> ng_conv = make_shared<ng::op::Convolution>(
       ng_input, ng_filter, ng_strides, ng_dilations, ng_padding_below,
       ng_padding_above);
-
+  if (!is_ndhwc) {
+    SaveNgOp(first_ng_op_map, op->name(), ng_conv);
+  }
```
kanvi-nervana: Q: I understand the the separate cases for _"is_ndhwc"_ and _"!is_ndhwc"_ , but why do we need a third one here: https://github.com/NervanaSystems/ngraph-tf/pull/413/files?utf8=%E2%9C%93&diff=unified&w=1#diff-148e2a1788305495ad1a483e17b53eaaR1419

src/ngraph_builder.cc
```diff
@@ -1597,7 +1665,9 @@ static Status TranslateDepthwiseConv2dNativeOp(
                                           input_shape[3]};
     auto ng_sliced_input =
         make_shared<ng::op::Slice>(ng_input, lower_bound, upper_bound);
-
+    if (!is_nhwc) {
+      SaveNgOp(first_ng_op_map, op->name(), ng_input);
+    }
```
kanvi-nervana: same as above

src/ngraph_builder.cc
mingshan-wang: I think `entry_nodes_ng_op_map` and `output_nodes_ng_op_map` make sense to me!

sindhu-nervana: outputs sound better than exit.

kanvi-nervana: In that case input_nodes_ng_op_map and output_nodes_ng_op_map makes more sense if you  don't want to change all the comments etc.

src/ngraph_builder.cc
shresthamalik: Yes, corrected.

src/ngraph_builder.cc
shresthamalik: They (ng_data_batch and ng_output_delta) are different data flows/ inputs so both of them will be entry nodes.

src/ngraph_builder.cc
shresthamalik: Yes, correct.

src/ngraph_builder.cc
```diff
@@ -785,12 +817,19 @@ static Status TranslateBatchMatMulOp(
...
+    SaveNgOp(ng_op_map, op->name(), dot_output);
+
   } else {
     auto output_shape = ng_lhs_shape;
     output_shape[n_dims - 1] = ng_rhs_shape[1];
     auto dot_output = make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs);
+    if (!got_first_node) {
+      SaveNgOp(first_ng_op_map, op->name(), dot_output);
+    }
```
sindhu-nervana: Q - Why do we have a boolean here to check if it is first node or not unlike other ops?

Can't we just do this for Saving the map in line #822, is there a reason in assigning it to dot_output?
```
SaveNgOp(first_ng_op_map, op->name(),
             make_shared<ngraph::op::Dot>(ng_lhs, ng_rhs));
```

src/ngraph_builder.cc
shresthamalik: I had actually removed all of them. But i guess i missed saving them. Thanks for pointing them out.

src/ngraph_builder.cc
shresthamalik: ng_filter and ng_input are different data flows, so this check I think is still valid

src/ngraph_builder.cc
shresthamalik: The boolean is to check if ng_lhs and ng_rhs (the input nodes) have been modified as they might be in few cases (when tf_adjx is true) etc. If they have been modified they are the entry nodes.

The boolean makes the check simpler otherwise we need to check for many conditions

src/ngraph_builder.cc
```diff
@@ -1597,7 +1665,9 @@ static Status TranslateDepthwiseConv2dNativeOp(
                                           input_shape[3]};
     auto ng_sliced_input =
         make_shared<ng::op::Slice>(ng_input, lower_bound, upper_bound);
-
+    if (!is_nhwc) {
+      SaveNgOp(first_ng_op_map, op->name(), ng_input);
```
mingshan-wang: Hi @shresthamalik, this line 1669  `SaveNgOp(first_ng_op_map, op->name(), ng_input);` seems collide with line 1635  `SaveNgOp(first_ng_op_map, op->name(), ng_input);`, so no matter is_nhwc or not, we are saving this ng_input op.
Wondering here is this `ng_input` to be `ng_sliced_input`?

src/ngraph_builder.cc
sindhu-nervana: hmm, ok. thanks for the explanation

src/ngraph_builder.cc
sindhu-nervana: same question as above here.

shresthamalik: should be ng_sliced_input. Thanks!

src/ngraph_builder.cc
shresthamalik: input_nodes_ng_op_map might be confused with the actual input ng nodes to the node being translated, which is not what we are looking at. We are looking at the first nodes we create as part of this TF node's translation

kanvi-nervana: entry/exit , input/output make sense but we can do _entry_nodes_ng_op_map_ and _output_nodes_ng_op_map_

src/ngraph_builder.cc
```diff
@@ -422,53 +443,69 @@ static Status TranslateBinaryOp(
 //
 //  if (n->type_string == "Add") {
 //    TF_RETURN_IF_ERROR(TranslateBinaryOp<ng::op::Add>(op, static_input_map,
-//    ng_op_map));
+//    output_nodes_ng_op_map,entry_nodes_ng_op_map));
```
```suggestion
//    output_nodes_ng_op_map, entry_nodes_ng_op_map));
```

PR #415, opened 2019-01-30
Add some more detailed noncontraction logging

aprocter:
This adds some more detail to the noncontraction logging, and incorporates a fix that @sayantan-nervana discovered for a corner-case that was causing `Const`-source noncontractions to be logged incorrectly.

If you run with `NGRAPH_TF_LOG_PLACEMENT=1` you will now see log lines like the following:

```
2019-01-29 19:03:54.915250: I /localdisk/amprocte/ngraph-tf/src/ngraph_assign_clusters.cc:475] NONCONTRACTION: STATICINPUT: transformer/attention_bias_lower_triangle/attention_bias_local/ones/packed<Pack>[0] -> transformer/attention_bias_lower_triangle/attention_bias_local/ones<Fill>[0]
2019-01-29 19:03:54.915256: I /localdisk/amprocte/ngraph-tf/src/ngraph_assign_clusters.cc:475] NONCONTRACTION: SAMECLUSTER: transformer/ones/Const<Const>[0] -> transformer/attention_bias_lower_triangle/attention_bias_local/ones<Fill>[1]
2019-01-29 19:03:54.915261: I /localdisk/amprocte/ngraph-tf/src/ngraph_assign_clusters.cc:475] NONCONTRACTION: UNSUPPORTED: transformer/zeros_19/Const<Const>[0] -> transformer/get_timing_signal_1d/range<Range>[0]
```

These indicate the noncontraction reason, source node/op/output, and destination node/op/input.


sayantan-nervana:
Documenting the const corner case:

C(c0)--->*X(c1)
|
|
v
Y(c0)-->U-->Z(c1)

Const C and node Y are in encapsulate c0, node X and Z are in encapsulate c1 and U is a TF node that is unsupported.

In this case the reason for C->X not merging should be "path exists" rather than "static input"

This PR fixes that.

PR #416, opened 2019-01-31
use -j=N to run N jobs in parallel

cepera:
hi,

Recently faced an issue that i didn't have enough memory on my machine to build TF+Ngraph using all the threads. I assume these changes would be useful for others too.


crlishka:
TESTNOW

PR #420, opened 2019-02-05
Sarkars/unify pooling translate functions

sayantan-nervana:


src/ngraph_builder.cc
```diff
@@ -648,15 +655,15 @@ static Status TranslateAvgPoolOp(
   Builder::MakePadding(tf_padding_type, ng_image_shape, ng_kernel_shape,
                        ng_strides, ng_padding_below, ng_padding_above);
 
-  std::shared_ptr<ng::Node> ng_avgpool =
-      make_shared<ng::op::AvgPool>(ng_input, ng_kernel_shape, ng_strides,
-                                   ng_padding_below, ng_padding_above, false);
+  std::shared_ptr<ng::Node> ng_pool =
```
mingshan-wang: Question:
Does TF AvgPool considers padding in average calculation? Just want to confirm here.

sayantan-nervana: Avg pool has this extra param whose default value is false:
https://github.com/NervanaSystems/ngraph/blob/066037c25901478ac79ead90aa6da721ab27c0c3/src/ngraph/op/avg_pool.hpp#L51

Earlier we were explicitly passing the `false` value, but even without it, the default value is going to be `false`

mingshan-wang: Cool! Sounds good!

PR #421, opened 2019-02-05
installation on current python environment

ami-GS:


crlishka:
TESTNOW

ami-GS:
I could run this with specific `ngraph_version`, let me debug this to run as it is.

PR #428, opened 2019-02-12
Langjian/cache evict

jianyinglang:
This branch does not compile with `v0.12` nGraph.


src/ngraph_encapsulate_op.cc
```diff
@@ -262,11 +287,16 @@ class NGraphEncapsulateOp : public OpKernel {
...
     if (it == m_ng_functions.end()) {
+      // Measure the total memory here first
+      mem_usage(vm0, rss0);
+
       NGRAPH_VLOG(1) << "Compilation cache miss: " << ctx->op_kernel().name();
       OP_REQUIRES_OK(
           ctx, Builder::TranslateGraph(input_shapes, static_input_map, &m_graph,
                                        ng_function));
 
+      auto function_size = ng_function->get_graph_size() / 1024.0;  // kb unit
```
avijit-nervana: So - it seems that the function size is pretty useless at this point as it really points to constants that (if) are part of that function parameter. On the other hand - the memory used by the process is really the useful one as that reflects the correct amount of memory consumed by this process - majority of which is from the tensors they require. We will have to think through this more. Fine for now.

src/ngraph_encapsulate_op.cc
```diff
@@ -212,6 +213,29 @@ class NGraphEncapsulateOp : public OpKernel {
     return Status::OK();
   }
 
+  void mem_usage(double& vm_usage, double& resident_set) {
```
avijit-nervana: Move this to some utility file such as ngraph_utils.h. Also follow the proper naming convention.

build_ngtf.py
```diff
@@ -432,7 +432,7 @@ def main():
     #-------------------------------
 
     # Component versions
-    ngraph_version = "v0.12.0"
+    ngraph_version = "v0.14.0-rc.0" #"484c0a0de9ae00aa13c7ecdcb3b33a136970c1c1" #"v0.12.0-rc.1"
```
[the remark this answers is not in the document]
jianyinglang: changed

CMakeLists.txt
```diff
@@ -239,7 +239,7 @@ if (NOT USE_PRE_BUILT_NGRAPH)
     ExternalProject_Add(
         ext_ngraph
         GIT_REPOSITORY https://github.com/NervanaSystems/ngraph
-        GIT_TAG v0.12.0
+        GIT_TAG v0.14.0-rc.0 #484c0a0de9ae00aa13c7ecdcb3b33a136970c1c1
```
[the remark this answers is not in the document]
jianyinglang: changed

src/ngraph_encapsulate_op.cc
```diff
@@ -212,6 +213,29 @@ class NGraphEncapsulateOp : public OpKernel {
...
+      std::string mem_in;
+      getline(ifs, mem_in);
+      vector<string> mem_str = ng::split(mem_in, ' ');
+      vsize = std::stol(mem_str[22]);
+      rss = std::stol(mem_str[23]);
+    }
+
+    long page_size_kb = sysconf(_SC_PAGE_SIZE) /
+                        1024;   // in case x86-64 is configured to use 2MB pages
+    vm_usage = vsize / 1024.0;  // unit kb
```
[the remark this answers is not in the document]
jianyinglang: changed

src/ngraph_encapsulate_op.cc
```diff
@@ -222,6 +246,7 @@ class NGraphEncapsulateOp : public OpKernel {
     ng::runtime::Backend* op_backend =
         BackendManager::GetBackend(m_op_backend_name);
 
+    double vm, rss, vm0, rss0;
```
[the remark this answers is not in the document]
jianyinglang: changed

src/ngraph_encapsulate_op.cc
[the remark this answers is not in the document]
jianyinglang: changed

src/ngraph_encapsulate_op.cc
```diff
@@ -282,9 +312,39 @@ class NGraphEncapsulateOp : public OpKernel {
...
 #endif
       }
-
+      // Evict the cache if the number of elements exceeds 16
+      const char* cache_depth_specified =
+          std::getenv("NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH");
+      if (cache_depth_specified != nullptr) {
+        NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH = atoi(cache_depth_specified);
+      }
+      if (m_ng_functions.size() > NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH) {
```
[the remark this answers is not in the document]
jianyinglang: changed

src/ngraph_encapsulate_op.cc
jianyinglang: Leave as is for now

src/ngraph_encapsulate_op.cc
jianyinglang: changed to `MemoryProfile` and moved to ngraph_utils.h and ngraph_utils.cc

src/ngraph_encapsulate_op.cc
```diff
@@ -282,9 +289,39 @@ class NGraphEncapsulateOp : public OpKernel {
                         ng_function);
 #endif
       }
-
+      // Evict the cache if the number of elements exceeds 16
+      const char* cache_depth_specified =
+          std::getenv("NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH");
+      if (cache_depth_specified != nullptr) {
+        NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH = atoi(cache_depth_specified);
```
sayantan-nervana: `atoi` might be unsafe, for example it might not throw any errors when the user mischievously sets `NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH=A_POTATO`

`stoi` might be slight safer, so we could convert the `char*` to a `string`.

However I am not exactly sure about `atoi` and `stoi`'s susceptibility to the `A_POTATO` attack. We might consider implementing a `is_valid_int` function to go through each char, and verify its a valid ASCII for an int. If this passes only then we call `atoi`

sayantan-nervana: We are reading `NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH`everytime in the `Compute` function. We should move this to some place like backend manager, or atleast to the constructor of the encapsulate op.

However if the intention is that we are able to change the `NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH` while the graph is running, then we can keep it here.

src/ngraph_encapsulate_op.cc
```diff
@@ -541,6 +579,8 @@ class NGraphEncapsulateOp : public OpKernel {
   std::vector<bool> m_input_is_static;
   std::mutex m_compute_lock;
   string m_op_backend_name;
+  std::list<std::string> LRU;
+  int NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH = 16;
```
sayantan-nervana: We could move `const char* cache_depth_specified = std::getenv("NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH");` here
(but then we wont be able to tweak cache size while the model is running).

Pros and Cons for doing it in `Compute` vs in the constructor?

src/ngraph_utils.cc
```diff
@@ -242,6 +242,26 @@ void NgraphSerialize(const std::string& file_name,
...
+  //Get the two fields we want
+  long vsize;
+  long rss;
+
+  std::ifstream ifs("/proc/self/stat", std::ios_base::in);
+  std::string mem_in; 
+  getline(ifs, mem_in);
+  vector<string> mem_str = ng::split(mem_in, ' ');
+  vsize = std::stol(mem_str[22]);
+  rss = std::stol(mem_str[23]);
```
sayantan-nervana: safe to use `stol` here, because `/proc/self/stat` will conceivably not use the `A_POTATO` attack

src/ngraph_encapsulate_op.cc
shresthamalik: Please change the name  of the data members (For e.g. m_cache_depth, m_lru_function) following the same convention as other data members.

src/ngraph_encapsulate_op.cc
```diff
@@ -34,6 +34,7 @@
 #include "ngraph_mark_for_clustering.h"
 #include "ngraph_utils.h"
 
+#include "ngraph/runtime/cpu/cpu_backend.hpp"
 #include "ngraph/runtime/interpreter/int_backend.hpp"
```
shresthamalik: Is it possible to include backend.hpp than the specific backend files?
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/backend.hpp

I haven't investigated whether there is a dependency or not.

src/ngraph_encapsulate_op.cc
sayantan-nervana: Marking this resolved, as discussed

src/ngraph_encapsulate_op.cc
jianyinglang: changed the name of LRU to m_lru and cache_depth_specified to m_cache_depth_specified

src/ngraph_encapsulate_op.cc
```diff
@@ -222,6 +223,7 @@ class NGraphEncapsulateOp : public OpKernel {
     ng::runtime::Backend* op_backend =
         BackendManager::GetBackend(m_op_backend_name);
 
+    long vm, rss, vm0, rss0;
```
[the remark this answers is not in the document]
jianyinglang: moved this before use

src/ngraph_encapsulate_op.cc
jianyinglang: changed to general backend

src/ngraph_encapsulate_op.cc
shresthamalik: I meant this variable, as this is the data member of the class. Please change this to m_cache_depth 
```
int NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH
```

src/ngraph_encapsulate_op.cc
```diff
@@ -34,6 +34,7 @@
 #include "ngraph_mark_for_clustering.h"
 #include "ngraph_utils.h"
 
+#include "ngraph/runtime/backend.hpp"
 #include "ngraph/runtime/interpreter/int_backend.hpp"
```
avijit-nervana: In that case `#include "ngraph/runtime/interpreter/int_backend.hpp"` is not needed any more?

PR #430, opened 2019-02-14
Sindhu/Upgrade to ngraph master

sindhu-nervana:


src/ngraph_encapsulate_op.cc
```diff
@@ -319,7 +319,8 @@ class NGraphEncapsulateOp : public OpKernel {
         m_ng_functions.erase(m_lru.back());
 
         // Call delete function here pf he erased func
-        op_backend->remove_compiled_function(evicted_ng_function);
+        auto exec = op_backend->compile(evicted_ng_function);
```
avijit-nervana: Why do we need to compile to remove the function? Talk to the nGraph folks and see if there is an equivalent method.

src/ngraph_encapsulate_op.cc
```diff
@@ -515,9 +516,8 @@ class NGraphEncapsulateOp : public OpKernel {
           << "NGraphEncapsulateOp::Compute call starting for cluster "
           << m_ngraph_cluster;
       try {
-        op_backend->call(op_backend->compile(ng_function), ng_outputs,
-                         ng_inputs);
-
+        auto exec = op_backend->compile(ng_function);
```
avijit-nervana: Find out if we can save the compiled function instead of compiling every time we call.

PR #433, opened 2019-02-15
Fix memory leak

avijit-nervana:
Added code to removed unneeded tensors after the execution of a cluster is done.

### DIFF
--- a/src/ngraph_encapsulate_op.cc
+++ b/src/ngraph_encapsulate_op.cc
@@ -63,14 +63,17 @@ REGISTER_OP("NGraphEncapsulate")
 
 class NGraphEncapsulateOp : public OpKernel {
  public:
+  //---------------------------------------------------------------------------
+  //  NGraphEncapsulateOp::ctor
+  //---------------------------------------------------------------------------
   explicit NGraphEncapsulateOp(OpKernelConstruction* ctx)
       : OpKernel(ctx),
         m_graph(OpRegistry::Global()),
         m_freshness_tracker(nullptr) {
-    my_instance_id_ = instance_id_;
-    instance_id_++;
+    my_instance_id = s_instance_count;
+    s_instance_count++;
 
-    NGRAPH_VLOG(1) << "NGraphEncapsulateOp: " << my_instance_id_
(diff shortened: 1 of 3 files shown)


src/ngraph_encapsulate_op.cc
```diff
@@ -68,7 +67,14 @@ class NGraphEncapsulateOp : public OpKernel {
...
         m_graph(OpRegistry::Global()),
         m_freshness_tracker(nullptr) {
-    NGRAPH_VLOG(2) << "NGraphEncapsulateOp: Name: " << name();
+    my_instance_id_ = instance_id_;
+    instance_id_++;
+
+    NGRAPH_VLOG(2) << "NGraphEncapsulateOp: " << my_instance_id_
+                   << " Name: " << name();
+
+    std::cout << "NGraphEncapsulateOp: " << my_instance_id_
```
shresthamalik: Please remove the cout.

src/ngraph_encapsulate_op.cc
```diff
@@ -582,8 +630,12 @@ class NGraphEncapsulateOp : public OpKernel {
   string m_op_backend_name;
   std::list<std::string> m_lru;
   int NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH = 16;
+  static int instance_id_;
+  int my_instance_id_{0};
```
shresthamalik: Should we rename my_instance_id_ to m_instance_id, keeping with the other member variables?

Same thing for the member variable NGRAPH_TF_FUNCTION_CACHE_ITEM_DEPTH, should it be named to m_cache_depth?

src/ngraph_encapsulate_op.cc
```diff
@@ -213,6 +219,29 @@ class NGraphEncapsulateOp : public OpKernel {
     return Status::OK();
   }
 
+  void mem_usage(double& vm_usage, double& resident_set) {
```
shresthamalik: We can remove this function. It was moved to ngraph_utils.h and renamed to MemoryProfile. 

https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_utils.cc#L247

PR #426, opened 2019-02-08
Support non-const min/max in quantized conv ops

jbobba:

PR #440, opened 2019-02-22
Allow non-const min/maxes for quantize and dequantize ops

jbobba:
Builds off #426.


mingshan-wang:
Verified by running quantized inceptionv3 and resnet50 inference with real data, the final accuracy looks good. 
But found out if setting BS=128, running quantized resnet50 inference SegFault, setting BS=56 works. May need to further debugging for this issue.

PR #445, opened 2019-03-03
avijit/Add diagnostics for debugging AllReduce ops

avijit-nervana:
* Setting the `friendly_name` of the nGraph AllReduce op to the TensorFlow op name for easy correlation in the debug logs
* Enabled printing the order of TensorFlow Horovod ops at the TensorFlow level to help debug out of order execution

### DIFF
--- a/build_ngtf.py
+++ b/build_ngtf.py
@@ -432,7 +432,7 @@ def main():
     #-------------------------------
 
     # Component versions
-    ngraph_version = "d3453447883fe48e63126316d5da72fcbd47dded"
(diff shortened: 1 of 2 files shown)


build_ngtf.py
```diff
@@ -432,7 +432,7 @@ def main():
     #-------------------------------
 
     # Component versions
-    ngraph_version = "d3453447883fe48e63126316d5da72fcbd47dded"
+    ngraph_version = "avijit/add_diag_4_distributed_debugging"
```
shresthamalik: We dont want this ngraph_version merged

PR #448, opened 2019-03-07
update distributed cmake flags

sasadep:
updating the cmake as an effect of this https://github.com/NervanaSystems/ngraph/pull/2512

to be merged only after this ngraph PR is merged 

these changes are **NOT** reflected on **v0.15.0**
but will be part of **v0.16**


sasadep:
these changes allow the user to use and modify just a single cmake flag

sasadep:
@avijit-nervana @jianyinglang  I have addressed the issue of passing flag. Let me know if any changes needed

PR #450, opened 2019-03-09
Fixed the broken unit test for GPU (and possibly other non CPU backends)

avijit-nervana:


test/opexecuter.cpp
```diff
@@ -89,17 +89,15 @@ void OpExecuter::ValidateGraph(const Graph& graph,
...
-  Node* test_op;
   for (Node* node : graph.nodes()) {
     if (node->IsSource() || node->IsSink()) {
       continue;
     } else if (node->type_string() == test_op_type_) {
       // only one node of type test_op
       ASSERT_FALSE(found_test_op) << "Only one op of type " << test_op_type_
                                   << " should exist in the graph. Found nodes "
-                                  << node->name() << " and " << test_op->name();
+                                  << node->name() << " and " << node->name();
```
mingshan-wang: The original version will print the duplicated node names, but after the change it will print one of those node names twice. If we are removing the `test_op`, we probably should rephrase this message to be sth like "Found duplicated type nodes for"<< node->name();

avijit-nervana: Yes - you are right. We are already printing the node->name() - so not sure why I didn't delete the second occurrence. The real issue with the original was an error of attempting to read a variable which may not have been initialized (which would be the case if the node type is a test op). This was genuinely flagged by the compiler warning - so I was removing this. Will clean this up.

test/opexecuter.cpp
```diff
@@ -89,17 +89,16 @@ void OpExecuter::ValidateGraph(const Graph& graph,
...
       // only one node of type test_op
-      ASSERT_FALSE(found_test_op) << "Only one op of type " << test_op_type_
-                                  << " should exist in the graph. Found nodes "
-                                  << node->name() << " and " << test_op->name();
+      ASSERT_FALSE(found_test_op)
+          << "Only one op of type " << test_op_type_
+          << " should exist in the graph. Found duplicate node: "
+          << node->name();
       found_test_op = true;
-      test_op = node;
```
shresthamalik: Not sure why we removed the test_op?

test_ngtf.py
```diff
@@ -260,11 +268,26 @@ def main():
...
+    if (arguments.gpu_unit_tests_enable):
+        os.environ['NGRAPH_TF_BACKEND'] = 'GPU'
+        run_ngtf_gtests(
+            build_dir, 
+            '-ArrayOps.Quanti*:ArrayOps.Dequant*:BackendManager.BackendAssignment:'
+            'MathOps.AnyKeepDims:MathOps.AnyNegativeAxis:MathOps.AnyPositiveAxis:'
+            'MathOps.AllKeepDims:MathOps.AllNegativeAxis:MathOps.AllPositiveAxis:'
+            'NNOps.Qu*:NNOps.SoftmaxZeroDimTest*:'
+            'NNOps.SparseSoftmaxCrossEntropyWithLogits'
+        )
```
sayantan-nervana: I suppose these tests fail on GPU?

test/opexecuter.cpp
avijit-nervana: This was triggering a warning (of uninitialized memory read - potentially crashing in case of a test failure) which was legit. The `test_op` isn't actually used any where - no?

test_ngtf.py
sayantan-nervana: For now we have added special flags for GPU (`gpu_unit_tests_enable` and `ngraph_also_build_gpu_backend`). When running this test suite on other backends, do we need to do anything special, in which case we might want to refactor this later to have a single unified flag for all backends


eg: `--ngraph_build_other_backend=GPU,INTERPRETER` and `--unit_test_backend=GPU,INTERPRETER`.

Each backend might have its exclusion list of tests etc.

test/opexecuter.cpp
shresthamalik: It is used in the error message to give more information to the user why the graph was invalid.
Suppose you are testing "Add" op's translation. And you created a graph with 2 add ops we would print the op names for both Add ops and error out as only 1 "add" op is allowed in the graph for unit testing.
The crash should not happen because of uninitialized memory as when found_test_op is set to true, test_op is also initialized. 

It is only used for providing more information to the user.

avijit-nervana: This is the warning I was getting:
```
/nfs/site/home/avijitch/workspace/ngraph-tf/test/opexecuter.cpp:100:80: warning: ‘test_op’ may be used uninitialized in this function [-Wmaybe-uninitialized]
                                    << node->name() << " and " << test_op->name();
```
I guess looking at the logic of this function - it's not possible for the `test_op` remain uninitialized. As soon as the control enters into the line 98 the first time the ASSERT_FALSE won't fire (as the found_test_op is set to false) - so the error message trying to print the test_op (which is uninitialized at this time) won't trigger. The static analyzer for the C++ compiler weren't able to analyze the complex scenario - so flagged this as a potential error condition. 
Initializing the `test_op` to `nullptr` is fixes this warning.

PR #451, opened 2019-03-09
Langjian/allreduce dependency

jianyinglang:
This PR provides one solution to resolve the allreduce op out-of-order issue.
Sort the allreduce ops to be deterministic order and add control dependencies between them to ensure the order will be the same across different ranks.
Some of the code was cherry-picked from https://github.com/NervanaSystems/ngraph-tf/pull/445


src/ngraph_builder.cc
```diff
@@ -33,6 +33,10 @@
 #include "tensorflow/core/graph/edgeset.h"
 #include "tensorflow/core/lib/core/errors.h"
 
+#if defined(NGRAPH_DISTRIBUTED)
+#include <mpi.h>
```
wenzhe-nrv: please use distributed class and get rank in this file.

src/ngraph_builder.cc
```diff
@@ -4172,6 +4185,18 @@ Status Builder::TranslateGraph(
   vector<const Node*> tf_ret_vals;
   vector<const Node*> tf_ops;
 
+#if defined(NGRAPH_DISTRIBUTED)
+  int flag = 0;
+  int mpi_rank = -1;
+  bool mpi_initialized = false;
+  MPI_Initialized(&flag);
```
wenzhe-nrv: same as above.

src/ngraph_encapsulate_op.cc
```diff
@@ -292,6 +292,59 @@ class NGraphEncapsulateOp : public OpKernel {
...
+            ngraph::Distributed dist;
+            int Rank_ID;
+            Rank_ID = dist.get_rank();
+            std::cout << "NGTF_Rank: " << Rank_ID << "  " << node->get_name()
+                      << "(";
+            vector<string> inputs;
+            for (const ng::descriptor::Input& input : node->get_inputs()) {
+              inputs.push_back(input.get_tensor().get_name());
+            }
+            std::cout << ng::join(inputs);
```
wenzhe-nrv: I assume we should use some debug log like `NGRAPH_VLOG` instead of `cout` for this logging messages. The same applies to logging message below.

src/ngraph_encapsulate_op.cc
```diff
@@ -292,6 +292,59 @@ class NGraphEncapsulateOp : public OpKernel {
 
       auto function_size = ng_function->get_graph_size() / 1024;  // kb unit
 
+      // Get the serialized ops and stored the allreduce ops to a vector and
+      vector<shared_ptr<ng::Node>> allreduce_op_list;
```
wenzhe-nrv: It would be better to use nGraph class `NodeVector`. https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/node_vector.hpp

src/ngraph_encapsulate_op.cc
```diff
@@ -292,6 +292,59 @@ class NGraphEncapsulateOp : public OpKernel {
...
+        }
+      }
+
+      // Sort the allreduce ops according to the TF names
+      std::sort(allreduce_op_list.begin(), allreduce_op_list.end(),
+                [](shared_ptr<ng::Node>& x, shared_ptr<ng::Node>& y) {
+                  return x->get_friendly_name() < y->get_friendly_name();
+                });
+      // Add control dependency in for the allreduce ops
+      if (allreduce_op_list.size() > 1) {
```
wenzhe-nrv: please use `!allreduce_op_list.empty()`.
https://en.cppreference.com/w/cpp/container/vector/empty

jianyinglang: This  only applies the number of allreduce is equal to or larger than 2. If the list size is equal to 1, we do not need to insert any dependency.

src/ngraph_builder.cc
jianyinglang: changed

src/ngraph_builder.cc
jianyinglang: changed

src/ngraph_encapsulate_op.cc
jianyinglang: I would like to get all the output in a single line so it is easy to do grep analysis. Tried `NGRAPH_VLOG` and it outputs separated lines. Not sure how to fix that

src/ngraph_encapsulate_op.cc
jianyinglang: Changed

src/ngraph_builder.cc
wenzhe-nrv: you don't have to say. Github shows it's `Outdated` after change on the line

src/ngraph_encapsulate_op.cc
wenzhe-nrv: @avijit-nervana any thought?

avijit-nervana: I think if it helps to debug the race conditions then it's ok to dump to stdout. However, please change the following:
1. Package the entire AllReduce processing into a separate function and call it from here if it's a distributed build. 
2. Within the function if you want you may use a different debug flag to enable/disable the logging to stdout

src/ngraph_encapsulate_op.cc
```diff
@@ -309,6 +309,60 @@ class NGraphEncapsulateOp : public OpKernel {
 
       auto function_size = ng_function->get_graph_size() / 1024;  // kb unit
 
+      // Get the serialized ops and stored the allreduce ops to a vector and
+      ng::NodeVector allreduce_op_list;
+      for (const shared_ptr<ng::Node>& node : ng_function->get_ordered_ops()) {
+        if (node->get_name().compare(0, 9, "AllReduce") == 0) {
```
shresthamalik: Can we use node->description() to get the type of op?
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/node.hpp#L115

src/ngraph_encapsulate_op.cc
shresthamalik: @jianyinglang You can create a debug_string and then print it using NGRAPH_VLOG.
stringstream is one way
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_encapsulate_op.cc#L263

src/ngraph_encapsulate_op.cc
```diff
@@ -309,6 +309,60 @@ class NGraphEncapsulateOp : public OpKernel {
...
+          allreduce_op_list.push_back(node);
+          // Diagnosis to check the input, dependencies, and output for each
+          // Allreduce op
+          if (std::getenv("NGRAPH_DISABLE_LOGGING") == nullptr) {
+#if defined NGRAPH_DISTRIBUTED
```
shresthamalik: This "#if defined NGRAPH_DISTRIBUTED" should be moved above allreduce_op_list is initialized?

shresthamalik: Builder would be a better place to add the control dependencies in the graph. You can move the code here. 
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_builder.cc#L4328

Once the function is created you can add the control dependencies.

src/ngraph_encapsulate_op.cc
```diff
@@ -309,6 +309,60 @@ class NGraphEncapsulateOp : public OpKernel {
...
+            ngraph::Distributed dist;
+            int Rank_ID;
+            Rank_ID = dist.get_rank();
+            std::cout << "NGTF_Rank: " << Rank_ID << "  " << node->get_name()
+                      << "(";
+            vector<string> inputs;
+            for (const ng::descriptor::Input& input : node->get_inputs()) {
+              inputs.push_back(input.get_tensor().get_name());
```
shresthamalik: Interesting, Do we have tensor and tensor names at this stage?

src/ngraph_encapsulate_op.cc
```diff
@@ -309,6 +309,60 @@ class NGraphEncapsulateOp : public OpKernel {
...
+          });
+      // Add control dependency in for the allreduce ops
+      if (allreduce_op_list.size() > 1) {
+        for (size_t i = 1; i < allreduce_op_list.size(); ++i) {
+          auto pre_node = allreduce_op_list[i - 1];
+          auto cur_node = allreduce_op_list[i];
+          cur_node->add_control_dependency(pre_node);
+        }
+      }
+
```
shresthamalik: In general this would take care of the dependencies within the cluster (nGraphEncapsulate). In future we might want to take care about inserting dependencies within these clusters if multiple clusters have all reduce ops and there is no data flow between them

src/ngraph_encapsulate_op.cc
jianyinglang: It seems yes. It will give the input tensor name for allreduce ops.

src/ngraph_builder.cc
```diff
@@ -4326,6 +4307,38 @@ Status Builder::TranslateGraph(
...
   ng_function = make_shared<ng::Function>(ng_result_list, ng_parameter_list);
 
+  // Get the serialized ops and stored the allreduce ops to a vector and
+  ng::NodeVector allreduce_op_list;
+  for (const shared_ptr<ng::Node>& node : ng_function->get_ordered_ops()) {
+    if (node->description() == "AllReduce") {
+      allreduce_op_list.push_back(node);
+// Diagnosis to check the input, dependencies, and output for each
+// Allreduce op
+#if defined NGRAPH_DISTRIBUTED
```
shresthamalik: Please move the #if defined NGRAPH_DISTRIBUTED to line 4309 i.e. the for loop to get the all_reduce node etc can all be moved inside the check?

src/ngraph_builder.cc
```diff
@@ -4054,6 +4035,38 @@ Status Builder::TranslateGraph(
   //
   ng_function = make_shared<ng::Function>(ng_result_list, ng_parameter_list);
 
+#if defined NGRAPH_DISTRIBUTED
```
avijit-nervana: Please create a function for this and call that function from here.

avijit-nervana: Add a unit test to test that function.

src/ngraph_utils.cc
```diff
@@ -265,6 +265,32 @@ void MemoryProfile(long& vm_usage, long& resident_set) {
     vm_usage = vsize / 1024;   // unit kb
     resident_set = rss * page_size_kb;
   }
+}
+
+void NgraphNodeDump(const shared_ptr<ng::Node>& node,
+                    const std::shared_ptr<ngraph::Function>& ng_function,
+                    int RankID) {
```
avijit-nervana: Why is RankID a CamelCase when other variables are snake_case? Please follow the correct coding convention.

src/ngraph_utils.cc
```diff
@@ -265,6 +265,32 @@ void MemoryProfile(long& vm_usage, long& resident_set) {
...
+  for (auto nd : node->get_control_dependencies()) {
+    dependencies.push_back(nd->get_name());
+  }
+  node_info_ss << ng::join(dependencies) << "] -> ";
+
+  vector<string> outputs;
+  for (size_t i = 0; i < node->get_output_size(); ++i) {
+    outputs.push_back(node->get_output_tensor(i).get_name());
+  }
+  node_info_ss << ng::join(outputs) << "  " << ng_function->get_name();
```
avijit-nervana: Just curious - how does the output look like? Why are you appending the function name at the end? Also - if you just need the function name - then pass that as a parameter - not the function itself.

src/ngraph_utils.h
```diff
@@ -205,6 +205,9 @@ void NgraphSerialize(const std::string&,
 // Collect the total memory usage through /proc/self/stat
 void MemoryProfile(long&, long&);
 
+// Get the inputs, outputs, and depdencies for the node
```
avijit-nervana: This function is very specific to nGraph Node - so it should really be added to nGraph so that other parts of nGraph (bridge/backend) can make use it.

NGraph Node already has a `<<` operator and a function called write_long_description() - which you should update with printing information you are printing here. 
https://github.com/NervanaSystems/ngraph/blob/aa52081c2e5acac6dd343d2c16cc3a00b7676dab/src/ngraph/node.hpp#L152

jianyinglang: I removed this node logging function from the bridge now and will add it to nGraph core.

PR #454, opened 2019-03-13
sarkars/refactor builder

sayantan-nervana:
1. Add friendly names to ngraph nodes and ng_function
2. refactor min, max, all, any, sum, prod and mean as a reduction op
3. refactor argmax and argmin
4. removed 2 compilation warnings (1 unused function, 1 unused variable). now there are no warnings when building
5. Added c++ unit tests for 3 ops (sum, prod, mean), which were missing from c++ or python unit tests

### DIFF
--- a/src/ngraph_builder.cc
+++ b/src/ngraph_builder.cc
@@ -2630,43 +2630,17 @@ static Status TranslateQuantizeAndDequantizeV2Op(
 }
 
 static Status TranslateQuantizedConv(
-    const Node* op, const std::vector<const Tensor*>& static_input_map,
-    Builder::OpMap& ng_op_map,
+    const Node* op, Builder::OpMap& ng_op_map,
     std::function<std::shared_ptr<ng::Node>(
         std::vector<std::shared_ptr<ng::Node>>, ng::Strides, ng::Strides,
-        ng::CoordinateDiff, ng::CoordinateDiff, ng::Strides,
-        std::vector<std::shared_ptr<ng::op::Constant>>)>
+        ng::CoordinateDiff, ng::CoordinateDiff, ng::Strides)>
         create_quantized_conv_node) {
   size_t num_tf_op_inputs = op->num_inputs();
-  std::vector<int32> static_inps_idxs;
-  if (GetNodeAttr(op->attrs(), "_ngraph_static_inputs", &static_inps_idxs) !=
-      Status::OK()) {
-    return errors::Internal(
-        "Did not find static inputs in quantized convolution op");
-  }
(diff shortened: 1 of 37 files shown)


src/ngraph_builder.cc
```diff
@@ -81,6 +81,16 @@ static void SaveNgOp(Builder::OpMap& ng_op_map, const std::string& op_name,
   // no need to try-catch, map[key] will create vector object
   // if not exists
   ng_op_map[op_name].push_back(output_node);
+  // TODO (sarkars): remove try catch when ngraph is updated
+  try {
+    output_node->set_friendly_name(op_name + "_" +
+                                   to_string(ng_op_map[op_name].size()));
+  } catch (const ngraph::ngraph_error& e) {
```
shresthamalik: Do we want to do size()-1 ? To match the indexing output indexing in TF?

sayantan-nervana: yes, of course!

shresthamalik: May be use "[]" for indexes. Also add a unit test to test the name setting

src/ngraph_encapsulate_op.cc
```diff
@@ -289,6 +289,7 @@ class NGraphEncapsulateOp : public OpKernel {
       OP_REQUIRES_OK(
           ctx, Builder::TranslateGraph(input_shapes, static_input_map, &m_graph,
                                        ng_function));
+      ng_function->set_friendly_name("enc_" + to_string(m_ngraph_cluster));
```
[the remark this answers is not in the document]
shresthamalik: Just name() also works

src/ngraph_builder.cc
```diff
@@ -2216,64 +2118,45 @@ static Status TranslateMeanOp(
...
-
-    ng::AxisVector ng_axis_order(ng_min->get_shape().size());
-    std::iota(ng_axis_order.begin(), ng_axis_order.end(), 0);
+static Status TranslateMeanOp(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  return TranslateReduceOp(
+      op, static_input_map, ng_op_map,
+      [](std::shared_ptr<ng::Node> ng_input, ng::AxisSet ng_reduction_axes) {
+        return ng::builder::mean(ng_input, ng_reduction_axes);
```
sayantan-nervana: certain ops like this use ngraph builder, which is adding ng_nodes to the ngraph graph outside our reach. So its not easy to set friendly names to intermediate names to nodes created by some builder function.

PR #458, opened 2019-03-17
Avijit/add trace events

avijit-nervana:
Use this environment variable to enable generating the traces: `NGRAPH_TF_ENABLE_TRACING`

### DIFF
--- a/src/ngraph_encapsulate_op.cc
+++ b/src/ngraph_encapsulate_op.cc
@@ -74,6 +74,10 @@ class NGraphEncapsulateOp : public OpKernel {
     my_instance_id = s_instance_count;
     s_instance_count++;
 
+    std::ostringstream oss;
+    oss << "Encapsulate_" << my_instance_id << ": " << name();
+    Event event(oss.str().c_str(), name().c_str());
+
     NGRAPH_VLOG(1) << "NGraphEncapsulateOp: " << my_instance_id
                    << " Name: " << name();
 
@@ -139,12 +143,18 @@ class NGraphEncapsulateOp : public OpKernel {
     OP_REQUIRES_OK(ctx,
                    ctx->GetAttr<string>("_ngraph_backend", &m_op_backend_name));
     BackendManager::CreateBackend(m_op_backend_name);
+    event.Stop();
+    Event::WriteTrace(event);
   }
 
   //---------------------------------------------------------------------------
   //  ~NGraphEncapsulateOp()
   //---------------------------------------------------------------------------
   ~NGraphEncapsulateOp() override {
+    std::ostringstream oss;
+    oss << "Destroy Encapsulate_" << my_instance_id << ": " << name();
+    Event event(oss.str().c_str(), name().c_str());
+
     // If the kernel goes away, we must de-register all of its cached
     // functions
     // from the freshness tracker.
@@ -161,6 +171,8 @@ class NGraphEncapsulateOp : public OpKernel {
       BackendManager::ReleaseBackend(m_op_backend_name);
       NGRAPH_VLOG(2) << "~NGraphEncapsulateOp()";
     }
+    event.Stop();
+    Event::WriteTrace(event);
   }
 
   template <typename T>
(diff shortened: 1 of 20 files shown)


test/test_timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+//   // v = Ab^T
+//   auto v =
+//       ops::MatMul(root.WithOpName("v"), A, b, ops::MatMul::TransposeB(true));
+//   std::vector<Tensor> outputs;
+//   ClientSession session(root);
+//   // Run and fetch v
+//   ASSERT_OK(session.Run({v}, &outputs));
+//   // Expect outputs[0] == [19; -3]
+//   LOG(INFO) << outputs[0].matrix<float>();
+// }
```
sayantan-nervana: this commented section can be removed

test/test_timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+    threads.push_back(std::move(next_thread));
+  }
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+  for (auto& next : threads) {
+    next.join();
+  }
+
+  trace_file << "]" << std::endl;
```
sayantan-nervana: `trace_file.close();`

test/test_timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+//       ops::MatMul(root.WithOpName("v"), A, b, ops::MatMul::TransposeB(true));
+//   std::vector<Tensor> outputs;
+//   ClientSession session(root);
+//   // Run and fetch v
+//   ASSERT_OK(session.Run({v}, &outputs));
+//   // Expect outputs[0] == [19; -3]
+//   LOG(INFO) << outputs[0].matrix<float>();
+// }
+
+TEST(timer, event_record) {
```
sayantan-nervana: It is recommended not to have `_` in gtest names. Quote from the gtest primer:
```
Both names must be valid C++ identifiers, and they should not contain underscore (_).
```

test/test_timer.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+  for (auto& next : threads) {
+    next.join();
+  }
+
+  trace_file << "]" << std::endl;
+}
+
+TEST(timer, event_file) {
```
sayantan-nervana: `_` in test name

src/ngraph_encapsulate_op.cc
```diff
@@ -240,6 +256,7 @@ class NGraphEncapsulateOp : public OpKernel {
     ng::runtime::Backend* op_backend =
         BackendManager::GetBackend(m_op_backend_name);
 
+    Event event_func_maybe_create("FunctionMaybeCreate", name().c_str());
```
sayantan-nervana: We could separate the "cache lookup" vs "TranslateGraph + compile" time and have 2 separate events for them (we could separate TranslateGraph and ngraph-compile as well, to get an idea which part of the function compilation takes more time)
OR
`m_args` could be set for event `event_func_maybe_create`, to say if its a cache hit or a cache miss.
Then we wouldn't have to look at the length of the `event_func_maybe_create` event and guess if there was a cache hit or miss

We could do it in a separate PR

src/ngraph_encapsulate_op.cc
```diff
@@ -74,6 +74,10 @@ class NGraphEncapsulateOp : public OpKernel {
     my_instance_id = s_instance_count;
     s_instance_count++;
 
+    std::ostringstream oss;
+    oss << "Encapsulate_" << my_instance_id << ": " << name();
+    Event event(oss.str().c_str(), name().c_str());
+
```
sayantan-nervana: Do we need `my_instance_id`? `name()` already provides a unique name and also embeds the encapsulate ID (`m_ngraph_cluster`) information in the name

src/ngraph_timer.h
```diff
@@ -16,23 +16,102 @@
...
+  static bool IsTracingEnabled() {
+    static bool enabled = (std::getenv("NGRAPH_TF_ENABLE_TRACING") != nullptr);
+    return enabled;
+  }
+
+  friend std::ostream& operator<<(std::ostream& out_stream, const Event& event);
+
+  Event(const Event&) = delete;
+  Event& operator=(Event const&) = delete;
+
```
sayantan-nervana: might need a destructor to `s_event_log.close();`

src/ngraph_encapsulate_op.cc
avijit-nervana: We will address it in a separate PR.

src/ngraph_encapsulate_op.cc
avijit-nervana: The instance id is useful for other correlations. So will keep it for now.

src/ngraph_timer.h
avijit-nervana: Since `s_event_log` is a static I am not sure how would a destructor be helpful here. Also we flush after every write - so the file will have all the necessary output before finish. The file handle will be closed by the destructor of the  `close()` isn't going to add anything extra (the destructor of the filebuf object will close the actual file anyway)

PR #461, opened 2019-03-19
Rel 0.12.0-rc1 prep

avijit-nervana:
* Changed the name of the build directory to `build_cmake` as the name `build` conflicts with the bazel `BUILD` file for the MacOS.
* Upgraded to TensorFlow 1.13.1 release
* Removed the build option 3 as it's not maintained. The new home for nGraph bridge will be `tensorflow/ngraph_bridge` after the current release.


crlishka:
TESTNOW

PR #462, opened 2019-03-19
Sindhu/topkv2,select

sindhu-nervana:


src/ngraph_builder.cc
```diff
@@ -3937,6 +3937,45 @@ static Status TranslateTileOp(
   return Status::OK();
 }
 
+// Translate TopKV2 Op using ngraph core op TopK
+static Status TranslateTopKV2Op(
+    const Node* op, const std::vector<const Tensor*>& static_input_map,
+    Builder::OpMap& ng_op_map) {
+  shared_ptr<ngraph::Node> ng_input;
+  ValidateInputCount(op, 2);
```
mingshan-wang: Is this for debugging purpose? Or we still need this check here.

src/ngraph_builder.cc
```diff
@@ -3937,6 +3937,45 @@ static Status TranslateTileOp(
...
+
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &ng_k));
+  k = ng_k[0];
+
+  GetNodeAttr(op->attrs(), "sorted", &sorted);
+
+  // ngraph topk computes topk min if sorted is false
+  // TODO:ngraph team need to add an attribute to always compute max for topkv2
+  if (sorted == false) {
+    return errors::InvalidArgument("topkv2 should compute max, not min");
```
mingshan-wang: For the error message here, maybe we should just say `TopKV2 doesn't support sorted equals False.`

src/ngraph_builder.cc
```diff
@@ -3937,6 +3937,45 @@ static Status TranslateTileOp(
...
+  }
+
+  shared_ptr<ngraph::Node> ng_result = make_shared<ngraph::op::TopK>(
+      ng_input, k_axis, ng::element::i32, k, sorted);
```
mingshan-wang: Should we fix the `ng::element::i32` here, this is for input data type or sth else's data type?

src/ngraph_builder.cc
sindhu-nervana: I added while i was debugging but just left it since it is good to validate.

src/ngraph_builder.cc
sindhu-nervana: sure, will change this.

src/ngraph_builder.cc
sindhu-nervana: this is for the index element type which is always an int, so it could be int32 or int34.

sayantan-nervana: Output of TF's topkv2 op is int32 (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/nn_ops.cc#L1250), `so ng::element::i32` is fine here

src/ngraph_builder.cc
```diff
@@ -3937,6 +3937,47 @@ static Status TranslateTileOp(
...
+  bool sorted = true;
+
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &ng_k));
+  k = ng_k[0];
+
+  GetNodeAttr(op->attrs(), "sorted", &sorted);
+
+  // ngraph topk computes topk min if sorted is false
+  // TODO:ngraph team need to add an attribute to always compute max for topkv2
+  if (sorted == false) {
```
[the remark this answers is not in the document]
sindhu-nervana: moved it

src/ngraph_mark_for_clustering.cc
```diff
@@ -360,6 +361,19 @@ Status MarkForClustering(Graph* graph) {
...
+      confirmation_function_map["TopKV2"] = [](Node* n, bool* result) {
+        bool sorted = true;
+        GetNodeAttr(n->attrs(), "sorted", &sorted);
+
+        // ngraph topk computes topk min if sorted is false
+        // TODO:ngraph team need to add an attribute to always compute max for
+        // topkv2
+        if (sorted == false) {
+          return errors::InvalidArgument(
+              "TopKV2 doesn't support sorted equals False.");
```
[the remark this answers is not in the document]
mingshan-wang: Right now ngraph topk will by default always sort the result, so if the attribute is false, then ngraph will return incorrect result, the unit test will fail.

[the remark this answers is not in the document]
sindhu-nervana: yeah additional to what Mingshan mentioned, ngraph has compute_max and compute_min parameters which sorts for both in ascending or descending orders. but we need descending for topkv2 and not sort unless specified, hence it returns error and fails when we set sort to false. i have the pytest for sorted=false disabled right now and i have comments added as well.

src/ngraph_encapsulate_op.cc
```diff
@@ -588,6 +587,7 @@ class NGraphEncapsulateOp : public OpKernel {
                         "Caught exception while executing nGraph computation: ",
                         exp.what(), "\n"));
       } catch (...) {
+        ng_function = m_ng_function_map[ng_exec];
```
sindhu-nervana: This is a fix to my earlier PR, without this line it would not dump serialize graph if backend compile call fails.

src/ngraph_mark_for_clustering.cc
```diff
@@ -360,6 +361,19 @@ Status MarkForClustering(Graph* graph) {
       confirmation_function_map["Tanh"] = SimpleConfirmationFunction();
       confirmation_function_map["TanhGrad"] = SimpleConfirmationFunction();
       confirmation_function_map["Tile"] = SimpleConfirmationFunction();
+      confirmation_function_map["TopKV2"] = [](Node* n, bool* result) {
+        bool sorted = true;
+        GetNodeAttr(n->attrs(), "sorted", &sorted);
```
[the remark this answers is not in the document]
sindhu-nervana: made this change.

test/python/test_topkv2.py
```diff
@@ -0,0 +1,67 @@
...
+import numpy as np
+import tensorflow as tf
+
+from common import NgraphTest
+
+
+class TestTopKV2(NgraphTest):
+
+    def test_topkv2_1d(self):
+        input = tf.constant((1.0, 5.0, 6.0, 12.0), dtype=tf.float32)
```
sayantan-nervana: change constant to placeholder to avoid constant folding

test/python/test_topkv2.py
```diff
@@ -0,0 +1,67 @@
...
+        out = tf.nn.top_k(input, k=3, sorted=True)
+
+        def run_test(sess):
+            return sess.run(out).values
+
+        assert (
+            self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
+
+    def test_topkv2_2d(self):
+        input = tf.constant(
```
sayantan-nervana: change constant to placeholder to avoid constant folding

test/python/test_topkv2.py
```diff
@@ -0,0 +1,67 @@
...
+        def run_test(sess):
+            return sess.run(out).values
+
+        assert (
+            self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
+
+    @pytest.mark.skip(
+        reason="Falls back to TF, sorted=False is not supported currently")
+    def test_topkv2_nosort(self):
+        input = tf.constant(
```
sayantan-nervana: change constant to placeholder to avoid constant folding

test/python/test_topkv2.py
```diff
@@ -0,0 +1,67 @@
...
+            [[40.0, 30.0, 20.0, 10.0], [10.0, 20.0, 15.0, 70.0]],
```
sayantan-nervana: Please add a test that does Topk on 3D or 3D tensors too

sindhu-nervana: @sayantan-nervana made these changes. please review

src/ngraph_builder.cc
```diff
@@ -3665,6 +3665,43 @@ static Status TranslateTileOp(
...
+  bool sorted = true;
+
+  TF_RETURN_IF_ERROR(GetStaticInputVector(op, 1, static_input_map, &ng_k));
+  k = ng_k[0];
+
+  // sorted = false is not supported right now, it falls back to TF if set to
+  // false.
+  GetNodeAttr(op->attrs(), "sorted", &sorted);
+
+  // index element type - currently only int32 or int64 are supported by ngraph
```
sayantan-nervana: We can remove this comment. TF specifies i32, so we are i32 here.

PR #464, opened 2019-03-20
Fix round mode in QDQV2

nbpatel:
TF uses HALF_TO_EVEN round mode and ngtf was using NEAREST_TOWARD_INFINITY.
This change fixes https://jira.devtools.intel.com/browse/NGTF-1921 and also the accuracy issue with quantized mnist


sayantan-nervana:
@libinta : We are changing the round mode from INFINITY to EVEN. It seems EVEN is the right thing to do https://github.com/tensorflow/tensorflow/blob/66fc32d73bb769d14d535c04345a2497983b7905/tensorflow/core/ops/array_ops.cc#L2736

Would this affect your backend in any way if the mode is changed? It definitely helps CPU backend, because now they can use mkldnn instead of reference kernels,

src/ngraph_builder.cc
```diff
@@ -2410,13 +2410,12 @@ static Status TranslateQuantizeAndDequantizeV2Op(
...
       op->name(), ng_q_et, ng::Shape(), std::vector<int>({0}));
   ng::op::Quantize::RoundMode ng_round_mode =
-      ng::op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_INFINITY;
-  auto ng_quant = ConstructNgNode<ng::op::Quantize>(
-      op->name(), ng_input, ng_scale, ng_offset, ng_q_et, ng::AxisSet(),
-      ng_round_mode);
-  SaveNgOp(ng_op_map, op->name(), ConstructNgNode<ng::op::Dequantize>(
-                                      op->name(), ng_quant, ng_scale, ng_offset,
-                                      ng_r_et, ng::AxisSet()));
+      ng::op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN;
```
adstraw: Round nearest toward even is fine as a default value but the TF definition of this op...

https://github.com/tensorflow/tensorflow/blob/66fc32d73bb769d14d535c04345a2497983b7905/tensorflow/core/ops/array_ops.cc#L2736

... includes two round modes:

HALF_UP -> ROUND_NEAREST_UPWARD
HALF_TO_EVEN -> ROUND_NEAREST_TOWARD_EVEN

This code should read round_mode attribute from TF and map to nGraph round modes.  Mapping for the two supported nGraph round modes is shown above.

PR #465, opened 2019-03-20
Kanvi/custom grappler optimizer

kanvi-nervana:


src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  // Convert the GraphDef to Graph
+  GraphConstructorOptions opts;
+  opts.allow_internal_ops = true;
+  opts.expect_device_spec = true;
+  Graph graph(OpRegistry::Global());
+  TF_RETURN_IF_ERROR(ConvertGraphDefToGraph(opts, item.graph, &graph));
+
+  // Get the nodes to be skipped
+  std::vector<string> skip_these_nodes;
```
shresthamalik: Can use a set instead of vector? Then we don't need to loop on the vector.

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+
+using namespace std;
+
+namespace tensorflow {
+namespace ngraph_bridge {
+
+Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
+                                 const tensorflow::grappler::GrapplerItem& item,
+                                 GraphDef* output) {
+  NGRAPH_VLOG(1) << "Here at NgraphOptimizer ";
```
shresthamalik: Change the log string to "In ngraphOptimizer"

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  // If requested, dump unmarked graphs.
+  if (DumpUnmarkedGraphs()) {
+    DumpGraphs(graph, idx, "unmarked", "Unmarked Graph");
+  }
+
+  // If ngraph is disabled via ngraph_bridge api or NGRAPH_TF_DISABLE is set
+  // we will not do anything; all subsequent
+  // passes become a no-op.
+  if (config::IsEnabled() == false ||
```
shresthamalik: We dont need this check as we are exiting after dumping precaptured.

src/ngraph_capture_variables.cc
```diff
@@ -28,25 +28,32 @@ namespace tensorflow {
...
-// Utility function to check if placement on the NGRAPH device has been
+// Utility function to check if placement on the NGRAPH has been
 // requested.
-//
-// FIXME(amprocte): stubbed out for now because NGRAPH device is gone.
-//
-static bool NGraphPlacementRequested(const Node* node) { return true; }
+static bool NGraphPlacementRequested(const Node* node,
+                                     std::vector<string> skip_these_nodes) {
+  for (string& f : skip_these_nodes) {
```
shresthamalik: You can avoid the for loop if you use a set.

src/ngraph_mark_for_clustering.cc
```diff
@@ -61,6 +61,18 @@ static Status NGraphPlacementRequested(Node* node, bool& placement_ok) {
...
+static Status CheckIfOutputNode(const Node* node,
+                                std::vector<string> skip_these_nodes,
+                                bool& skip_it) {
+  for (string& f : skip_these_nodes) {
+    if (node->name() == f) {
+      skip_it = true;
+      return Status::OK();
+    }
+  }
+  return Status::OK();
```
shresthamalik: May be move this check as part of NGraphPlacementRequested like in the Captured pass

src/ngraph_mark_for_clustering.cc
```diff
@@ -594,6 +606,14 @@ Status MarkForClustering(Graph* graph) {
...
       }
 
+      bool skip_it = false;
+      TF_RETURN_IF_ERROR(CheckIfOutputNode(node, skip_these_nodes, skip_it));
+      if (skip_it) {
+        NGRAPH_VLOG(5) << "Found Output Node: " << node->name()
+                       << " - skip marking it for clustering";
+        break;
+      }
+
```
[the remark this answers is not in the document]
shresthamalik: Or moved to NGraphPlacementRequested

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+
+namespace tensorflow {
+namespace ngraph_bridge {
+
+Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
+                                 const tensorflow::grappler::GrapplerItem& item,
+                                 GraphDef* output) {
+  NGRAPH_VLOG(1) << "Here at NgraphOptimizer ";
+
+  // Convert the GraphDef to Graph
```
avijit-nervana: Move this block after line 84 i.e., after checking whether to continue with nGraph.

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+  if (DumpPrecaptureGraphs()) {
+    DumpGraphs(graph, idx, "precapture", "Pre-Capture Graph");
+  }
+
+  // If ngraph is disabled via ngraph_bridge api or NGRAPH_TF_DISABLE is set
+  // we will not do anything; all subsequent
+  // passes become a no-op.
+  if (config::IsEnabled() == false ||
+      std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
+    return Status::OK();
```
avijit-nervana: Add a VLOG(0) indicating that NGRAPH is disabled.

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+    DumpGraphs(graph, idx, "tracked",
+               "Graph with Variables Rewritten for Tracking");
+  }
+
+  // Convert the graph back to Graphdef
+  graph.ToGraphDef(output);
+  return Status::OK();
+}
+
+void NgraphOptimizer::Feedback(tensorflow::grappler::Cluster* cluster,
```
avijit-nervana: Remove the parameter names. Often compilers warn about unused parameters.

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+  NGRAPH_VLOG(0) << "NGRAPH-TF OPTIMIZER Dumping main graph to "
+                 << dot_filename;
+  NGRAPH_VLOG(0) << "NGRAPH-TF OPTIMIZER Dumping main graph to "
+                 << pbtxt_filename;
+
+  GraphToDotFile(&graph, dot_filename, title);
+  GraphToPbTextFile(&graph, pbtxt_filename);
+}
+
+REGISTER_GRAPH_OPTIMIZER_AS(NgraphOptimizer, "ng-optimizer");
```
avijit-nervana: call it `ngraph-optimizer`

src/grappler/ngraph_optimizer.h
```diff
@@ -0,0 +1,159 @@
...
+
+namespace ngraph_bridge {
+
+// Custom Grappler Optimizer for NGraph-TF
+class NgraphOptimizer : public tensorflow::grappler::CustomGraphOptimizer {
+ public:
+  NgraphOptimizer() = default;
+  ~NgraphOptimizer() override = default;
+
+  string name() const override { return "NgraphOptimizer"; };
```
avijit-nervana: Shouldn't it be `ngraph-optimizer`?

src/grappler/ngraph_optimizer.h
```diff
@@ -0,0 +1,159 @@
...
+
+  Status Optimize(tensorflow::grappler::Cluster* cluster,
+                  const tensorflow::grappler::GrapplerItem& item,
+                  GraphDef* output) override;
+
+  void Feedback(tensorflow::grappler::Cluster* cluster,
+                const tensorflow::grappler::GrapplerItem& item,
+                const GraphDef& optimize_output, double result) override;
+
+ protected:
```
avijit-nervana: Why protected? Which other class is deriving from this? Make it private.

src/grappler/ngraph_optimizer.h
```diff
@@ -0,0 +1,159 @@
...
+ private:
+  static std::string DotFilename(std::string kind, int idx) {
+    return GraphFilenamePrefix(kind, idx) + ".dot";
+  }
+
+  static std::string PbtxtFilename(std::string kind, int idx) {
+    return GraphFilenamePrefix(kind, idx) + ".pbtxt";
+  }
+
+  static std::string GraphFilenamePrefix(std::string kind, int idx) {
```
avijit-nervana: Move all the implementations to the cc file.

src/ngraph_capture_variables.cc
```diff
@@ -28,25 +28,32 @@ namespace tensorflow {
...
+    }
+  }
+  return true;
+}
 
 //
 // Main entry point for the variable-capture.
 //
-Status CaptureVariables(Graph* graph) {
+Status CaptureVariables(Graph* graph, std::vector<string> skip_these_nodes) {
```
avijit-nervana: Why pass by value?

src/ngraph_capture_variables.h
```diff
@@ -22,7 +22,8 @@ namespace tensorflow {
 
 namespace ngraph_bridge {
 
-Status CaptureVariables(Graph* graph);
+Status CaptureVariables(Graph* graph,
+                        std::vector<string> skip_these_nodes = {});
```
avijit-nervana: Remove the default value. Make the caller explicitly provide all the parameters.

src/grappler/ngraph_optimizer.h
```diff
@@ -0,0 +1,159 @@
...
+  static int FreshIndex() {
+    mutex_lock l(s_serial_counter_mutex);
+    return s_serial_counter++;
+  }
+
+  static bool DumpAllGraphs() {
+    return std::getenv("NGRAPH_TF_DUMP_GRAPHS") != nullptr;
+  }
+
+  static bool DumpPrecaptureGraphs() {
```
sayantan-nervana: These functions like `DumpPrecaptureGraphs` etc are also present in `class NGraphRewritePass : public GraphOptimizationPass`. They do not use any class variables etc, and might be moved out of both classes to utils, so that we avoid this duplication

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
...
+  if (DumpEncapsulatedGraphs()) {
+    DumpGraphs(graph, idx, "encapsulated", "Graph with Clusters Encapsulated");
+  }
+
+  // Rewrite for tracking then, if requested, dump the graphs.
+  TF_RETURN_IF_ERROR(RewriteForTracking(&graph));
+  if (DumpTrackedGraphs()) {
+    DumpGraphs(graph, idx, "tracked",
+               "Graph with Variables Rewritten for Tracking");
+  }
```
sayantan-nervana: This core logic of `MarkForClustering`-->`AssignClusters`-->`DeassignClusters`-->`EncapsulateClusters`-->`RewriteForTracking` is also present in `class NGraphVariableCapturePass : public NGraphRewritePass`. The functions in use here are independent of the classes and might be moved out to some `core_rewrite` function, which is used by both.

examples/mnist/mnist_deep_simplified_with_grappler.py
```diff
@@ -0,0 +1,262 @@
...
+
+    rewrite_options = rewriter_config_pb2.RewriterConfig(custom_optimizers=[
+        rewriter_config_pb2.RewriterConfig.CustomGraphOptimizer(
+            name="ng-optimizer")
+    ])
+    config = tf.ConfigProto(
+        allow_soft_placement=True,
+        log_device_placement=False,
+        inter_op_parallelism_threads=1,
+        graph_options=tf.GraphOptions(rewrite_options=rewrite_options))
```
sayantan-nervana: We can modify the original `mnist_deep_simplified` file with a argument that chooses if the network is run in older mode or grappler mode. Then this whole file does not need to be duplicated

src/ngraph_mark_for_clustering.cc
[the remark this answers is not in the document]
kanvi-nervana: Moved it to the top. 
Do we need the NGraphPlacementRequested ? I was thinking of removing that check.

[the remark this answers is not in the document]
shresthamalik: Currently it is used at many places in the bridge. You can remove it from everywhere, maybe better to do it as a separate PR.  I suggested to move it to NGraphPlacementRequested as in capture_variables pass it was being done there.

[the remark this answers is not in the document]
kanvi-nervana: I will rename it appropriately in the capture variable pass and remove it elsewhere in a separate PR.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Currently, used a vector because it was small list. Will be replacing it with a map since we will need the name as well as the id. This is being done in the separate PR so did not want make a lot of changes here.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Done. Thanks for catching this.

src/ngraph_capture_variables.cc
kanvi-nervana: Currently, I used a vector because it was small list. Will be replacing it with a map since we will need the name as well as the id. This is being done in the separate PR so did not want make a lot of changes here.

src/ngraph_mark_for_clustering.cc
kanvi-nervana: I will remove the unused NGraphPlacementRequested function in a separate PR, so we can leave this check separate.

src/grappler/ngraph_optimizer.cc
```diff
@@ -0,0 +1,188 @@
+/*******************************************************************************
+ * Copyright 2017-2018 Intel Corporation
```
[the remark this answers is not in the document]
kanvi-nervana: Changed.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Changed

src/grappler/ngraph_optimizer.h
kanvi-nervana: Done

src/grappler/ngraph_optimizer.h
kanvi-nervana: Done

src/ngraph_capture_variables.h
kanvi-nervana: Done. Removed the default value from here as well as the MarkForClustering, in the process had to touch some test files as well which were calling these functions directly.

src/grappler/ngraph_optimizer.h
kanvi-nervana: I think here we need to return the name of the class.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Done

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Moved the check to the very top.

src/grappler/ngraph_optimizer.cc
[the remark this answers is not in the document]
kanvi-nervana: Did you mean higher log level?
Changed it to 5.
I am not sure what all things we can print here besides the grappler item id, feed nodes, fetch nodes etc. ? Would that be it?

src/grappler/ngraph_optimizer.h
kanvi-nervana: Done

examples/mnist/mnist_deep_simplified_with_grappler.py
kanvi-nervana: Done

src/ngraph_mark_for_clustering.cc
```diff
@@ -570,6 +582,15 @@ Status MarkForClustering(Graph* graph) {
     bool mark_for_clustering = false;
 
     do {
+      // check if output node
+      bool skip_it = false;
+      TF_RETURN_IF_ERROR(CheckIfOutputNode(node, skip_these_nodes, skip_it));
+      if (skip_it) {
+        NGRAPH_VLOG(5) << "Found Output Node: " << node->name()
```
sindhu-nervana: should we reduce the log level here?

examples/mnist/mnist_deep_simplified.py
```diff
@@ -125,10 +126,22 @@ def bias_variable(shape):
...
+        config = tf.ConfigProto(
+            allow_soft_placement=True,
+            log_device_placement=False,
+            inter_op_parallelism_threads=1,
+            graph_options=tf.GraphOptions(rewrite_options=rewrite_options))
+    else:
+        config = tf.ConfigProto(
+            allow_soft_placement=True,
+            log_device_placement=False,
+            inter_op_parallelism_threads=1)
```
sayantan-nervana: We can minimize the amount of code change by doing this. This would be an example on the example minimal code change required to use grappler style ngtf: Just add a new field to some existing ConfigProto

```
config = tf.ConfigProto(
            allow_soft_placement=True,
            log_device_placement=False,
            inter_op_parallelism_threads=1)
if (FLAGS.use_grappler):
    rewrite_options = rewriter_config_pb2.RewriterConfig(custom_optimizers=[
            rewriter_config_pb2.RewriterConfig.CustomGraphOptimizer(
                name="ngraph-optimizer")
        ])
    config.MergeFrom(tf.ConfigProto(graph_options=tf.GraphOptions(rewrite_options=rewrite_options)))
```

src/ngraph_capture_variables.cc
```diff
@@ -28,25 +28,32 @@ namespace tensorflow {
...
+static bool CheckIfOutputNode(const Node* node,
+                              std::vector<string> skip_these_nodes) {
+  for (string& f : skip_these_nodes) {
+    if (node->name() == f) {
+      NGRAPH_VLOG(5) << "Found Output Node: " << node->name()
+                     << " - skip capturing it";
+      return false;
+    }
+  }
+  return true;
```
sayantan-nervana: we could use find:
```
bool found = std::find (skip_these_nodes.begin(), skip_these_nodes.end(), node->name()) != skip_these_nodes.end();
if (!found){
    //VLOG
}
return found;
```

Once we update to set, you could use set's find instead of the generic find

src/ngraph_mark_for_clustering.cc
kanvi-nervana: I did not want a lot of logs to be displayed at the lower log level.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Ok. I will need to move the check for NGraph enable back to where it was since in this case we need to get the graph from the graphdef to be able to dump it.

PR #467, opened 2019-03-25
Avijit/build script changes

avijit-nervana:
* Refactored the build_ngtf.py by creating tools/build_utils.py 
* Added an option "--artifact_dir" to allow specifying the artifacts directory


tools/test_inference_models.py
```diff
@@ -0,0 +1,102 @@
...
+        raise "Need to provide artifacts directory"
+
+    if not (arguments.python_location):
+        raise "Need to provide the location of Python directory"
+
+    if not (arguments.model_location):
+        raise "Need to provide the location of models directory"
+
+    # Check to make sure that there is TensorFlow installed and will work
+    # TODO
```
sayantan-nervana: Do you mean something like this in the TODO? 
```
import tensorflow as tf
assert (tf.__version__ == '1.13.1')
```

sayantan-nervana: Should we put `test_inference_models.py` under `tools` or `test`?

avijit-nervana: No. We need to find out what version of TF was used to build NG and what version of TF this one is.

avijit-nervana: @sayantan-nervana 
>Should we put test_inference_models.py under tools or test

No. I am planning to move these to a new ci directory. for now this is temp.

PR #471, opened 2019-03-26
Sindhu/path fix to run tf python unit tests

sindhu-nervana:


avijit-nervana:
@crlishka We need to add better error checking in the test_ngtf.py. Something we will consider in a future PR.

sindhu-nervana:
> These changes look good. Before I approve, though, I would like to ask both you and Avijit if there is anything we can do in test_ngtf.py to fail CI in the future if run_tensorflow_pytests() cannot find any tests to run? A major issue is that test_ngtf.py was silent (i.e. returned with exit code 0) even though 300 or so tests stopped running. How do we fix test_ngtf.py to avoid this?

sindhu-nervana:
Hi @crlishka I made changes to exit with error when the test runner's test list is empty. This should throw error when test_ngtf.py fails to run `run_tensorflow_pytests`

PR #473, opened 2019-03-27
Sindhu/onehot op

sindhu-nervana:


test/test_array_ops.cpp
```diff
@@ -316,6 +316,188 @@ TEST(ArrayOps, ExpandDims) {
...
+}
+
+TEST(ArrayOps, OneHot1d) {
+  Scope root = Scope::NewRootScope();
+
+  vector<int> static_input_indexes = {1};
+
+  Tensor indices(DT_INT32, TensorShape({4}));
+
+  AssignInputValues<int>(indices, {0, 2, -1, 1});
```
sayantan-nervana: Here the absolute values of all the requested indices are < 4 (the size). Does TF accept numbers larger than 4?

test/test_array_ops.cpp
```diff
@@ -316,6 +316,188 @@ TEST(ArrayOps, ExpandDims) {
...
+  Tensor depth(DT_INT32, TensorShape({}));
+  Tensor on_value(DT_FLOAT, TensorShape({}));
+  Tensor off_value(DT_FLOAT, TensorShape({}));
+
+  AssignInputValues<int>(depth, 3);
+  AssignInputValues<float>(on_value, 5.0);
+  AssignInputValues<float>(off_value, 0.0);
+
+  auto attrs = ops::OneHot::Attrs();
+  attrs.axis_ = -1;
```
sayantan-nervana: I think we checked it, but can you please confirm that axis can only take values [-1, rank)

src/ngraph_builder.cc
```diff
@@ -2233,6 +2233,66 @@ static Status TranslateDirectReduceOp(
...
+        "OneHot Op: depth of one hot dimension must be scalar ", depth.size());
+  }
+
+  int one_hot_axis;
+  TF_RETURN_IF_ERROR(GetNodeAttr(op->attrs(), "axis", &one_hot_axis));
+
+  std::vector<int> shape;
+  for (int i = 0; i < ng_features_rank; i++) {
+    shape.push_back(ng_features_shape[i]);
+  }
```
sayantan-nervana: ngraph's `Shape` is subclassed from `vector<size_t>`. So you can use it as a vector, no need to create a new vector out of it
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/shape.hpp#L28

```
ng::Shape output_shape(ng_features_shape);
// declare pos
if (one_hot_axis == -1) {
one_hot_axis = ng_features_rank;
pos = output_shape.end();
} else {
pos = output_shape.begin() + one_hot_axis;
}
output_shape.insert(pos, depth[0]);
```

test/test_array_ops.cpp
```diff
@@ -316,6 +316,188 @@ TEST(ArrayOps, ExpandDims) {
...
+
+  AssignInputValuesRandom<int>(indices, -2, 5);
+
+  Tensor depth(DT_INT32, TensorShape({}));
+  Tensor on_value(DT_FLOAT, TensorShape({}));
+  Tensor off_value(DT_FLOAT, TensorShape({}));
+
+  AssignInputValues<int>(depth, 2);
+  AssignInputValues<float>(on_value, 5.0);
+  AssignInputValues<float>(off_value, 0.0);
```
sayantan-nervana: Can you please change one of the UTs to accept on-off vals that are simply 1 and 0
Also can you confirm if there is no constructor of TF onehot that does not accept on-off values?

test/test_array_ops.cpp
sindhu-nervana: i already verified this while writing the unit tests and right, axis can only take values -1 to rank.

test/test_array_ops.cpp
sindhu-nervana: It does. modified unit tests with indices 50 and -50. onehot(50) or onehot(-50) will return all zeroes.

PR #478, opened 2019-03-29
Mingshan/Adding resnet50 validation script

mingshan-wang:
This PR added the validation script for resnet50 training with both synthetic data and real data. 

The tf result references under tfGPU/ folder is collected running the same command in the script on TF GPU. 

The patch to make the data loader for real data deterministic is also included, and also the patch to eliminate the average_loss encapsulates in the training graph.


test/validate_resnet50/validation.py
```diff
@@ -0,0 +1,263 @@
...
+    # reapply the patch
+    output, error_output = command_executor(
+        'git apply -R ' + './datasets_make_deterministic.patch')
+
+
+# Validation with synthetic data
+
+if __name__ == "__main__":
+    reference_file_name_realData = ''
+    reference_file_name_syntheticData = ''
```
sayantan-nervana: we are sending in empty strings, and counting on the `run_validation` to detect that and create new files if needed, right?

test/validate_resnet50/validation.py
```diff
@@ -0,0 +1,263 @@
...
+    batch_size = 100
+    run_validation("real_data", reference_file_name_realData, batch_size)
+    batch_size = 100
```
sayantan-nervana: this line can be removed

test/validate_resnet50/validation.py
```diff
@@ -0,0 +1,263 @@
...
+        else:
+            command_executor('git apply --whitespace=nowarn ' +
+                             './datasets_make_deterministic.patch')
+
+    # Run the validation command on NGraph
+    if (data_format == "real_data"):
+        command_to_run = validate_with_real_data_command_NG + str(batch_size)
+    elif (data_format == "synthetic_data"):
+        command_to_run = validate_with_synthetic_data_command_NG + \
+            str(batch_size)
```
sayantan-nervana: an `else` can be added that throws an exception to complete out this ladder. perhaps `exit(1)` in keeping with the other parts of this script (eg line 169) in case of fatal errors

test/validate_resnet50/validation.py
```diff
@@ -0,0 +1,263 @@
...
+                top1_acc.append(line.split()[-2])
+                top5_acc.append(line.split()[-1])
+
+    return total_loss, top1_acc, top5_acc
+
+
+def check_validation_results(norm_dict, metric):
+    test_pass = True
+    for norm in norm_dict:
+        if norm_dict[norm] > 0.1:
```
sayantan-nervana: so if we get ref accuracy = 75, and ng accuracy = 75.3, then is it a failure?

test/validate_resnet50/validation.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if to_parse:
+            total_loss.append(line.split()[-3])
+            top1_acc.append(line.split()[-2])
+            top5_acc.append(line.split()[-1])
+
+    return total_loss, top1_acc, top5_acc
+
+
+def parse_reference_file(filename):
```
sayantan-nervana: `parse_reference_file` and `parse_training_output` can be a single function... I think they are separate because one parses a file, and the other parses string. Maybe we keep the string parsing function and just read the file into a string and reuse.

test/validate_resnet50/validation.py
shresthamalik: yes.

test/validate_resnet50/validation.py
shresthamalik: yes

shresthamalik: This script is not comparing the accuracy. It compares the training loss value at every iteration

shresthamalik: Let me check.

sayantan-nervana: Same thing for loss. If ref loss is 1, and we get 0.8, is the test passing?

Issue #89, opened 2018-07-20
How to use resnet50 model in tf_cnn_benchmarks.py?


ljppro:
Hi,
I download the savedmodel and checkpoints from tensorflow/models and untar them into a single directory. But when I run this benchmark, an error reported as:
NotFoundError (see above for traceback): Key v0/cg/affine0/biases not found in c                                                                             heckpoint
         [[Node: save/RestoreV2 = RestoreV2[dtypes=[DT_INT64, DT_FLOAT, DT_FLOAT                                                                             , DT_FLOAT, DT_FLOAT, ..., DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT], _d                                                                             evice="/job:localhost/replica:0/task:0/device:CPU:0"](_arg_save/Const_0_0, save/                                                                             RestoreV2/tensor_names, save/RestoreV2/shape_and_slices)]]

I tried another models from other sites also. But the almost same error is there.

What resnet model can work with this benchmark and how should I specify the data_dir parameter?

Thanks!

sharathns93:
Hi,
Thanks for trying out the ngraph-tf bridge. The pretrained models downloaded from tensorflow/models repo are  unfortunately not compatible with tf_cnn_benchmarks.py script. 

However you can train a model locally with dummy data for a few iterations using this command:
  python tf_cnn_benchmarks.py --batch_size 16 --model resnet50 --train_dir <path_to_resnet_model> --num_batches=10 --data_format NHWC --select_device CPU

and then use the pretrained model generated here for evaluation. 
Here is the command line: 
python tf_cnn_benchmarks.py --batch_size 128 --model resnet50 --train_dir <path_to_resnet_model> --num_batches=1000   --data_format NCHW --eval --select_device NGRAPH

For best performance on  2-socket Intel(R) Xeon(R) Platinum 8180 add the following before running the command:
KMP_BLOCKTIME=20 OMP_NUM_THREADS=28  KMP_AFFINITY=granularity=fine,proclist=[0-27]

Please note that since the model is trained with dummy data, the accuracy numbers generated are random and not correct. You can also train it with real data (tf_records for Imagenet or CIFAR10) to completion and then verify to get the right accuracy numbers.

PR #485, opened 2019-04-04
kanvi/Grappler optimizer improvements

kanvi-nervana:
Add IdentityN nodes with the same name as the fetch node so that the fetch node can be encapsulated if it is supported.


src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
   if (config::IsEnabled() == false ||
       std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
     NGRAPH_VLOG(0) << "Ngraph is disabled ";
+    graph.ToGraphDef(output);
```
sayantan-nervana: should we move this to the very top? the if
```
if (config::IsEnabled() == false ||
      std::getenv("NGRAPH_TF_DISABLE") != nullptr)
```

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
     return Status::OK();
   }
 
   // Get the nodes to be skipped
   std::vector<string> fetch_nodes;
   for (const string& f : item.fetch) {
-    NGRAPH_VLOG(5) << "Skip fetch node: " << f;
     int pos = f.find(":");
     fetch_nodes.push_back(f.substr(0, pos));
```
sayantan-nervana: could we make `fetch_nodes` a set. As a vector it can potentially look like this:
<node1, node1, node2> (because node1 had 2 output tensors node1:0, node1:1)

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+        NGRAPH_VLOG(5) << "Creating an IdentityN node";
+        Node* identityN_node;
+        if (NodeBuilder(node->name(), "IdentityN")
+                .Attr("T", input_types)
+                .Input(inputs)
+                .Device(node->assigned_device_name())
+                .Finalize(input_graph, &identityN_node) != Status::OK()) {
+          NGRAPH_VLOG(5) << "Unsuccessful in constructing the node";
+        } else {
+          NGRAPH_VLOG(5) << "Successfully constructed IdentityN node";
```
sayantan-nervana: Maybe we should wrap it with a TF_RETURN_IF_ERROR. not sure if we should just print and ignore the failure to construct identityN

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+        }
+
+        identityN_node->set_assigned_device_name(node->assigned_device_name());
+
+        // Rename the skip node
+        NGRAPH_VLOG(5) << "Renaming node";
+        // Get a new name for the node with the given prefix
+        // We will use the 'original-node-name_ng' as the prefix
+        string new_name = input_graph->NewName(node->name() + "_ng");
```
sayantan-nervana: Lets add a TODO here to ensure that we get a unique name (or does `NewName` already take care of that?)

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+        node->set_name(new_name);
+        NGRAPH_VLOG(5) << "New name for skip node " << node->name();
+      } else {
+        NGRAPH_VLOG(5) << "num outputs " << node->num_outputs();
```
sayantan-nervana: here num_outputs() is definitely zero right?

src/grappler/ngraph_optimizer.h
```diff
@@ -22,6 +22,7 @@
 #include "tensorflow/core/graph/graph.h"
 #include "tensorflow/core/graph/graph_constructor.h"
 #include "tensorflow/core/graph/graph_def_builder.h"
+#include "tensorflow/core/graph/node_builder.h"
```
sayantan-nervana: With this upgrade is `skip_these_nodes` needed?
```
Status MarkForClustering(Graph* graph,
                         const std::vector<string> skip_these_nodes) {
```

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+        NGRAPH_VLOG(5) << "Cannot constructed an IdentityN node";
+      }
+    }
+  }
+
```
sayantan-nervana: We need to remove `skip_these_nodes` from places like this right? Now nothing is to be skipped right?
`TF_RETURN_IF_ERROR(CaptureVariables(&graph, skip_these_nodes));`

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+                .Attr("T", input_types)
+                .Input(inputs)
+                .Device(node->assigned_device_name())
+                .Finalize(input_graph, &identityN_node) != Status::OK()) {
+          NGRAPH_VLOG(5) << "Unsuccessful in constructing the node";
+        } else {
+          NGRAPH_VLOG(5) << "Successfully constructed IdentityN node";
+        }
+
+        identityN_node->set_assigned_device_name(node->assigned_device_name());
```
sayantan-nervana: Maybe we add:
`identityN_node->AddAttr("_ngraph_marked_for_clustering", false);`
to make sure we never cluster Identity_N nodes added here in case we start supporting IdentityN in ngtf.

A related change there would be:
https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_mark_for_clustering.cc#L714
here we set to true, only if the attribute is not set to false before

src/grappler/ngraph_optimizer.cc
shresthamalik: Keeping it here allows us to dump graphs even though ngraph is disabled. Helps in scoping new models.

kanvi-nervana: Yes.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: @sindhu-nervana Here https://github.com/NervanaSystems/ngraph-tf/blob/893ef86a043d1dee59ceddb746fb92c3c28ea48a/src/ngraph_rewrite_for_tracking.cc#L74 the error is returned from Node building and not the NewName. NewName only gives a new name to an existing node with the prefix you provide.
@sayantan-nervana The comment here mentions that NewName gives a new unique name. https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/graph/graph.h#L569

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Yes. Added that log to figure out why the node wasn't created.

src/grappler/ngraph_optimizer.cc
```diff
@@ -66,18 +66,81 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+            NGRAPH_VLOG(5) << DataTypeString(node->output_type(i))
+                           << " is ref type";
+            ref_type = true;
+            break;
+          }
+          inputs.push_back(NodeBuilder::NodeOut(node, i));
+        }
+
+        if (ref_type) {
+          NGRAPH_VLOG(5) << "Cannot create an IdentityN node";
```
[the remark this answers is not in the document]
kanvi-nervana: It is printed in line 103, that the datatype is ref type.

src/grappler/ngraph_optimizer.h
kanvi-nervana: Yes, keeping it around for future changes like we discussed.

src/grappler/ngraph_optimizer.cc
```diff
@@ -40,8 +40,9 @@ namespace ngraph_bridge {
 Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
                                  const tensorflow::grappler::GrapplerItem& item,
                                  GraphDef* output) {
-  NGRAPH_VLOG(5) << "Here at NgraphOptimizer ";
-  NGRAPH_VLOG(5) << "NgraphOptimizer : grappler item id " << item.id;
+  NGRAPH_VLOG(0) << "[NGTF-OPTIMIZER] Here at NgraphOptimizer ";
+  NGRAPH_VLOG(5) << "[NGTF-OPTIMIZER] NgraphOptimizer : grappler item id "
+                 << item.id;
```
sindhu-nervana: Minor: maybe rename `[NGTF_OPTIMIZER]` in brackets to just `NGTF_OPTIMIZER` like how we have `NGTF_SUMMARY`

src/grappler/ngraph_optimizer.cc
```diff
@@ -40,8 +40,9 @@ namespace ngraph_bridge {
 Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
                                  const tensorflow::grappler::GrapplerItem& item,
                                  GraphDef* output) {
-  NGRAPH_VLOG(5) << "Here at NgraphOptimizer ";
-  NGRAPH_VLOG(5) << "NgraphOptimizer : grappler item id " << item.id;
+  NGRAPH_VLOG(0) << "[NGTF-OPTIMIZER] Here at NgraphOptimizer ";
```
sindhu-nervana: remove line #43 and keep just #44? 
incase you want to keep both for different log levels, then we can remove "Here at NgraphOptimizer" and just keep [NGTF-OPTIMIZER]  right?

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
   // passes become a no-op.
   if (config::IsEnabled() == false ||
       std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
-    NGRAPH_VLOG(0) << "Ngraph is disabled ";
+    NGRAPH_VLOG(0) << "[NGTF-OPTIMIZER] Ngraph is disabled ";
```
sindhu-nervana: same as above

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+      // Check the number of outputs of the 'fetch_node'
+      // Only move further to create an IdentityN node
+      // if it is greater than 0
+      // Also, make sure that none of the output types is
+      // a ref type
+      if (node->num_outputs()) {
+        std::vector<NodeBuilder::NodeOut> inputs;
+        std::vector<DataType> input_types;
+        for (int i = 0; i < node->num_outputs(); i++) {
+          if (!IsRefType(node->output_type(i))) {
```
sayantan-nervana: Can you please add some comments why ref_type is getting special treatment. In case of ref_type would grappler path fail then? Also should we mark this node as skipped, so that its not clustered

sayantan-nervana: Maybe in a separate PR we have to consider adding unit tests

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
-    NGRAPH_VLOG(5) << "Skip fetch node: " << f;
     int pos = f.find(":");
-    fetch_nodes.push_back(f.substr(0, pos));
+    fetch_nodes.insert(f.substr(0, pos));
+  }
+  std::set<string>& skip_these_nodes = fetch_nodes;
+
+  // Rewrite graph to add identity node so the skip node can be encapsulated
+  // as well
+  Graph* input_graph = &graph;
```
shresthamalik: We can move this graph rewrite into a separate function call like the other graph-rewrite function calls (MarkForClustering, AssignClusters) etc.

sayantan-nervana: So `MarkForClustering`, `AssignClusters` etc are used by both grappler path and normal path. This code is specific to grappler path.

So the options we have for the identityN adding code are:
1. the code stays here
2. the code is parceled off to a separate file, but it stays in `src/grappler`
3. The code is parceled off into a separate and placed in `src`

I  like the idea that this portion is a separate function (Then it will be easy to unit test it etc). But unsure where to place it, `src/grappler` or `src`.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Using [NGTF-OPTIMIZER] as a TAG with every log and I want to keep both because LOG(0) would show up anyway, making it easier to figure out that the grappler is being used.

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+        // Get a new name for the node with the given prefix
+        // We will use the 'original-node-name_ng' as the prefix
+        string new_name = input_graph->NewName(node->name() + "_ngraph");
+        node->set_name(new_name);
+        NGRAPH_VLOG(5) << "[NGTF-OPTIMIZER] New name for fetch node "
+                       << node->name();
+      } else {
+        NGRAPH_VLOG(5) << "[NGTF-OPTIMIZER] num outputs "
+                       << node->num_outputs();
+        NGRAPH_VLOG(5) << "[NGTF-OPTIMIZER] Cannot construct an IdentityN node";
```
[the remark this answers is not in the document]
kanvi-nervana: The else is for information purposes as to why the IdentityN node wasn't added. But this node will still be skipped.

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+                               .Attr("T", input_types)
+                               .Input(inputs)
+                               .Device(node->assigned_device_name())
+                               .Finalize(input_graph, &identityN_node));
+
+        identityN_node->set_assigned_device_name(node->assigned_device_name());
+
+        // Rename the skip node
+        // Get a new name for the node with the given prefix
+        // We will use the 'original-node-name_ng' as the prefix
```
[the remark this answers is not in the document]
kanvi-nervana: good catch

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +66,79 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+      // a ref type
+      if (node->num_outputs()) {
+        std::vector<NodeBuilder::NodeOut> inputs;
+        std::vector<DataType> input_types;
+        for (int i = 0; i < node->num_outputs(); i++) {
+          if (!IsRefType(node->output_type(i))) {
+            input_types.push_back(node->output_type(i));
+          } else {
+            NGRAPH_VLOG(5) << "[NGTF-OPTIMIZER] "
+                           << DataTypeString(node->output_type(i))
```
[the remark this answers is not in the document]
kanvi-nervana: Node name is printed on line #89 irrespective.

src/grappler/ngraph_optimizer.cc
kanvi-nervana: I did add comments on line 94. I'll add more info.

src/grappler/ngraph_optimizer.cc
shresthamalik: option 2. function in separate file in src/grappler

src/grappler/ngraph_optimizer.cc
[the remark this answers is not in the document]
shresthamalik: We can print the output index that is  a ref type. The log currently seems to print "[NGTF-OPTIMIZER] <ref_type> is ref type"

src/grappler/ngraph_optimizer.cc
kanvi-nervana: Will be done in the next PR

src/grappler/ngraph_optimizer.cc
kanvi-nervana: next PR

src/grappler/ngraph_optimizer.cc
```diff
@@ -65,18 +65,83 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
...
+                               .Input(inputs)
+                               .Device(node->assigned_device_name())
+                               .Finalize(input_graph, &identityN_node));
+
+        identityN_node->set_assigned_device_name(node->assigned_device_name());
+
+        // Rename the skip node
+        // Get a new name for the node with the given prefix
+        // We will use the 'original-node-name_ngraph' as the prefix
+        string new_name = input_graph->NewName(node->name() + "_ngraph");
```
sayantan-nervana:
```suggestion
        string new_name = input_graph->NewName(node->name() + "_ngraph");
        // TODO: Use (guaranteed) unique name here
```

PR #486, opened 2019-04-05
Avijit/more traces

avijit-nervana:
* Removed the event trace from the bridge as it's now available in nGraph
* Replace the call sites with nGraph APIs for event tracing. 
* Enabled flags for running unit/other tests on GPU


test/ci/buildkite/test_runner.py
```diff
@@ -86,13 +91,31 @@ def main():
...
-        run_ngtf_cpp_gtests(arguments.artifacts_dir, './', None)
+        test_filter = None
+        if arguments.backend:
+            if 'GPU' in arguments.backend:
+                test_filter = str(
+                    "-ArrayOps.Quanti*:ArrayOps.Dequant*:BackendManager.BackendAssignment:"
+                    "MathOps.AnyKeepDims:MathOps.AnyNegativeAxis:MathOps.AnyPositiveAxis:"
+                    "MathOps.AllKeepDims:MathOps.AllNegativeAxis:MathOps.AllPositiveAxis:"
+                    "NNOps.Qu*:NNOps.SoftmaxZeroDimTest*:"
+                    "NNOps.SparseSoftmaxCrossEntropyWithLogits")
```
sayantan-nervana: Minor:
lets add a `else <fail>` here, since only GPU backend is in the test setup right now

tools/test_utils.py
```diff
@@ -394,3 +395,20 @@ def run_bazel_build_test(venv_dir, build_dir):
 
     # Return to the original directory
     os.chdir(root_pwd)
+
+
+def run_bazel_build():
```
sayantan-nervana: this is a cool addition

PR #488, opened 2019-04-05
Sindhu/Update Select Op to accept more complex shapes as inputs

sindhu-nervana:


test/python/test_select.py
```diff
@@ -64,3 +64,20 @@ def run_test(sess):
...
+
+    def test_select_complexshape(self):
+        a = [1, 0, 1, 0, 1, 1, 0]
+        x = [[[[4], [2], [9]], [[3], [8], [2]]],
+             [[[6], [6], [2]], [[5], [4], [1]]],
+             [[[7], [6], [0]], [[1], [1], [1]]],
+             [[[5], [0], [9]], [[8], [7], [6]]],
+             [[[1], [9], [6]], [[3], [2], [8]]],
+             [[[9], [2], [5]], [[3], [3], [6]]],
+             [[[4], [9], [3]], [[6], [1], [4]]]]
```
sayantan-nervana: please add a `# comment` beside each line indicating their shapes
Also the predicate in this test case is of shape [7]

we need more combinations (eg, [7 2 1] as predicate shape, [5 3 7 2 1] as x, y shape

Also you dont need to hand populate x, just use np's random function or np.ones([3,4,5]) etc

src/ngraph_builder.cc
```diff
@@ -4187,14 +4187,71 @@ static Status TranslateSelectOp(
...
+                                                  second, ng_input3);
+    }
+    if (first == ng_input2) {
+      std::tie(first, ng_input_new) =
+          ng::builder::numpy_broadcast(std::make_pair(first, ng_input_new));
+      std::tie(ng_input_new, ng_input3) =
+          ng::builder::numpy_broadcast(std::make_pair(ng_input_new, ng_input3));
+      ng_select = ConstructNgNode<ng::op::Select>(op->name(), first,
+                                                  ng_input_new, ng_input3);
+    }
```
sayantan-nervana: lines 4227-4230 could possibly be done without any if-else I think. but we need to check

For example:
```if (first == ng_input1)
   {ng::builder::numpy_broadcast(std::make_pair(ng_input_new, second));}```
means: `{ng::builder::numpy_broadcast(std::make_pair(ng_input_new, ng_input2));}`
vs
```if (first == ng_input2) 
{ng::builder::numpy_broadcast(std::make_pair(first, ng_input_new));}```
means: `{ng::builder::numpy_broadcast(std::make_pair(ng_input2, ng_input_new));}`

so we are doing a broadcast on `ng_input_new` and `ng_input2` in both cases (except in a different order). After the reshape would the order matter?

eg: are these 2 same:  `numpy_broadcast([5 7 3 2 1], [7 1 1 1])` vs  `numpy_broadcast([7 1 1 1], [5 7 3 2 1])`
I think they are same (but need to check), hence the order does not matter. hence that whole block of if-else is not needed perhaps

Comment (sindhu-nervana) [reply src/ngraph_builder.cc]: Figured I do not have to have 2 conditions. Condition tensor and input tensor can either have same shape or when input tensor rank is higher than condition, condition tensor can only be rank 1 as per TF where Op. made changes accordingly.

File: src/ngraph_builder.cc [hunk 3]
```diff
@@ -4187,14 +4187,63 @@ static Status TranslateSelectOp(
...
 
-  // broadcast to make all tensors same shape, as required by ngraph select op
-  std::tie(ng_input1, ng_input2) =
-      ng::builder::numpy_broadcast(std::make_pair(ng_input1, ng_input2));
-  std::tie(ng_input2, ng_input3) =
-      ng::builder::numpy_broadcast(std::make_pair(ng_input2, ng_input3));
+  if (ng_input2->get_shape() != ng_input2->get_shape()) {
+    return errors::InvalidArgument(
+        "Input tensors 2 and 3 should have same shape");
+  }
```
sayantan-nervana: lets add the other check: (rank of pred == 1) || (rank of pred > 1 && rank of pred == rank of input 1 && pred_shape[0] == input1[0])

We should never fail this check ever as TF will barf before we reach this stage, but its good to have as documentation. Lets have a test or a comment to this effect

src/ngraph_builder.cc
```diff
@@ -4187,14 +4187,63 @@ static Status TranslateSelectOp(
...
+
+  auto ng_input1_shape = ng_input1->get_shape();
+  auto ng_input2_shape = ng_input2->get_shape();
+
+  auto ng_input1_size = ng_input1->get_shape().size();
```
sayantan-nervana: minor: maybe lets call it `ng_input1_rank`. `size` generally means the number of elements (size = product(shape_vector)) while rank=len(shape_vector)

src/ngraph_builder.cc
```diff
@@ -4187,14 +4187,63 @@ static Status TranslateSelectOp(
...
+                                                    ng_axis_order, tmp_vector);
+
+    // broadcast to make all tensors same shape, as required by ngraph select op
+    std::tie(ng_input_new, ng_input2) =
+        ng::builder::numpy_broadcast(std::make_pair(ng_input_new, ng_input2));
+    std::tie(ng_input2, ng_input3) =
+        ng::builder::numpy_broadcast(std::make_pair(ng_input2, ng_input3));
+
+    ng_select = ConstructNgNode<ng::op::Select>(op->name(), ng_input_new,
+                                                ng_input2, ng_input3);
```
sayantan-nervana: the broadcast and select op can move outside the `if`

simply: `auto broadcast_input_1 = length != 0 ? ng_input_new : ng_input1` or something like that

src/ngraph_builder.cc
```diff
@@ -4187,14 +4187,63 @@ static Status TranslateSelectOp(
...
+    }
+  }
 
-  auto ng_select = ConstructNgNode<ng::op::Select>(op->name(), ng_input1,
-                                                   ng_input2, ng_input3);
+  if (length != 0) {
+    // Condition tensor will be modified to align the condition tensor
+    // shape with input tensor shape index and fill the rest of the vector with
+    // 1s
+    // Eg: consition tensor [7], input tensor [7, 3, 2, 1]
```
```suggestion
    // Eg: condition tensor [7], input tensor [7, 3, 2, 1]
```

src/ngraph_builder.cc
```diff
@@ -4187,14 +4187,63 @@ static Status TranslateSelectOp(
   TF_RETURN_IF_ERROR(
       GetInputNodes(ng_op_map, op, &ng_input1, &ng_input2, &ng_input3));
 
-  // broadcast to make all tensors same shape, as required by ngraph select op
-  std::tie(ng_input1, ng_input2) =
-      ng::builder::numpy_broadcast(std::make_pair(ng_input1, ng_input2));
-  std::tie(ng_input2, ng_input3) =
-      ng::builder::numpy_broadcast(std::make_pair(ng_input2, ng_input3));
+  if (ng_input2->get_shape() != ng_input2->get_shape()) {
```
```suggestion
  if (ng_input2->get_shape() != ng_input3->get_shape()) {
```

test/python/test_select.py
```diff
@@ -64,3 +64,44 @@ def run_test(sess):
...
+    )
+    def test_select_fail(self):
+        a = np.random.randint(2, size=[5])
+        x = np.random.uniform(0, 11, [4, 3, 1])
+
+        p = tf.placeholder(dtype=tf.bool)
+        out = tf.where(p, x, x)
+
+        def run_test(sess):
+            return (sess.run(out, feed_dict={p: a})).shape
```
sayantan-nervana: why are we returning `shape` from `run_test`. Shouldn't we return the actual data tensor?

sindhu-nervana: ah! i did this for debugging purpose to print shapes. fixed this!

test/python/test_select.py
```diff
@@ -64,3 +64,47 @@ def run_test(sess):
...
+        p = tf.placeholder(dtype=tf.bool)
+        out = tf.where(p, x, x)
+
+        def run_test(sess):
+            return (sess.run(out, feed_dict={p: a}))
+
+        assert (
+            self.with_ngraph(run_test) == self.without_ngraph(run_test)).all()
+
+    #failin test case
```
sayantan-nervana:
```suggestion
    #failing test case
```

test/python/test_select.py
```diff
@@ -64,3 +64,47 @@ def run_test(sess):
...
+    @pytest.mark.skip(
+        reason=
+        "First dimension of input tensor should match the shape of condition tensor"
```
sayantan-nervana: this fails since its not the right input for TF itself, right?

sindhu-nervana: true, i added while testing. should we keep or remove?

sindhu-nervana: I removed this test.

src/ngraph_builder.cc
```diff
@@ -4273,14 +4273,60 @@ static Status TranslateSelectOp(
...
+    if (ng_input1_shape[0] == ng_input2_shape[i]) {
+      length = ng_input2_rank - i - ng_input1_rank;
+    }
+  }
+
+  if (length != 0) {
+    // Condition tensor will be modified to align the condition tensor
+    // shape with input tensor shape index and fill the rest of the vector with
+    // 1s
+    // Eg: condition tensor [7], input tensor [7, 3, 2, 1]
```
sayantan-nervana: to make the example more generic, maybe we can say `[5, 7, 3, 2, 1]` . (to show that the realignment is not to make the sizes same, just to match up the first matching number)

src/ngraph_builder.cc
```diff
@@ -4273,14 +4273,57 @@ static Status TranslateSelectOp(
...
+    for (size_t i = 0; i < ng_input1_rank; i++) {
+      ng_axis_order.push_back(i);
+    }
+
+    ng_input_new = ConstructNgNode<ng::op::Reshape>(op->name(), ng_input1,
+                                                    ng_axis_order, tmp_vector);
+  }
+
+  std::tie(ng_input1, ng_input2) = ng::builder::numpy_broadcast(std::make_pair(
+      ng_input1 = length != 0 ? ng_input_new : ng_input1, ng_input2));
```
sayantan-nervana: though the assign in the 1st argument to `make_pair` is not effecting the logic of the code, i think its probably unintended and can be removed

src/ngraph_builder.cc
```diff
@@ -4273,14 +4273,57 @@ static Status TranslateSelectOp(
...
+    // shape with input tensor shape index and fill the rest of the vector with
+    // 1s
+    // Eg: condition tensor [7], input tensor [7, 3, 2, 1]
+    // After Reshape, condition tensor will be [7, 1 ,1 ,1] for auto broadcast.
+
+    std::vector<size_t> tmp_vector((length + ng_input1_rank), 1);
+    tmp_vector[0] = ng_input1_shape[0];
+    for (size_t i = 0; i < ng_input1_rank; i++) {
+      ng_axis_order.push_back(i);
+    }
```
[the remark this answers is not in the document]
sindhu-nervana: right! these were as per our old assumptions. i will change these

src/ngraph_builder.cc
sindhu-nervana: i did not understand this.

sindhu-nervana: Done

PR #497, opened 2019-04-10
sindhu/UT runner update, enable Topk and onehot python tests

sindhu-nervana:


test/python/tensorflow/tf_unittest_runner.py
```diff
@@ -330,4 +322,8 @@ def print_results(test_result, invalid_list):
 
 
 if __name__ == '__main__':
-    main()
+    status, results = main()
+    if status == False:
+        for key in ["ERRORS", "FAILED"]:
+            test_name = results[key][0][0][0].id()
+            raise Exception(test_name + ' failed due to ' + key)
```
avijit-nervana:
```suggestion
     raise Exception("Failed")
```
No need to print any test name. If there are multiple failures - then you will only print one.

sindhu-nervana: made this change and removed the for loop here.

sindhu-nervana:
Enabled one_hot and topk python unit tests here.

PR #500, opened 2019-04-10
Shrestha/fix tensor destructor seg fault

shresthamalik:
Fixed Seg Fault Seen in --enable_variables_and_optimizers version. 
Fixed Backend Release Call in master version. 
Added some logs.

### DIFF
--- a/src/enable_variable_ops/ngraph_encapsulate_op.cc
+++ b/src/enable_variable_ops/ngraph_encapsulate_op.cc
@@ -146,6 +146,7 @@ class NGraphEncapsulateOp : public OpKernel {
     // Set the backend type for the op
     OP_REQUIRES_OK(ctx,
                    ctx->GetAttr<string>("_ngraph_backend", &m_op_backend_name));
+    NGRAPH_VLOG(4) << "NGraphEncapsulateOp::Create backend " << def().name();
     BackendManager::CreateBackend(m_op_backend_name);
     event.Stop();
     ngraph::Event::write_trace(event);
@@ -158,7 +159,7 @@ class NGraphEncapsulateOp : public OpKernel {
     std::ostringstream oss;
     oss << "Destroy Encapsulate_" << my_instance_id << ": " << name();
     ngraph::Event event(oss.str(), name(), "");
-
+    NGRAPH_VLOG(2) << "~NGraphEncapsulateOp::" << name();
     // If the kernel goes away, we must de-register all of its cached
     // functions
     // from the freshness tracker.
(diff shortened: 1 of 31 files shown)


src/enable_variable_ops/ngraph_encapsulate_op.cc
```diff
@@ -172,22 +173,17 @@ class NGraphEncapsulateOp : public OpKernel {
       // m_freshness_tracker->Unref();
     }
 
-    string node_name = "_ngraph_cluster_" + to_string(m_ngraph_cluster);
-    // TODO(malikshr) : Could be erroreneous we dont know if this
```
shresthamalik: There are only two main changes. Other diffs are because of logs that i added. I will point them out.
The key being generated was wrong due to the string above.

src/ngraph_encapsulate_op.cc
```diff
@@ -168,11 +169,10 @@ class NGraphEncapsulateOp : public OpKernel {
...
       // seems to screw things up in the C++ unit tests.
       // m_freshness_tracker->Unref();
-
-      // Release the backend
-      BackendManager::ReleaseBackend(m_op_backend_name);
-      NGRAPH_VLOG(2) << "~NGraphEncapsulateOp()";
     }
+    // Release the backend
+    NGRAPH_VLOG(2) << "~NGraphEncapsulateOp():: ReleaseBackend";
+    BackendManager::ReleaseBackend(m_op_backend_name);
```
shresthamalik: We were releasing the backend only when freshness tracker was not null. It should be released in all the scenarios, so moved it outside the loop.

src/enable_variable_ops/ngraph_encapsulate_op.cc
```diff
@@ -172,22 +173,17 @@ class NGraphEncapsulateOp : public OpKernel {
...
-    // destructor is called at the very end, if some modifier that uses this
-    // tensor is still active.
     for (int i = 0; i < m_number_outputs; i++) {
-      string key = NGraphCatalog::CreateNodeKey(m_graph_id, node_name, i);
+      string key = NGraphCatalog::CreateNodeKey(m_graph_id, name(), i);
       if (NGraphCatalog::ExistsInEncapOutputTensorMap(key)) {
-        auto temp = NGraphCatalog::GetTensorFromEncapOutputTensorMap(key);
-        temp.reset();
         NGraphCatalog::DeleteFromEncapOutputTensorMap(key);
+        NGRAPH_VLOG(2) << "Deleting from output tensor map " << key;
```
shresthamalik: Checking for Exists before deleting though it is not required to log which tensors are being deleted.

PR #502, opened 2019-04-10
Changes GatherV2 indices to Node

jiminha:


jiminha:
- In SSD model, the input for gather indices are dynamically came from output of other op such as TopK or NMSV2. To support that, we need to change the indices from constant to Node tensor.  
- Able to run 4 gather tests successfully in test/test_gather.py

src/ngraph_mark_for_clustering.cc
```diff
@@ -431,7 +431,7 @@ Status MarkForClustering(Graph* graph,
       type_constraint_map["FusedBatchNormV2"]["T"] = {DT_FLOAT};
       type_constraint_map["FusedBatchNormGrad"]["T"] = NGraphNumericDTypes();
       type_constraint_map["GatherV2"]["Tparams"] = NGraphDTypes();
-      type_constraint_map["GatherV2"]["Tindices"] = NGraphIndexDTypes();
+      type_constraint_map["GatherV2"]["Tindices"] = NGraphDTypes();
```
sayantan-nervana: This should be `NGraphIndexDTypes`

jiminha: Fixed

PR #506, opened 2019-04-11
fix build problem for when building from `build_nnp.py --distributed`

XWangIntel:


tools/build_utils.py
```diff
@@ -71,10 +71,11 @@ def build_ngraph(build_dir, src_location, cmake_flags, verbose):
 
     import psutil
     num_cores = str(psutil.cpu_count(logical=True))
-    cmd = ["make", "-j" + num_cores, "install"]
```
avijit-nervana: Can you please explain why you are removing the "install" command?

XWangIntel: This command will try to run like `make -j80 install`, which doesn't work when call by the build script in NNP-transformer. We usually build nGraph with `make -j80` then `make install`.
I do have the concern why it works as it has been used for a long time.

avijit-nervana: You need to root cause this and find out why this isn't working for you and fix that. Without the `install` this script won't work. One suggestion I can provide is to break this into two pieces. First is to do the `make -j<N>` - as we have it now and then the next step is `make install` - see if that helps your use case.

avijit-nervana: Actually never mind - I see that you have already done what I am suggesting - so please ignore my comment above

XWangIntel:
@avijit-nervana Will this be merged? Really need it so that `build_nnp.py` can pull this change and build.

XWangIntel:
TESTNOW

PR #512, opened 2019-04-12
Sindhu/build_ngtf fix and update readme

sindhu-nervana:


README.md
```diff
@@ -57,6 +57,8 @@ TensorFlow model scripts and running them the usual way:
 
         import ngraph_bridge
 
+Examples on how to use ngraph_bridge is here : https://github.com/NervanaSystems/ngraph-tf/tree/master/examples
```
avijit-nervana: There is a better way to refer to this URL (as it's within the repo). But ok for now.

PR #526, opened 2019-04-16
Shrestha/Added python function to query ngraph-bridge library

shresthamalik:
Add function to query ngraph bridge python module for build flag --enable_variables_and_optimizers


python/ngraph_bridge/__init__.in.py
```diff
@@ -39,7 +39,7 @@
     'enable', 'disable', 'is_enabled', 'backends_len', 'list_backends',
     'set_backend', 'is_supported_backend', 'get_currently_set_backend_name',
     'start_logging_placement', 'stop_logging_placement',
-    'is_logging_placement', '__version__'
+    'is_logging_placement', '__version__', 'is_grappler_enabled' ,'are_variables_enabled'
```
avijit-nervana: Place this to the next line

python/ngraph_bridge/__init__.in.py
```diff
@@ -188,10 +189,15 @@ def is_logging_placement():
 def is_grappler_enabled():
     return ngraph_bridge_lib.ngraph_tf_is_grappler_enabled()
 
+def are_variables_enabled():
```
avijit-nervana: Also add a function to return the CXX11ABI (i.e., a python function that returns an int reflecting the ngraph_bridge_lib.ngraph_tf_cxx11_abi_flag)
Something like: 
```
int cxx11_abi_flag() { return ngraph_bridge_lib.ngraph_tf_cxx11_abi_flag }
```

shresthamalik: It's already there, no?
https://github.com/NervanaSystems/ngraph-tf/blob/master/python/ngraph_bridge/__init__.in.py#L195

shresthamalik: oh, python function is missing. Working on it.

PR #529, opened 2019-04-18
Sarkars/flib ngenc compute

sayantan-nervana:


src/ngraph_cluster_manager.cc
```diff
@@ -34,7 +34,7 @@ int NGraphClusterManager::NewCluster() {
 
 GraphDef* NGraphClusterManager::GetClusterGraph(int idx) {
   std::lock_guard<std::mutex> guard(s_cluster_graphs_mutex);
-  return s_cluster_graphs[idx];
+  return idx < s_cluster_graphs.size() ? s_cluster_graphs[idx] : nullptr;
```
sayantan-nervana: Need this change since now we might have nothing stored in `NGraphClusterManager`, so need to safely avoid accessing out-of-bounds requests

src/ngraph_encapsulate_op.cc
```diff
@@ -18,6 +18,7 @@
 #include <utility>
 
 #include "tensorflow/core/common_runtime/dma_helper.h"
+#include "tensorflow/core/common_runtime/function.h"
```
sayantan-nervana: Included for `FunctionBody`

test/python/flib_graph_1.pbtxt
```diff
@@ -0,0 +1,284 @@
...
+    key: "T"
+    value {
+      list {
+        type: DT_FLOAT
+      }
+    }
+  }
+}
+library {
+  function {
```
sayantan-nervana: The library of graphdefs

test/python/flib_graph_1.pbtxt
```diff
@@ -0,0 +1,284 @@
...
+    value {
+      list {
+        type: DT_FLOAT
+      }
+    }
+  }
+}
+node {
+  name: "Sigmoid"
+  op: "IdentityN"
```
sayantan-nervana: `IdentityN` due to Kanvi's change

test/python/test_flib.py
```diff
@@ -0,0 +1,89 @@
...
+
+        res1 = self.with_ngraph(sess_fn, graph=graph)
+        res2 = self.without_ngraph(sess_fn, graph=graph)
+        exp = [np.full((2, 3), 3.0), np.full((2, 3), 0.95257413)]
+        # Note both run on Host (because NgraphEncapsulate can only run on host)
+        assert np.isclose(res1, res2).all()
+        # Comparing with expected value
+        assert np.isclose(res1, exp).all()
+
+    @pytest.mark.skip(reason="Not passing through grappler")
```
sayantan-nervana: @kanvi-nervana : I think this is a minimal repro of your squeeze-net issue

test/python/test_flib.py
```diff
@@ -0,0 +1,92 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+# ==============================================================================
+"""nGraph TensorFlow bridge floor operation test
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import pytest, pdb
```
sayantan-nervana: remove pdb

src/ngraph_encapsulate_op.cc
```diff
@@ -93,9 +94,30 @@ class NGraphEncapsulateOp : public OpKernel {
     OP_REQUIRES_OK(ctx, ctx->GetAttr<int>("ngraph_cluster", &m_ngraph_cluster));
     graph_def = NGraphClusterManager::GetClusterGraph(m_ngraph_cluster);
 
-    GraphConstructorOptions opts;
-    opts.allow_internal_ops = true;
-    OP_REQUIRES_OK(ctx, ConvertGraphDefToGraph(opts, *graph_def, &m_graph));
+    if (graph_def == nullptr) {
```
sindhu-nervana: Can we wrap this with OP_REQUIRES too?

test/python/common.py
```diff
@@ -50,7 +50,7 @@ def with_ngraph(self, l, config=tf.ConfigProto()):
 
         os.environ['NGRAPH_TF_DISABLE_DEASSIGN_CLUSTERS'] = '1'
         ngraph_bridge.enable()
-        with tf.Session(config=config) as sess:
+        with tf.Session(graph=graph, config=config) as sess:
```
shresthamalik: Why do we need this change? Just Curious.

src/ngraph_encapsulate_op.cc
```diff
@@ -93,9 +94,30 @@ class NGraphEncapsulateOp : public OpKernel {
...
+      // Read graphdef from function library
+      const FunctionLibraryDefinition flib =
+          *ctx->function_library()->GetFunctionLibraryDefinition();
+      const FunctionDef* fdef =
+          flib.Find("Enc_" + to_string(m_ngraph_cluster) + "_native_segment");
```
shresthamalik: Can we use the node name (eg. ngraph_cluster_251) instead of this "Enc_" + to_string(m_ngraph_cluster)" ?

test/python/common.py
sayantan-nervana: when we write python code like `a = tf.constant(...); b = tf.constant(...); c = a+b`, this underlying graph is added to the default graph. When we read it from a pbtxt (like in this case), the graph is not added to default graph. So was passing the graph along to `common.py` to use during session construction.

But found a way to set the graph read from pbtxt as default graph (`as_default`)

src/grappler/ngraph_optimizer.cc
```diff
@@ -64,7 +64,8 @@ Status NgraphOptimizer::Optimize(tensorflow::grappler::Cluster* cluster,
   // we will not do anything; all subsequent
   // passes become a no-op.
   if (config::IsEnabled() == false ||
-      std::getenv("NGRAPH_TF_DISABLE") != nullptr) {
+      std::getenv("NGRAPH_TF_DISABLE") != nullptr ||
+      IsProcessedByNgraphPass(&graph)) {
     NGRAPH_VLOG(0) << "NGTF_OPTIMIZER: Ngraph is disabled ";
```
shresthamalik: Include the or condition in the error message.

